PR #1390, opened 2014-03-16
Troll guards remix

gdpelican:
Cherry picked troll guard commits
- Tested user deletion capability
- Tested throttling with low tolerances (1 group creation / min)
- Used Measurement class for measuring throttles (removed Harness)


gdpelican:
Necromancy! :8ball:

gdpelican:
Alright so the gist of this is as follows:

- Every time a user makes a request to a particular endpoint (create group, create vote, create other stuff in this instance), Rack Attack logs the IP address of that user. (There's other things you can log by, but Kickstarter uses IP address and I'm not sure why our use case would be different)
- If a user does this action more than X times in Y seconds (as defined by a yml configuration), they start getting 429 responses (429 is HTML SC for 'too many requests')

rdbartlett:
Sounds great James. Are we notified when someone goes over limit?

gdpelican:
No, but maybe I could write a SLACK WEBHOOK for it! :lightning:!

robguthrie:
Can it please just be an exception that happens on 1, 10, 100, 1000 times?

gdpelican:
Okay, this is finally ready to go, with a cuke to prove it works and everything. Want to give it a onceover @robguthrie?

app/views/discussions/_discussion_context.html.haml
```diff
@@ -17,7 +17,7 @@
     - if can? :set_volume, @discussion
       - current_volume = @discussion_reader.volume
       .btn-group.dropdown.volume-dropdown
-        %button{type: 'button', class:"btn btn-left btn-default dropdown-toggle", 'data-toggle': "dropdown", 'aria-expanded': false}
+        %button{type: 'button', class:"btn btn-left btn-default dropdown-toggle", 'data-toggle' => "dropdown", 'aria-expanded' => false}
```
gdpelican: I'm not sure what the cause of this is (a gem version difference maybe?), but I get compile errors when using this 'string': 'value' syntax. Rather just change it than have it trip me up anymore.

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,23 @@
+Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+Rack::Attack.throttled_response = ->(env) { [429, {}, [ActionView::Base.new.render(file: 'public/429.html')]] }
+
+ActiveSupport::Notifications.subscribe('rack.attack') do |name, start, finish, request_id, req|
+  Airbrake.notify Exception.new(message: "#{req.ip} has been rate limited for url #{req.url}", data: req.env['rack.attack.throttle_data'])
```
gdpelican: Notifies airbrake with a message and the throttle data, which looks like:

```
{ count: 6, limit: 5, period: 60}
```
Ie, this person attempted to hit this endpoint 6 times in the last 60 seconds, which is greater than the rate limit of 5.

gdpelican: I'll readily admit this might not be the optimal info to shoot through to airbrake, but once we start getting some of these come through in prod I'll have a much better idea what's useful to put here.

robguthrie: This looks great. 
Only thing that I feel needs to change is the MemoryStore usage - seeing as we're seeing memory problems.

Can you use ActiveSupport::Cache::MemCacheStore in production?   You might even be able to use the cache store defined in the environment.

robguthrie:
A million thank yous j money. I will deploy this tomorrow eh?

gdpelican:
Yeah, MemCacheStore looks like the way to go for prod.

rdbartlett:
Nice one James

robguthrie:
Indeed. Blimmin great James.

PR #1715, opened 2014-10-07
Facebook locale 2

s01us:
see Part 1 here https://github.com/loomio/loomio/pull/1703

looking for a good way to test this in a light-weight way. FB uses strange locales (like `el_GR` for greek, when everyone else uses `el`, they also use only underscore).


robguthrie:
please continue this without the open pull request mix

PR #1731, opened 2014-10-30
make RTL languages profile options even though incomplete

s01us:
make RTL locales that are not complete selectable. This is so that they can post and have the app recognise they are typing in a RTL (right to left) script.

I wanted to make it clear these are not complete in the language dropdown menu, so added a section for them at the bottom 

looks like this : 

image

### DIFF
--- a/app/helpers/locales_helper.rb
+++ b/app/helpers/locales_helper.rb
@@ -9,6 +9,10 @@ module LocalesHelper
     Loomio::I18n::DETECTABLE_LOCALES
   end
 
+  def rtl_locales
+    Loomio::I18n::RTL_LOCALES
+  end
+
   def locale_name(locale)
     I18n.t(locale.to_sym, scope: :native_language_name)
   end
@@ -48,7 +52,7 @@ module LocalesHelper
   end
 
   def language_dir(locale)
-    Loomio::I18n::RTL_LOCALES.include?(locale) ? 'RTL' : 'LTR'
+    rtl_locales.include?(locale) ? 'RTL' : 'LTR'
   end
 
   # View helper methods for language selector dropdown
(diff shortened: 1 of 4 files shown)


app/helpers/locales_helper.rb
```diff
@@ -60,9 +64,17 @@ def linked_language_options
...
       [locale_name(locale), locale]
     end
+
+    extra_options = (rtl_locales - selectable_locales).map do |locale|
+      ["incomplete: " + locale_name(locale), locale]
+    end
+
+    supported_options +
+    [['---------------------------------', 'disabled_option']] +
+    extra_options
```
robguthrie: so you want to pair with me on this @s01us ?

gdpelican:
Hey @s01us @robguthrie I want to do something with this PR, seeing as it's 6 months old at this point. Either we should do a bit of work and get it merged in, or close it and move on. Thoughts? :bulb:

s01us:
Sorry I've not been reading my emails. Too much shit going on.

I'm keen to get this worked into a mergeable state

gdpelican:
The primary bit of work is cleaning up that `language_options` method, yeah?

gdpelican:
I had a wee poke at streamlining that language options method a little. Let me know what you think?

Some thoughts:
- Should we be translating that 'incomplete' prefix?
- The `LocalesHelper` could probably use a bit of love somewhere; I can't seem to shake the idea that a 'Locale' and a 'LocaleCollection' should be models which we can manipulate so we don't need to have so much logic in helper methods. I envision something like

```
locale = Locale.new(:en)
locale.direction # => "LTR"
locale.option # => ["English", "en"]
locale.option(link: true) # => ["English", "/users/sign_in?locale=en"]
locale.to_s # => "en"
locale.rtl? # => false
locale.current?
locale.selectable?
locale.detectable?
locale.incomplete?
```

etc.

Then a LocaleCollection might be able to do stuff like:

```
def detectable
  @locales.select(&:detectable?)
end

def selectable_options
  @locales.select(&:selectable?).map(&:option)
end

def incomplete
  @locales.select(&:rtl?).reject(&:selectable?)
end
```

Thoughts?

gdpelican:
I'm happy to merge this in @robguthrie @s01us unless there are qualms.

(The test suite failure is an intermittent off-by-1-second error; this build looks good to me.)

PR #1804, opened 2014-12-15
Angular search

gdpelican:


gdpelican:
This is probably ready for a look, @robguthrie

Probably something we should have a chat about before I dive further, as it's already kinda a big PR.

It currently returns results for discussions, motions, and comments (limited to 5 each), sorted (somewhat haphazardly) by relevance, although it's set up so that tweaking the ranking algorithms will be easy.

I'm caching tsvectors in a separate table as the model gets updated, which is the gem of how Discourse does their search... anyway blah blah technical stuff let's talk about it.

robguthrie:
Great. Had a look, it's kinda crazy huh? The mix between loomio specific
stuff and general "searchable" stuff is a bit confusing at first.

Maybe there is a gem we can extract from this? Do you think so?

Tangential question: is summer 2015 the time we rename motion to proposal
throughout the system?

gdpelican:
Definitely a potential gem in there, probably around saving search vector
data.

I was a little hesitant to just generalize the crap out of it, though,
because there'll be model-specific conditions which will start popping up
at some point, like 'Weight proposals from the past month higher', or
'weight comments from users who I interact with regularly higher', or
'weight discussions which I've participated in', or whatever else.

The idiom was kinda 'general on the way in, specific on the way out' (So,
discussions, motions, and comments all save their search data the same way
when they're saved, but on the way out we have places to manipulate each
set of results individually in whatever way makes it most relevant.)

robguthrie:
So what is it that we're saving?

gdpelican:
Ah, we're storing a tsvector, which is like a hash of lexemes in a particular model; it'll look like this:

title: "Let's go to the beach"
tsvector: "let: 1 go: 2 beach: 3"

Postgres wants to search against these ts_vectors, rather than having to deal with walls of text, in a query that typically looks like

`SELECT * FROM discussions WHERE to_tsvector("Let's go to the beach") @@ tsquery('beach');`

Except, it takes a little while to run to_tsvector, especially if there's a lot of text and more especially if we have to run it against a million discussions every time we search. So instead, after the discussion is saved, we create a ts_vector for it and save it in a separate `discussion_search_vectors` table (I asked why a separate table and got [this response](https://meta.discourse.org/t/discourses-search-implementation/23762/2). 

That way, when we search, it doesn't have to touch the discussion table at all, since all the data it needs is already stored:

`SELECT discussion_id FROM discussion_search_vectors WHERE ts_vector @@ tsquery('beach')`

robguthrie:
Arrr! That makes a lot of sense. Thank you for that.

app/models/search_vectors/base.rb
```diff
@@ -0,0 +1,74 @@
+class SearchVectors::Base < ActiveRecord::Base
```
robguthrie: As we discussed, it would be nice to be singular for the namespace

spec/models/search_result_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+describe Comment do
```
robguthrie: SearchResult

PR #1825, opened 2014-12-31
add gitter badge to README.md

s01us:
add's a badge that looks like this
[Gitter](https://gitter.im/loomio/loomio)
to the front page. 
I'm interested in running this as an experiment for a fixed period of time with @gdpelican

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![Build Status](https://travis-ci.org/loomio/loomio.png?branch=master)](https://travis-ci.org/loomio/loomio) 
 [![Code Climate](https://codeclimate.com/github/loomio/loomio.png)](https://codeclimate.com/github/loomio/loomio) 
 [![Dependency Status](https://gemnasium.com/loomio/loomio.png)](https://gemnasium.com/loomio/loomio) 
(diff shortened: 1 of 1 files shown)


robguthrie:
what would this mean for the hipchat based chat that is already public? I would rather put a sweet link into that.

s01us:
interesting question.
I think gitter is just super accessible compared to those things (because
it just uses your github account).

Doesn't mean we can't have other chat channels. I'd be interested to run a
trial and see how it is.

re: hipchat, I'm helping enspiral move over to slack in mid Jan, so imagine
loomio would be interested in that shift too?

robguthrie:
I reckon you should ask the team properly before doing this. 

I feel like I'm going to have to support this if it goes ahead, and I'm not sure I want another comms channel, and I don't think this is superior to the existing channel you could easily link to in the same way.

elimisteve:
> what would this mean for the hipchat based chat that is already public?

I have searched for a Loomio HipChat channel and have been unable to find it; got a link?  (Also, there are but 2 of us in #loomio on Freenode :-)

gdpelican:
Going to close this for now, we can merge if we end up making a decision to open a gitter chat channel and use it more regularly.

s01us:
good call. can you post in the gitter about where and how people should do
comms?

PR #1828, opened 2015-01-06
recreate locale updater in Node

s01us:
a learning project that has generated better bug catching for translations.

It's a node script that asynchronously updates translations.
I've isolated the script in `lib/tasks/locales` and linked to it with a shell script `locales.sh` 

I'm going to work with @ahdinosaur to refactor this in the futue


s01us:
and damn is it fast

robguthrie:
If it's a node project, and it applies generally to translations in any
rails app, I think this should not be in the main Loomio source mix. Are
you happy to make another repo under the loomio organisation?

s01us:
@robguthrie  good idea! 
that way I can make a mess and it's contained

xx

PR #1831, opened 2015-01-08
add simple translation stats store

s01us:
@robguthrie in my node script I've added a function which stores the current percentage translated of each locale.

I'm interested in storing this in the app so that : 

- I can easily see which locales were in TEST but should now be LIVE
- in the future we can show users how complete translations are.

where should this file live. I don't think `config/locales` is a good choice, since rails loads the files in there and adds them to the list of translations. Having said that I could make the top levels of this yaml `en.translation_stats`. Would certainly make pulling these numbers into the app easier.

### DIFF
--- a/lib/tasks/translation_stats.yaml
+++ b/lib/tasks/translation_stats.yaml
@@ -1,0 +1,144 @@
+ar:
+  main: 48
+  frontpage: 100
+an:
+  main: 99
+  frontpage: 100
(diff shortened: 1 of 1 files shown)


robguthrie:
I think just putting it in the root would be ok /.translation_coverage or something

s01us:
cool, I'll rework that

PR #1835, opened 2015-01-12
Fix for angular invitations

gdpelican:


lineman/app/models/base_endpoints_interface.coffee
```diff
@@ -0,0 +1,22 @@
+angular.module('loomioApp').factory 'BaseEndpointsInterface', (RestfulClient) ->
```
gdpelican: This fulfills a need that came up a couple times in the past, which is that we want to be able to hit an endpoint without having to actually try to store anything to the record store. This means I don't have to make `invitable_model.coffee` or `invitation_model.coffee`, which would just be cruft.

robguthrie: Thanks James. I'm not sure I totally get this yet. Would you mind helping me understand?

I get some of it.. it's nice to have some stuff around RestfulClient, but I don't get why it needs to be knitted into the Records interface if it's not doing anything with the rest of the records. Do invitation records get relationship access to other records or something?

gdpelican: I'm not 100% on it either, but basically I know that

- We have use cases here where we want to interact with the API, but not
necessarily deal with the record store
- The only way to access the RestfulClient at the moment is either directly
(ick), or through the BaseRecordsInterface, which requires a model class to
make it work.
(I'm looking at the line `@restfulClient = new
RestfulClient(@model.plural)` in the BaseRecordsInterface)

It felt pretty good to write Records.invitables.fetch() and
Records.invitations.create(), but maybe it wants to be even more separate,
like Endpoints.invitables.fetch() and Endpoints.invitations.create()

??

robguthrie: I think it makes Records and the related classes potentially way more complex if there is a similar but different thing in there. It's about as complex as I can handle already.

Could this just be InvitationsClient or InvitaitonsEndpoint?  If it's going to be in Records then I think it needs to be actually using the record store, otherwise it will get in the way of the continuing development of the Records concept.

Can you remind me why this isn't a normal kind of record like everything else?

robguthrie: Oh, i remember,  because Invitable is an interface wrapper around groups, people, email addresses, which are all stored differently in the backend.

My preference would be to highlight that this is just building upon the restfulclient (I'm not quite sure why you said yuck before) and call it an InvitablesClient, and keep it out of the Records stuff.

gdpelican:
Fixed this up so it uses classes extended off of RestfulClient.

robguthrie:
OMG. It's awesome. You're even more awesome.

lineman/app/modules/group_page/invitations_modal_controller.coffee
```diff
@@ -38,6 +38,7 @@ angular.module('loomioApp').controller 'InvitationsModalController', ($scope, $m
     $scope.isDisabled = false
     $scope.invitations = []
     $modalInstance.close()
+    Records.memberships.fetchByGroup $scope.group
```
robguthrie: please add comment:
reget memberships for group, as this are not returned with invitations create response

robguthrie:
superb!

PR #1836, opened 2015-01-12
small stylefix for nested bullet-point markdown lists

s01us:
there's a problem with the styling on markdown. 

I've extended what looks like a patch-job.
this needs testing in browser. I don't have time to update ruby atm :(


robguthrie:
Thanks Mix

PR #1840, opened 2015-01-12
Disable group edit submit button while submitting

### LINKED ISSUE #1832: Create subgroup button is not disabled once clicked #angular
It happens in the angular version. 
This leads to multiple subgroup creation when users click multiple time before the subgroup is created.
screen shot 2015-01-08 at 2 00 37 pm

yourcelf:
(Angular site).  Fixes #1832

### DIFF
--- a/lineman/app/templates/group_form.haml
+++ b/lineman/app/templates/group_form.haml
@@ -1,76 +1,77 @@
 .container#main{ng-class: '{inboxPinned: inboxPinned()}'}
   %form.form-card{name: 'groupForm', ng_submit: 'submit()'}
-    %h1{ng_show: 'group.isNew() && !group.parentId', translate: 'group_form.new_group_title'}
-    %h1{ng_show: 'group.isNew() && group.parentId', translate: 'group_form.new_subgroup_title', translate-value-parent-group-name: '{{group.parentName()}}'} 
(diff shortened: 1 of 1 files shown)


gdpelican:
Hey @yourcelf, thanks for your contributions!

The Loomio dev team is currently still on summer holiday, so we won't be merging anything in for the next week or two, after which we can put our heads together and make some decisions surrounding open source contributions and contributors. :+1:

robguthrie:
This kind of contribution is cool. Thanks @yourcelf. I think it's the right thing to do, but I think in general we will want most/all of our user facing forms to completely disable while pending network operation is happening. 

We've been doing a bit of that around the place.. IE: ng_disabled at form level rather than button level.

As james said, we're not very responsive right now, which is a bit sad when it comes to supporting you at this very moment, but we'll be back full time in a week or so.

yourcelf:
Makes sense. I've updated the PR to add a ``<fieldset>`` on which to apply ``ng_disabled`` (disabled attribute doesn't work on ``<form>`` elements themselves).  The resulting diff looks substantially more intimidating because of the indentation change, but it's just the one tag added.

gdpelican:
Thanks @yourcelf!

PR #1842, opened 2015-01-12
Translations update (ru, el,sv)

s01us:
hey @robguthrie this is ready. It's: 

- standard locale update
- inclusion of the new `.translation_stats.yml` in the app root, as [discussed previously](https://github.com/loomio/loomio/pull/1831)

### DIFF
--- a/.translation_stats.yml
+++ b/.translation_stats.yml
@@ -1,0 +1,144 @@
+ar:
+  frontpage: 100
+  main: 48
+an:
+  frontpage: 100
+  main: 99
+be_BY:
+  frontpage: 100
+  main: 98
+bg_BG:
+  frontpage: 18
+  main: 44
+ca:
+  frontpage: 81
+  main: 83
+cmn:
+  frontpage: 0
+  main: 31
+zh_TW:
+  frontpage: 100
+  main: 100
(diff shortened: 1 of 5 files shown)


s01us:
@robguthrie I've had another good go at this with James, and can't deploy this without pulling from heroku or forcing the commit.

Can we make a time to troubleshoot what's going on with this deploy issue please?
maybe easy if we're sitting next to each other

robguthrie:
Same advice as before: Force it and see what happens. Don't worry!

robguthrie:
I thought we agreed on translation_coverage... which is more accurate than
"stats"

s01us:
I'll try a force next time.
I think you said translation_coverage "or something".
If it's a problem i/you can change it, i don't mind

robguthrie:
you're right, I probably did. I just thing that the file won't make much
sense to someone if it's called stats, but coverage really gets close to
what the information in the file is.

s01us:
I honestly don't mind. I do have a preference towards acting and moving on.

If you'd like it changed, can you either ask me to do that or change it?

Xx

robguthrie:
Heh, thanks please change.

PR #1843, opened 2015-01-13
Increase limit on memberships returned by API

### LINKED ISSUE #1838: memberships API endpoint only lists first 5 users
The API endpoint at ``/api/v1/memberships?group_id=<id>`` limits the returned results to [5 users](https://github.com/loomio/loomio/blob/51c799a239e7be5e26f1fae8217a7d54a067d25e/app/controllers/api/memberships_controller.rb#L7) (presumably because it was originally written just to support autocomplete?). This breaks the angular memberships display (go to https://www.loomio.org/angular and click "Show More" under members), as well as limiting utility of the endpoint for 3rd party integrations.

Is it safe to simply remove the "limit: 5" from the API call?  Or would this break an autocomplete or some such elsewhere?

yourcelf:
Adds "limit" parameter to enable limiting of returned memberships, e.g.
``/api/v1/memberships?group_id=<id>&limit=5``.  Limit defaults to 1000
when not provided.  Includes tests.

Fixes #1838.

### DIFF
--- a/app/controllers/api/memberships_controller.rb
+++ b/app/controllers/api/memberships_controller.rb
@@ -4,7 +4,9 @@ class API::MembershipsController < API::RestfulController
     @group = Group.find(params[:group_id])
     authorize! :show, @group
 
-    @memberships = Queries::VisibleMemberships.new(user: current_user, group: @group, limit: 5)
+    @memberships = Queries::VisibleMemberships.new(user: current_user,
+                                                   group: @group,
+                                                   limit: params[:limit] || 1000)
     respond_with_collection
   end
 
--- a/spec/controllers/api/memberships_controller_spec.rb
+++ b/spec/controllers/api/memberships_controller_spec.rb
@@ -7,6 +7,7 @@ describe API::MembershipsController do
   let(:user_named_bang) { create :user, name: "Bang Whamfist" }
   let(:alien_named_biff) { create :user, name: "Biff Beef" }
   let(:group) { create :group }
+  let(:biggroup) { create :group }
   let(:discussion) { create :discussion, group: group }
   let(:comment_params) {{
     body: 'Yo dawg those kittens be trippin for some dippin',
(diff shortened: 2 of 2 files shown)


gdpelican:
Hm. You'll have to, at the very least, update the frontend to only return 5 members when we're loading the group page. As it stands we'd be loading up a thousand people each time we hit the group page.

I'm wondering if there's a different way we want to approach this, @robguthrie? I have a feeling the reason it was left like this is because we ended up a little unclear on how we wanted to handle pagination for these index calls.

robguthrie:
Sorry for the slow reply. If we allow each controller to have a default limit that suits the main loomio client use case, and allow people to specify their own limit when the default is not useful, would that work for everyone?

yourcelf:
Just a note on my selfish use-case: I'm working towards implementing an OAuth-accessible API to let loomio groups federate with 3rd party sites.  For this use-case, I need to be able to retrieve full membership rosters for groups.  A group with thousands of members would present difficulty -- I could query with an absurdly high limit, but it could be harmful to loomio's performance.

Most other APIs solve this sort of problem with pagination and a deterministic order (e.g. you could retrieve all of Lady Gaga's 43 million followers using their API, [20 at a time](https://dev.twitter.com/rest/reference/get/followers/list)).  This sort of approach would work fine for me and would be future-proof against large groups, but also more involved to implement.

gdpelican:
I think pagination is the way to go here as well, but I'd be interested in
generalizing that across all of our indexing actions, which I don't think
we're quit ready to stick our head into yet.

robguthrie:
Oh yes definitely pagination. Using Kaminari should be simple and easy for
most resources. It would be cool to include totals/pagination metadata in a
uniform way too

robguthrie:
Closing pull request for now. Expect to see solution for all endpoints soon. Thanks for bringing this up!

PR #1859, opened 2015-01-30
Add support for signing up with Twitter

jakub-:


jakub-:
I've looked through past discussions and the Trello board for some clues as to why Twitter would not have been supported but couldn't find anything. I therefore assume that is not intentional.

There is one notable difference between Twitter and the already supported Facebook & Google Plus in that the identity that the latter two ascertain is arguably of stronger credibility than Twitter due to their real name policies. I am not sure how much that matters for Loomio, though, and if anything I imagine the importance of this would also differ from use case to use case.

app/assets/stylesheets/unstructured/_pages.css.scss
```diff
@@ -609,55 +609,37 @@ body.pages.privacy .container h3.subheader { margin-bottom: 24px; }
...
-  filter: progid:DXImageTransform.Microsoft.gradient(startColorstr="#163e82", endColorstr="#0f2956");
-  background-image: -khtml-gradient(linear, left top, left bottom, from(#163e82), to(#0f2956));
-  background-image: -moz-linear-gradient(top, #163e82, #0f2956);
-  background-image: -ms-linear-gradient(top, #163e82, #0f2956);
-  background-image: -webkit-gradient(linear, left top, left bottom, color-stop(0%, #163e82), color-stop(100%, #0f2956));
-  background-image: -webkit-linear-gradient(top, #163e82, #0f2956);
-  background-image: -o-linear-gradient(top, #163e82, #0f2956);
-  background-image: linear-gradient(#163e82, #0f2956);
-  border-color: #0f2956 #0f2956 hsl(218, 70%, 17.5%);
+  filter: progid:DXImageTransform.Microsoft.gradient(startColorstr="$from", endColorstr="$to");
```
gdpelican: Can I get a vote of confidence that this transposes correctly?

gdpelican:
This looks pretty great, @jakub-, thanks! I'm looking forward to spinning it up and testing it out. :walking:

app/assets/stylesheets/unstructured/_pages.css.scss
jakub-: @gdpelican Good catch, thanks! Fixed. :)

gdpelican:
Seems to work wonderfully. Thanks again @jakub-!

jakub-:
@gdpelican Very happy to be contributing. :-) Thanks for making Loomio!

rdbartlett:
beautiful, thanks @jakub- :)

robguthrie:
This is nice work, thanks @jakub-. 

@gdpelican - I would have liked to have seen/approved a screenshot of the changes to the login page, before this was merged, as that's a really important page in the UX.

jakub-:
@robguthrie Good call, I'll try to remember to attach a screenshot next time.

gdpelican:
Just successfully signed using twitter on www.loomio.org. That rocks!

robguthrie:
So did I!

PR #1861, opened 2015-01-31
Upgrade Rails to 4.2

jakub-:


gdpelican:
Hey, this is a pretty great start, @jakub-, thanks! Let me know if you need support pushing it further along.

robguthrie:
Thanks for starting this off @jakub- 

Do you have any thoughts on supporting ActiveJob? I see it's part of rails 4.2, but I think it would mean that we need to change EmailMailer.delay calls into an ActiveJob subclass, which might actually be quite nice, but I'm not sure if it is required or just the new way to do it in 4.2

jakub-:
@robguthrie Rails 4.2 integrates ActiveJob with ActionMailer such that you can now simply call `group_email(...).deliver_later` as opposed to `delay.group_email(...)`. I don't think it's necessary that we do this migration now but since it's trivial enough, may as well. :)

I made some progress on this but one feature is still failing, I'll take a look tomorrow!

gdpelican:
Hey @jakub- I thought I'd check in and see if you were still working on this? 
Would be nice to get up to 4.2, but don't want to take it over and step on your toes if you'd like to continue work on it.

gdpelican:
Migrating this to #1931

PR #1869, opened 2015-02-05
translation update (be-BY, de, fr, hu, uk, ry, sv, uk)

s01us:


s01us:
@gdpelican let me know when things are safe to deploy so I can announce this update

thanks :fish:

PR #1870, opened 2015-02-05
Fix a constraint violation when revising a vote on a proposal more than twice

jakub-:


jakub-:
Steps to reproduce:
- Open a proposal
- Vote.
- Update the vote.
- Update the vote again.

You get:
> ERROR:  duplicate key value violates unique constraint "vote_age_per_user_per_motion"

app/models/vote.rb
```diff
@@ -90,6 +90,7 @@ def associate_previous_vote
   end
 
   def age_previous_votes
+    Vote.connection.execute("SET CONSTRAINTS vote_age_per_user_per_motion DEFERRED")
```
robguthrie: I think this needs to be in a database migration, rather than in application code. It seems like something that only needs to be run once on an existing database.

robguthrie:
Great patch. See my notes about making a database migration for the constraint line. Thank you!

spec/models/vote_spec.rb
```diff
@@ -30,6 +30,30 @@
...
+        before do
+          vote3
+        end
+
+        subject { vote3 }
+        its(:age) { should == 0 }
+
+        it "should age the first vote" do
+          vote.reload
+          vote.age.should == 2
```
gdpelican: This should syntax is deprecated; please use the newer expect..to syntax

```
expect(vote.reload.age).to eq 2
```

app/models/vote.rb
jakub-: We already have a migration that ensures the constraint is deferrable until commit. By default, however, it is still immediate. This line will defer the constraint only within the current transaction hence it's in application code.

jakub-:
Updated the spec to use the newer `expect()` constructs.

app/models/vote.rb
robguthrie: Ok. I'll take your word for it. Thanks for this.

jakub-: No worries! There are some gems that make this more seamless but sadly they're either out of date (https://github.com/pedz/activerecord_constraints) or only support deferring foreign key constraints (https://github.com/zoocasa/activerecord-postgresql-extensions).

jakub-:
Hmm... this PR is not going to fly, sorry! I should have done more research before opening it.

Essentially, the problem does not exist if you set up the database by running all the migrations. It only exists if you use `schema.rb` as the canonical source of truth (since the DEFERRABLE property of the index is lost). Hence why I couldn't reproduce this on loomio.org.

What really needs to happen is that I make `schema.rb` correctly capture the DEFERRABLE property. This may involve migrating to [schema_plus](https://github.com/SchemaPlus/schema_plus) and potentially also patching it as, again, it only seems to support to `DEFERRABLE` on foreign keys.

Closing this PR for now but I think it's important to address this so I'll open a new one soonish.

robguthrie:
Ohhh! Now that makes way more sense to me!

Thanks for all your hard work on this, anyway.

PR #1873, opened 2015-02-06
Angular search deux

gdpelican:
(Not for merging)

Hey sooo.. I kinda rewrote search over a few hours yesterday. Here's my revamped solution. (I'll annotate is some shortly)


app/models/discussion.rb
```diff
@@ -5,7 +5,16 @@ class Discussion < ActiveRecord::Base
 
   include ReadableUnguessableUrls
   include Translatable
-  include Searchable
+
```
gdpelican: No more searchable; the only search vectors we have are discussion search vectors.

app/models/search_result_child.rb
```diff
@@ -0,0 +1,14 @@
+SearchResultChild = Struct.new(:type, :id, :blurb) do
```
gdpelican: A 'search result child' is a discussion, motion, or comment within a search result.

Each search result contains:

1 discussion
up to 2 proposals (the 2 most relevant to the search term)
up to 2 comments (the 2 most relevant to the search term)

app/models/search_vector.rb
```diff
@@ -0,0 +1,8 @@
+class SearchVector < ActiveRecord::Base
+  belongs_to :discussion
+  self.table_name = 'discussion_search_vectors'
```
gdpelican: If we end up going with this solution, we'll change this to just be 'search_vectors'

app/serializers/search_result_serializer.rb
```diff
@@ -1,39 +1,12 @@
 class SearchResultSerializer < ActiveModel::Serializer
-  embed :ids, include: true
```
gdpelican: This serialization is now a million times nicer.

Here's a sample serialized search result, which I think contains everything we'd need to put a nice UI for it together (and we can include anything else on the objects we'd like)

  {:priority=>"24.2",
   :query=>"loomio",
   :discussion=>
    {:id=>"2644",
     :blurb=>"<b>Loomio</b> Community! \r\n\r\nThis is a place where everyone can introduce themselves to the group, so we can all get to know",
     :title=>"Introduce yourself to the Loomio Community",
     :created_at=>Sat, 16 Mar 2013 00:28:57 UTC +00:00},
   :proposals=>[],
   :comments=>
    [{:id=>"206075", :blurb=>"<b>Loomio</b> right now to participate in the build up process of a new political party", :author=>"Hannah Salmon"},
     {:id=>"202156", :blurb=>"about it. I've been using <b>loomio</b> for a while and I find it very useful.", :author=>"Nico Aumar"}]}}

(Note that the <b> tags around 'loomio' are denoting that that's the term we searched for)

app/serializers/search_results/base_serializer.rb
```diff
@@ -0,0 +1,3 @@
+class SearchResults::BaseSerializer < ActiveModel::Serializer
```
gdpelican: This base serializer can be extended to include extra attributes from the model for a discussion, motion, or comment.

app/services/search_algorithms.rb
```diff
@@ -0,0 +1,88 @@
+class SearchAlgorithms
```
gdpelican: I think it's nice to have all the raw SQL in once place, separate from the service.

app/services/search_algorithms.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def self.sync_search_vector
```
gdpelican: So, on insert, we do a bit of extra work to create a big search vector for each discussion. This search vector includes weighted data for the discussion, its proposals, and its comments.

I haven't quite decided whether this is crazy or not. It happens in a background job, and the query is still really performant even on big discussions with tons of comments, but I wonder whether doing this every time anything gets made in the app will blow out our delayed job queue.

You can check out the weights I chose in `search_algorithms#discussion_vector`

app/services/search_service.rb
```diff
@@ -0,0 +1,55 @@
...
+    ).map { |result| build_search_result result }
+  end
+
+  private
+
+  def self.visible_results_for(user)
+    Queries::VisibleDiscussions.new(user: user).pluck(:id)
+  end
+
+  def self.build_search_result(result)
```
gdpelican: So, now, when we get a result back, we know how relevant it is, but we don't know *which* proposals or comments made it more relevant. So we have to do a separate query to get these things.

This is the slowest part of the process, but it's still faster than the existing solution (I benchmarked it at ~1.1 seconds for a typical search on my prod database)

I'd also imagine there's ways to speed these relevant_motion/comment queries up a bit.

lineman/app/modules/navbar/navbar_search_items/navbar_search_comment.haml
```diff
@@ -1,2 +1 @@
-.navbar-search-comment
-  {{comment.body}}
\ No newline at end of file
+.navbar-search-comment{btf-markdown: 'result.blurb'}
```
gdpelican: Haven't done this all the way out the frontend yet, but can verify that we can handle rendering blurbs by making em markdown, woo!

robguthrie: Love to see all this experimentation. Thank you so much. This looks really
cool.

PR #1874, opened 2015-02-07
extract locale config into yml

s01us:
in creating a node script for updating the locales it's become clear our locales config is based into some ruby.

I've extracted them into yaml, so that they are : 

- easier for node to read 
- easier for users to find and edit

### DIFF
--- a/lib/loomio_i18n.rb
+++ b/lib/loomio_i18n.rb
@@ -1,29 +1,19 @@
 module Loomio
-  module I18n
+  class I18n
+    require 'yaml'
+    config = YAML.load_file('config/loomio_i18n.yml')['loomio_i18n']
 
-    SELECTABLE_LOCALES = %i( en an be-BY bg-BG ca cs zh-TW da de eo es el fr id it he hu ja ko ml nl-NL pt-BR ro sr sr-RS sk sl sv vi tr uk )
+    SELECTABLE_LOCALES = Array(config['selectable_locales']).map(&:to_sym)
+    DETECTABLE_LOCALES = Array(config['additional_detectable_locales']).map(&:to_sym) + SELECTABLE_LOCALES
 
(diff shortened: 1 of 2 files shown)


robguthrie:
FYI - we're on ruby 2.2 now, which garbage collects symbols... So not need to have the silly string stuff in places, you can convert to symbols as soon as you like and treat everything as symbols.

s01us:
sweet, I'll see what I can do. 
yaml doesn't really do symbol, so we'll still have some string being pulled in

lib/loomio_i18n.rb
```diff
@@ -1,29 +1,23 @@
 module Loomio
-  module I18n
+  class I18n
+    require 'yaml'
+    config = YAML.load_file("config/loomio_i18n.yml")['loomio_i18n']
 
-    SELECTABLE_LOCALES = %i( en an be-BY bg-BG ca cs zh-TW da de eo es el fr id it he hu ja ko ml nl-NL pt-BR ro sr sr-RS sk sv vi tr uk )
+    def self.symbolize(object)
```
gdpelican: Not totally sure you need this method to symbolize everything.

Would something like this do the job?

```
module Loomio
  class I18n
    require 'yaml'
    config = YAML.load_file("config/loomio_i18n.yml")['loomio_i18n']

    SELECTABLE_LOCALES = Array(config['selectable_locales']).map(&:to_sym)
    DETECTABLE_LOCALES = Array(config['additional_detectable_locales']).map(&:to_sym) + SELECTABLE_LOCALES

    RTL_LOCALES  = Array(config['rtl_locales']).map(&:to_sym)
    TEST_LOCALES = Array(config['test_locales']).map(&:to_sym)

    FALLBACKS = Hash(config['fallbacks']).with_indifferent_access

  end
end
```

gdpelican:
Looks pretty great to me. I'll merge once the build's done.

PR #1875, opened 2015-02-07
Make the database initialisable from `schema.rb` alone

jakub-:
This is a continuation of https://github.com/loomio/loomio/pull/1870.

With these changes, it will be possible to setup a functioning database solely from `schema.rb` without running a single migration, which is the best practice for Active Record.

This PR:
* Adds a dependency on `schema_plus` and, more specifically, my fork of `schema_plus_pg_indexes` that adds support for DEFERRABLE unique constraints that we use for `votes.age`.
* Updates schema to properly reflect the state of the post-migration database.
* Adds specs to ensure the original scenario from https://github.com/loomio/loomio/pull/1870 works.

Possible follow-ups:
* Delete old migrations so that we don't rely on them for bootstrapping of the database.


jakub-:
This isn't ready yet, by the way.

cc @robguthrie if you have opinions about the direction. :)

jakub-:
Hmm, I also have a simpler PR that addresses this problem without extra dependencies. Let's try that one first. Sorry for the noise!

PR #1871, opened 2015-02-05
Add margin to motion icons in the list of votes

jakub-:

PR #1876, opened 2015-02-07
Make the database initialisable from `schema.rb` alone

jakub-:
This is a continuation of #1870.

With these changes, it will be possible to setup a functioning database solely from schema.rb without running a single migration, which is the best practice for Active Record.

This PR:
* Adds a SQL statement to `db/seeds.rb` to ensure that the `vote_age_per_user_per_motion` constraint is deferrable. This would normally happen as part of the relevant migration but not when using `rake db:schema:load`.
* Adds specs to ensure the original scenario from #1870 works.
* Reverts a recent work-around to .travis.yml as it's no longer necessary.

Possible follow-ups:

* Delete old migrations so that we don't rely on them for bootstrapping of the database. Not necessarily a good idea given the sheer user base of Loomio and the non-zero likelihood of someone wanting to update a 2011 instance. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,7 @@ rvm:
   - 2.2.0
 before_script:
   - "cp config/database.yml.$DB config/database.yml"
-  - "bundle exec rake db:create RAILS_ENV=test"
-  - "bundle exec rake db:migrate RAILS_ENV=test"
-  #- "bundle exec rake db:schema:load RAILS_ENV=test"
-  - "bundle exec rake db:seed RAILS_ENV=test"
+  - "bundle exec rake db:setup RAILS_ENV=test"
   - "sh -e /etc/init.d/xvfb start"
 script: "bundle exec rake travis"
 addons:
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,0 +1,7 @@
+connection = ActiveRecord::Base.connection()
+connection.execute <<SQL
+	ALTER TABLE votes
+	ADD CONSTRAINT vote_age_per_user_per_motion
+	UNIQUE USING INDEX vote_age_per_user_per_motion
+	DEFERRABLE INITIALLY IMMEDIATE
+SQL
--- a/spec/models/vote_spec.rb
+++ b/spec/models/vote_spec.rb
@@ -10,7 +10,7 @@ describe Vote do
     let(:vote) { Vote.create(user: user, motion: motion, position: "no") }
     subject { vote }
 
-    its(:age) { should == 0 }
+    its(:age) { is_expected.to eq 0 }
 
     context 'user changes their position' do
       let(:vote2) { Vote.create(user: user, motion: motion, position: "yes") }
@@ -20,15 +20,39 @@ describe Vote do
       end
 
       subject { vote2 }
-      its(:age) { should == 0 }
+      its(:age) { is_expected.to eq 0 }
 
       it "should age the first vote" do
         vote.reload
-        vote.age.should == 1
+        expect(vote.age).to eq 1
       end
 
       it 'the second vote should associate the first as the previous' do
-        vote2.previous_vote_id.should == vote.id
+        expect(vote2.previous_vote_id).to eq vote.id
+      end
+
+      context 'user changes their position again' do
+        let(:vote3) { Vote.create(user: user, motion: motion, position: "no") }
+        before do
+          vote3
+        end
+
+        subject { vote3 }
+        its(:age) { is_expected.to eq 0 }
+
+        it "should age the first vote" do
+          vote.reload
+          expect(vote.age).to eq  2
+        end
+
+        it "should age the second vote" do
+          vote2.reload
+          expect(vote2.age).to eq  1
+        end
+
+        it 'the second vote should associate the first as the previous' do
+          expect(vote3.previous_vote_id).to eq vote2.id
+        end
       end
     end
   end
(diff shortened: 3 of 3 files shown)


jakub-:
cc @robguthrie :)

robguthrie:
This is really great. Are you able to squash this into one commit? (see git merge --squash)

jakub-:
Thanks @robguthrie.

I'll be happy to do this to follow your conventions but I would like to take this opportunity to ask what the rationale for doing so is?

I've seen remarks that it somehow makes the history of the master branch cleaner but in principle that's a bit of a Subversion-minded approach. With Git, it's a good practice to have individual commits for logically distinct and independent changes and have the merge commit signify that they're part of the same chunk of work, i.e. a branch.

Just a few days ago in https://github.com/loomio/loomio/pull/1871#issuecomment-73180291 we were trying to track down why a particular change has been made. `git blame`, however, would not help with that - the commit is https://github.com/jakub-/loomio/commit/7c4a156f133c7fa22352e130138584e607754be1. While this may be an extreme case, I think that's an example where this approach can be rather harmful.

What do you think?

gdpelican:
Two reasons for me:

a: It makes the repository smaller to download. More commits = more
information to checkout the repo = higher barrier to entry for new devs,
especially on large projects with long histories (as this one is on the
verge of becoming). It's the kind of thing that's miniscule for a single
commit or branch, but over a long period of time of not being vigilant (and
having many different devs on a project), it can make a big difference. For
example, there was a brief period of time in our history where we
(accidentally?) added a bower_components folder of several tens of
thousands of lines, then removed it a few days later within the same
feature branch. When you checkout the repo, you have to download all of
that information. (Twice!)

b: It's also a lot easier for me as a dev hunting a regression to hop (or
for some, git bisect, what have you) through 10 features, rather than 100
commits. When I hear 'this started breaking two days ago' (or worse, 'I
don't know, it used to work?'), I want to narrow down the number of diffs I
need to look at to isolate the problem as much as possible.

It's certainly a matter up for debate (well, not in our repo, but on a
broader 'should this be an open source standard' sense); here's a couple
thoughts I found on the subject:
http://programmers.stackexchange.com/questions/263164/why-squash-git-commits-for-pull-requests
http://jamescooke.info/git-to-squash-or-not-to-squash.html

jakub-:
@gdpelican Thanks for the detailed response!

a) Notice that I'm not advocating for keeping commits in a state that reflects how my work evolved over time and that are mutually contradictory but rather commits that represent logically distinct changes. In that case, having those separate commits is barely contributing to the size of the repository. I'm all for cleaning up silly commits, including the example you've given.
b) `git bisect` will normally work on merge commits so you can in fact easily identify the merge commit (i.e. the PR) that introduced the change and proceed from there.

The discussions that you've pointed to in fact advocate for the same as I am – squashing silly commits like "added a line here", "oh, scratch that, let's remove it again", "dumdidum" but in fact do suggest it being okay for a PR to consist of more than one commit in total as is the case here.

robguthrie:
I don't take a very hard line on this stuff. That's because I don't have a
coherent answer, I don't feel super strongly about it, and I've been a
really messy incoherent commiter all my life.

These are my thoughts:
I don't think that number of commits in cases where you're not committing
something really unnecessary (eg: stuff that should never have been
committed) matter.

For me it's just about assembling this into "nice" pieces, and by that I'm
mostly thinking about git bisect case.. eg: if I take this piece off, does
the system still make sense.

It is nice if individual, complete features can be added as single commits,
but in reality this can only happen some of the time, and only if a single
person has worked on the feature - which is less common these days.

For small stuff like this, where you're fixing a bug, and it's unlikely
that the smaller commits are going to help someone in their understanding
of your work, and you are a single author, then it's my preference to
squash this.

For larger features, it's still nice to squash, but I tend to skip
squashing if multiple people collaborated to build the feature, as that's
the easiest way to not throw away the individual line information about who
wrote what.

I have a bash function which takes the current branch and squashes it into
a single commit against master, it's called squash-branch.

```

squash-branch() {

  name=`git rev-parse --abbrev-ref HEAD`

  echo "Squashing $name into one commit"

  g co master

  g co -b squashtmp

  g merge $name --squash

  g commit

  g br -D $name

  g br -m $name

}

```


I think in this case I would prefer it to be squashed, but I don't think
this is worth spending much time worrying about, and I think I'm happy for
you to do what you prefer, especially if it means you're more likely to
keep helping out.

jakub-:
@robguthrie Right. I also prefer not to think of squashing as a special operation, it's really history rewriting that's important. Where such a rewrite may involve an arbitrary sequence of commits being squashed, pulled out, split up and reworded.

Each of the three commits here are in fact fully functioning but also unrelated. They progressively improve the state of the project - the first one fixes the actual problem, the second one improves on the testing configuration of Travis by reverting an old change and the third one adds an actual test for *an unrelated problem* that is indirectly caused by what's being fixed. They're independent changes. I could, in fact, imagine creating three PRs for them and us merging them in in some order.

jakub-:
In the spirit of not letting this meta discussion take over this PR, I squashed the commits as suggested.

I do, however, believe that the quality of the history has strictly decreased in that it's now less clear the Travis change is in fact a revert of a previous one or that it could be made independently of this change altogether. It's also not clear why I added these specs and what they could possibly have to do with the change to `db/seeds.rb`. So I'd humbly encourage you to reconsider your position on the workflow. :)

Thanks!

jakub-:
@robguthrie Ah, sorry, your response got ill-formatted by GitHub so I didn't notice your closing remark. In that case, I will take you up on your offer and bring back the three commits, sorry! :-)

robguthrie:
Yikes, I guess I was pushing it trying to send that snippet via email. 
I totally agree with what you've chosen to do here. Thanks very much!

jakub-:
No probs, it's a pleasure contributing! Keep up the great effort. :)

Issue #1872, opened 2015-02-05
Require passwords to be sufficiently complex


jakub-:
Right now it looks like you can sign up with any password, including an empty one (in which case you won't be able to subsequently sign in... fun times!). I think a certain degree of complexity should be required.

There are off-the-shelf gems that make this very easy  such as [devise_security_extension](https://github.com/phatworx/devise_security_extension).

denjello:
I think that is a great idea - and probably helps to resolve [this trello card](https://trello.com/c/JysuTj4o/343-do-something-about-spam).

jakub-:
FYI, I will do this, just filing for the record. :)

denjello:
adding a the required flag to the registration form input field for password might also help.

`required='required'`

robguthrie:
This article is really great: http://blog.codinghorror.com/the-god-login/

restricting passwords to being a minimum length, and not in top 10,000 words seem like reasonable restrictions. 

I'll just say that I cannot stand rules about including particular characters etc - eg requiring a number or uppercase etc.

gdpelican:
+1 to avoiding stupid password rules.

denjello:
This is also a great resource: https://passfault.appspot.com/password_strength.html

For your reference, I just created a [gist with the 10000 most commons from 2014](https://gist.github.com/denjello/7e1c70ea77f3b9c6c711)

denjello:
And here is a shortened one using `awk 'length($0)>=8' 10000.txt > 2000.txt` because if the first check is `at least 8 characters long`: 
https://gist.github.com/denjello/d326051a643aa8521660

robguthrie:
Cool. I think the next most effective and least annoying thing we can do is
ensure that we have good rate limiting.

jakub-:
I think blacklisting that particular list is a great start and an improvement over what we have now but it would be nice to scale this solution beyond the Anglophone world. :)

Certainly +1 to avoiding annoying password rules.

robguthrie:
Just to be clear, the suggestion was to include a list of top 10,000 passwords (I don't know how it was collected, possibly from english websites), not english words.

See https://github.com/discourse/discourse/blob/master/lib/common_passwords/10k-common-passwords.txt

jakub-:
@robguthrie The point still stands, though, that the more generic password strength validators (like the one from Dropbox) would be language-agnostic. Looking through this list it's definitely skewed towards English-speaking websites. I think that's fine for now, though. :-)

[...]

PR #1877, opened 2015-02-07
Add basic password validation

jakub-:
Passwords are required to be at least 8 characters long.
In addition, a blacklist of ~2000 trivial passwords is introduced.


jakub-:
Partially addresses #1872 (rate limiting not yet in place).

cc @robguthrie

robguthrie:
This is a great start. Thank you.

I see this being potentially problematic when a user model fails password validation but the action the user is performing is not password related. 
It might be necessary to include this check only when updating password and creating a user account.

denjello:
@jakub - nice job for removing the passwords shorter than 8 characters.

jakub-:
@denjello Thanks!

@robguthrie This should be good to go, I think.

With regards to your suggestion to only perform this validation when necessary – that should already be the case. The `password` attribute is always nil unless it's being updated as the encrypted field has a different name altogether (`encrypted_password`).

jakub-:
Rebased.

robguthrie:
Thanks @jakub- I'll give this a smoke test today and merge it if all is well.

db/migrate/20150207203328_create_blacklisted_passwords.rb
```diff
@@ -0,0 +1,18 @@
...
+  def up
+    create_table :blacklisted_passwords do |t|
+      t.string :string
+
+      t.timestamps
+    end
+    add_index :blacklisted_passwords, :string, using: :hash
+
+    passwords = File.readlines(File.join(__dir__, '../password_blacklist.txt'))
+      .map(&:strip)
```
gdpelican: Can we squishy this down to 

```
.map { |string| string.strip }
```
?

gdpelican:
This looks great, thanks @jakub-.

If you're interested in putting in some rate limiting next, have a gander at the closed #1390, which was an early attempt at some rate limiting which never made it into master.

db/migrate/20150207203328_create_blacklisted_passwords.rb
jakub-: We can and I did!

It's certainly a matter of taste but I myself prefer keeping distinct transformations separate as it makes it easier to see what it is that each of them does individually rather than having them all squeezed together. But that's me! :)

gdpelican:
Thanks @jakub-! :boat:

PR #1878, opened 2015-02-08
disregard subgroups and archived groups

rdbartlett:


gdpelican:
AW HECK YEAH :facepunch:

gdpelican:
Not a whole lot to nit pick here; can we merge this?

PR #1888, opened 2015-02-10
Link up notifications to urls

gdpelican:
Hook up notifications links

It'll be cool to revisit this after pagination is done, so we can look at how to make the anchors work properly.

:us:


gdpelican:
Going to merge this in and revisit after notification revamp is complete.

Issue #1853, opened 2015-01-26
use_gravatar Naming Error in locales


denjello:
How did no one ever see this?
```
    /config/locales/en.yml
    use_gravtar: 'Use Gravatar'

    /config/locales/de.yml
    use_gravtar: Gravatar benutzen


```

gdpelican:
Good catch!

@s01us, what's the best way to update a translation key without losing the
translations?

How did no one ever see this?

    /config/locales/en.yml
    use_gravtar: 'Use Gravatar'

    /config/locales/de.yml
    use_gravtar: Gravatar benutzen



—
Reply to this email directly or view it on GitHub
<https://github.com/loomio/loomio/issues/1853>.

s01us:
Oh no. I think this was my doing. ironically i was making this more
translatable but uses a typo.

Changing keys is a pain, you would have to edit transifex. My suggestions
are we can either eat the typo (add a line to deal with it), or just
correct the key and it will have to be translated by everyone again.

gdpelican:
Hm. I think this particular bit of code is a little gnarly to jam a typo into; I don't have a great sense for how much of an ask it is to have that string translated again across languages, but I think I prefer that approach.

denjello:
This is on the transifex website:

As long as your source file is accessible from a public URL, such as [this raw file](https://github.com/transifex/txc/raw/master/transifex/locale/en/LC_MESSAGES/django.po) on GitHub, Transifex will periodically check the URL for updates. You can also use a file from a private repo in Github as long as the URL has the respective token that Github provides. If there are updates in the file, Transifex automatically fetches the file and updates your content.

You can find this option under Project Details → Resources → Auto-update Resources.

It is important to note this solution depends on Transifex polling for updates. As a result, there will a delay between the moment you update your file and Transifex picking up the changes. If you would rather update your content in Transifex as soon as your files change, use the client or API instead.

Here’s how differences in the content will be handled:

    New strings will be added
    Modified strings will be considered as new ones and added
    Strings that are not in the new source file (including ones which have been modified) will be removed along with their translations.

PR #1889, opened 2015-02-10
[#1853] Use gravtar typo

marceloboeira:
Would this solve? Or there is some stuff on Transfirex ?

### DIFF
--- a/config/locales/an.yml
+++ b/config/locales/an.yml
@@ -748,7 +748,7 @@ an:
     picture: Imachen
     use_initials: Usar inicials
     use_uploaded: Puyar nueva imachen
-    use_gravtar: Usar Gravatar
+    use_gravatar: Usar Gravatar
     upload_new_image: Cargar nueva imachen
   group_setup:
     set_up_your_group: Establir grupo
--- a/config/locales/be-BY.yml
+++ b/config/locales/be-BY.yml
@@ -748,7 +748,7 @@ be-BY:
     picture: Фота
     use_initials: Выкарыстоўваць ініцыялы
     use_uploaded: Выкарыстоўваць запампаваны здымак
-    use_gravtar: Выкарыстоўваць Gravatar
+    use_gravatar: Выкарыстоўваць Gravatar
     upload_new_image: Загрузіць новы здымак
   group_setup:
     set_up_your_group: Стварэнне новай групы
--- a/config/locales/cs.yml
+++ b/config/locales/cs.yml
@@ -630,7 +630,7 @@ cs:
     picture: Fotka
     use_initials: Použít iniciály
     use_uploaded: Použít nahraný obrázek
-    use_gravtar: Použít Gravatar
+    use_gravatar: Použít Gravatar
     upload_new_image: Nahrát novou fotku
   group_setup:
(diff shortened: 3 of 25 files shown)


gdpelican:
Hey @marceloboeira, thanks for the PR!

This is my preferred fix, but interested in hearing what @s01us has to say on the matter. Can we one-off script a key change for transifex somehow?

marceloboeira:
@gdpelican branch is failing but I believe to be not related with my changes.

denjello:
@marceloboeira You are right. I am having the same issue on another pull request.
```
(::) failed steps (::)
Circular dependency detected while autoloading constant NotificationsController (RuntimeError)
features/admin/unarchive_group.feature:12:in `Then they should be able to view their group page'
Failing Scenarios:
cucumber features/admin/unarchive_group.feature:8 # Scenario: Loomio admin unarchives group
```

gdpelican:
It's an intermittent failure, I'll have a peek at why it's happening at
some point in the near future.

denjello:
So what is the solution?

s01us:
Changing one key everywhere requires some research.
I'll have a look if i have time tmrw.

s01us:
Oh, to be honest, it's a typo in the key and this is used in one place in the app. I'm not keen to spend time editing this. 
There are more pressing translation questions, like offering users approriate translations reliabily on the front page. 

Basically we would have to upload this change to transifex for every file. Not hard, just takes time that i don't have

marceloboeira:
@s01us I already searched for the translation, and replaced everywhere ...

denjello:
@marceloboeira - Great, its in the pull-request, unfortunately the larger scope of the issue is getting those changes back into transifex - which seems to be a relatively labor-intensive task.

gdpelican:
The other solution is to eat the typo and just put `t(:user_gravtar)` in the code, but it's a little tricky because we're switching on a variable in a partial.

If it was as simple as
**profile.html.haml**:
```
 = render 'avatar_setting_option', user: @user, kind: 'gravtar'
```

I might consider it, but any more work than that and it's just a hack that we'll have to spend more time on later.

denjello:
As a professional translator I hate typos. As a coder I use terms to describe things. If that thing is already described and has a name, then when I want to use the word for the thing that exists I always spell it correctly, because otherwise I am not naming that thing. My opinion is fix the incoming descriptor:

`gravtar` > `gravatar`

Who knows where it might be needed later in further iterations of the code, and if every new coder has to read in the wiki "by the way, we misspelled gravatar". we're basically just propagating incompetence.

gdpelican:
I appreciate your point @denjello, but this bug isn't high profile enough to warrant the Loomio team members spending any more than a trivial amount of time on it. 

I'm happy to look at a simple pull request to fix the problem if it can be done with a pull request, but I'm not happy ask @s01us to spend a bunch of time fixing something this minor. I'm also very happy to leave it as is until such time as we can get to a proper fix (or, until this page gets phased out of the interface)

gdpelican:
Closing this for now, can reopen later if we find ourselves willing to spend some time on updating transifex to match this PR.

PR #1897, opened 2015-02-11
Simple Fix to parse HTML in Search Results

### LINKED ISSUE #1881: Search Results in Beta have unparsed markdown.
search

I see that this is not really the case in 1.0 - but think that whatever is shown should be parsed correctly and Beta is still going to be around for a while...

Here is my solution in `app/views/searches/_discussion_line_item.html.haml`
```
      %span.description= truncate(render_rich_text(item.description,item.uses_markdown), length: 180, omission: '...', separator: " ", :escape => false) unless item.description.blank?
```

denjello:
This resolves #1881 for me. Love to hear how it works for you.


s01us:
this has been cherry-picked into : https://github.com/loomio/loomio/pull/1936

Issue #1858, opened 2015-01-29
Language switcher default from browser language?


peteruithoven:
Could it be that the language selector on the bottom of for example the sign up form always defaults to English? 
Could this use the language set in the browser settings?

denjello:
Strangely the "best_cachable_locale" (which is also spelled incorrectly btw.) is to blame here I believe.

in `app/helpers/locales_helper.rb` it reads: 
```
  def best_locale
    if user_signed_in?
      best_available_locale
    else
      best_cachable_locale
    end
  end

  def best_available_locale
    selected_locale || detected_locale || default_locale
  end

  def best_cachable_locale
    selected_locale || default_locale
  end
```

This change worked for me as a quick fix:
```
  def best_cachable_locale
    selected_locale || detected_locale || default_locale
  end
```

I assume that this was done to improve anonymous-user  performance, but I find it to be a bit of an anti-pattern. Which is why I would recommend simplifying the first snippet into the following:

```
  def best_locale
    selected_locale || detected_locale || default_locale
  end
```

What do you think?

denjello:
To make another comment about this language switcher, it is mysteriously missing from the landing page when not logged in.

denjello:
@gdpelican -> close?

PR #1898, opened 2015-02-11
Add detected locale for non-signed in users.

denjello:
This is a fix for not catching the locale of users that aren't signed in as discussed here: #1858

### DIFF
--- a/app/helpers/locales_helper.rb
+++ b/app/helpers/locales_helper.rb
@@ -88,19 +88,7 @@ module LocalesHelper
   end
 
   def best_locale
-    if user_signed_in?
-      best_available_locale
-    else
-      best_cachable_locale
-    end
-  end
-
-  def best_available_locale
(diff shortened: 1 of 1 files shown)


gdpelican:
Hm, cool. @s01us, how does this fix sit with you?

robguthrie:
This is stuff I wrote. Basically the cacheable thing is no longer a requirement - we can take out best_cachable_locale all together, which means that you can take out best_available_locale too.. and just put the logic directly into best_locale

denjello:
@robguthrie - yeah that's what i recommended two weeks ago.
```
  def best_locale
    selected_locale || detected_locale || default_locale
  end
```

robguthrie:
Oh OK cool. Go ahead with that then.

denjello:
@robguthrie - btw. its ready to pull.

robguthrie:
Thanks @denjello

Issue #1882, opened 2015-02-09
admin created categories are not translatable


denjello:
To make this happen create a category at `/admin/categories/new` in this case `Off Topic`

and visit the page `/explore`

![group](https://cloud.githubusercontent.com/assets/6129121/6109653/a290ac80-b07c-11e4-90e1-92db136900be.png)

Here is a fix:

in `app/controllers/explore_controller.rb`
```
  def category_names_and_paths
    @categories = Category.all
    names_and_paths = @categories.map{|c| [I18n.t(c.translatable_name, default: c.name), category_explore_path(c)]}

    names_and_paths.prepend [t(:"group_categories.all"), category_explore_path(:all)]
    names_and_paths.prepend ["", explore_path]
  end
```

gdpelican:
Er, we're not going to provide translations for off-instance admin-created categories. 

You should be able to provide your own translations by editing the `en.yml` file (as well as however many other languages you'd care to support) :family:

denjello:
Did you look at my proposal? All I added was a change so that if no translation is available it shows the original string.

robguthrie:
Please have another look at this @gdpelican. I think you've missed the point. This looks like a neat, general purpose pull request.

PR #1900, opened 2015-02-11
Admin created categories are never translated.

denjello:
Simple fix to add a default to the translating system so that this behaviour never manifests.
see  #1882


s01us:
nice one @denjello

robguthrie:
Thank you @denjello!

PR #1905, opened 2015-02-12
work around for the fact postgres does not implement constraint checking...

robguthrie:
... properly

### DIFF
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -29,7 +29,7 @@ class Vote < ActiveRecord::Base
 
   before_create :age_previous_votes, :associate_previous_vote
 
-  after_save :update_motion_vote_counts
+  after_create :update_motion_vote_counts
   after_destroy :update_motion_vote_counts
 
   alias :author= :user=
@@ -90,12 +90,13 @@ class Vote < ActiveRecord::Base
   end
 
   def age_previous_votes
-    #Vote.transaction do
-      #Vote.where(motion_id: motion_id).order('age desc').each do |vote|
-        #vote.update_attribute(:age, vote.age + 1)
-      #end
-    #end
-    motion.votes.where(user_id: user_id).update_all('age = age + 1')
+    raise "only for new votes" if persisted?
+    Vote.transaction do
+      motion.votes.where(user_id: user_id).order('age desc').each do |vote|
+        vote.update_attribute(:age, vote.age + 1)
+      end
+    end
+    #motion.votes.where(user_id: user_id).update_all('age = age + 1')
   end
 
   def update_motion_vote_counts
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,0 @@
-connection = ActiveRecord::Base.connection()
-connection.execute <<SQL
-	ALTER TABLE votes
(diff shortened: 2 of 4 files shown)


robguthrie:
@gdpelican If you could give this whatever kind of check you can, asap, that would be great.

This essentially works around the postgres wontfix, and does the update manually within a transaction. Rather than trying to set the deferred thing, which rails does not support reliably.

gdpelican:
I'll give it a onceover now.

gdpelican:
This looks totally fine to me, and I trust the test. Mergin'! :hand:

jakub-:
@robguthrie Hey, if you don't mind some feedback — as this change reverts a previous one of mine, it'd have been nice if I had been CCed on it.

Just so that we can all learn, could you briefly elaborate what problem with the previous change this on is solving? I'm curious what you mean by "the deferred thing, which rails does not support reliably."

robguthrie:
@jakub- Yeah, of course! Thanks for the prompt... still getting use to the idea you guys are out there, sorry! And happy to explain.

So after your patch to do with INITALLY DEFERRED on the votes table, we've started to see the bug you were reporting in the first place - constraint violation exceptions when updating votes.

I had a go at fixing the constraint parameters, but in the end felt that it was fiddly and unreliable, seeing as you we're having problems when we we not, then vice versa. 

Just for a refresh, and for other readers, this is the problem with PostgreSQL:

> When a UNIQUE or PRIMARY KEY constraint is not deferrable, PostgreSQL checks for uniqueness immediately whenever a row is inserted or modified. The SQL standard says that uniqueness should be enforced only at the end of the statement; this makes a difference when, for example, a single command updates multiple key values. To obtain standard-compliant behavior, declare the constraint as DEFERRABLE but not deferred (i.e., INITIALLY IMMEDIATE). Be aware that this can be significantly slower than immediate uniqueness checking.

My expectation when I first wrote that UPDATE statement (setting age = age + 1) was that PSQL would behave as the standard expects... we've all since found out that it does not do this.

My reasoning in the end for this approach, was that since Postgres is weird in this case, and it seems like all of us are having trouble getting the constraints setup correctly, and because Rails does not have proper support for carrying this setting via schema.db, and finally because the update itself is not performance critical or worth spending a lot of time upon..... I decided to change the method to update each of the other votes manually, in the correct order, so that the constraint can be simpler, and we can have reliability without this fiddly stuff.

jakub-:
@robguthrie Thanks for the detailed explanation, appreciated!

Were you seeing this behaviour on new Loomio instances? Any chance the database had been initialised with `db:create` & `db:schema:load` rather than `db:setup`?

Sorry, I don't want you to waste any more time on this and I think your fix is a good compromise. It's slightly less future proof though as we'll have to be wary of the same issue any time we do a mass-update of similarly constrained fields.

robguthrie:
No we we're seeing it on loomio.org, it had the migration applied to it, then things started going wrong.

Locally, I was seeing test failures using both db:setup and db:create, db:migrate... it was becoming a nightmare and in the end, this detail does not matter as much as more feature development.

Obviously I would have preferred the single update statement. It would be easy to switch back to single update statement if you feel super confident that we can reliably configure the constraint.

PR #1907, opened 2015-02-13
update invitation+added_to_group emails to use group.full_name fixes#1865

s01us:
we've QA'd this as well @HSalmon and I

### DIFF
--- a/spec/mailers/user_mailer_spec.rb
+++ b/spec/mailers/user_mailer_spec.rb
@@ -44,6 +44,10 @@ describe UserMailer do
     it 'renders the subject' do
       expect(@mail.subject).to eq "#{@inviter.name} has added you to #{@group.full_name}"
     end
+
+    it 'uses group.full_name in the email body' do
+      expect(@mail.body.encoded).to  include @group.full_name
+    end
   end
 
 end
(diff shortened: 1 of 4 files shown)


spec/mailers/user_mailer_spec.rb
```diff
@@ -32,4 +32,24 @@
       @mail.body.encoded.should match("http://localhost:3000/g/#{@group.key}")
     end
   end
+
+  context '#added_to_group' do 
+    before do
+      @user    = double(:user, name: 'Wayne G', email: 'w_recipient@g.com', name_and_email: nil)
```
gdpelican: Rob and I are both moving away from using doubles in the test suite; we're willing to sacrifice a tiny bit of speed for the knowledge that we're actually testing how the application works, instead of the construct we've imagined.

Also, the `.should` syntax, while rampant in our test suite, is less than ideal ([And rspec themselves agree](https://github.com/rspec/rspec-expectations/blob/master/Should.md)). Please use `expect(...).to eq` instead. :+1:

s01us:
@gdpelican  i sorted the merge conflict
:tropical_fish:

gdpelican:
Looks like a winner, thanks @s01us

PR #1908, opened 2015-02-13
add Follow action to emails. Fixes #1711

### LINKED ISSUE #1711: Should be able to follow a thread from email.
Currently when you receive any emails from a followed thread, there's a link to unfollow the thread, directly from the email. For emails about unfollowed threads, it would be great to add a follow link in the same way. The email_actions controller is all ready and waiting for this extension.

s01us:
controller test
no cuke 
no qa yet 

does need ? 

forgot charger ... in hurry

### DIFF
--- a/spec/controllers/email_actions_controller_spec.rb
+++ b/spec/controllers/email_actions_controller_spec.rb
@@ -19,6 +19,24 @@ describe EmailActionsController do
     end
   end
 
+  describe "follow_discussion" do
+    before do
+      @user = FactoryGirl.create(:user)
+      @group = FactoryGirl.create(:group)
+      @group.add_member!(@user)
+
+      @discussion = FactoryGirl.build(:discussion, group: @group)
+      DiscussionService.create(discussion: @discussion, actor: @user)
+      DiscussionReader.for(discussion: @discussion, user: @user).unfollow!
+    end
+
+    it 'follows the discussion' do
+      expect(DiscussionReader.for(discussion: @discussion, user: @user).following?).to eq false 
+      get :follow_discussion, discussion_id: @discussion.id, unsubscribe_token: @user.unsubscribe_token
+      expect(DiscussionReader.for(discussion: @discussion, user: @user).following?).to eq true
+    end
+  end
+
   describe "mark_discussion_as_read" do
     before do
       @user = FactoryGirl.create(:user)
--- a/app/controllers/email_actions_controller.rb
+++ b/app/controllers/email_actions_controller.rb
@@ -2,7 +2,13 @@ class EmailActionsController < AuthenticateByUnsubscribeTokenController
   def unfollow_discussion
     discussion = Discussion.find(params[:discussion_id])
     DiscussionReader.for(discussion: discussion, user: user).unfollow!
-    redirect_to dashboard_or_root_path, notice: "You are no longer following the discussion: #{discussion.title}"
+    redirect_to dashboard_or_root_path, notice: t(:'notifications.email_actions.not_following_thread', thread_title: discussion.title)
+  end
+
+  def follow_discussion
+    discussion = Discussion.find(params[:discussion_id])
+    DiscussionReader.for(discussion: discussion, user: user).follow!
+    redirect_to dashboard_or_root_path, notice: t(:'notifications.email_actions.following_thread', thread_title: discussion.title)
   end
 
   def mark_discussion_as_read
--- a/app/views/thread_mailer/_footer.html.haml
+++ b/app/views/thread_mailer/_footer.html.haml
@@ -8,10 +8,13 @@
   %br
   = t(:'email.thread_mailer.reply_or_view_online_html', url: link)
 
+  %br
   - if @following
-    %br
     - unfollow_url = email_actions_unfollow_discussion_url(discussion_id: @discussion.id, unsubscribe_token: @recipient.unsubscribe_token)
     = t :'email.thread_mailer.unfollow_discussion_html', unfollow_url: unfollow_url
+  - else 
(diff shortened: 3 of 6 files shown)


spec/controllers/email_actions_controller_spec.rb
```diff
@@ -19,6 +19,24 @@
     end
   end
 
+  describe "follow_discussion" do
+    before do
+      @user = FactoryGirl.create(:user)
```
gdpelican: I don't think you need to call `FactoryGirl.` here, it can just be

```
@user = create(:user)
```

Not sure why the current tests are using this format, but it's less than great form.

spec/controllers/email_actions_controller_spec.rb
```diff
@@ -19,6 +19,24 @@
...
+      @group = FactoryGirl.create(:group)
+      @group.add_member!(@user)
+
+      @discussion = FactoryGirl.build(:discussion, group: @group)
+      DiscussionService.create(discussion: @discussion, actor: @user)
+      DiscussionReader.for(discussion: @discussion, user: @user).unfollow!
+    end
+
+    it 'follows the discussion' do
+      DiscussionReader.for(discussion: @discussion, user: @user).following?.should be false
```
gdpelican:
```
expect(DiscussionReader.for(discussion: @discussion, user: @user).following?).to be_false
```

Hm, this `.should` thing's popped up frequently enough recently that I'm thinking we might need to put some prominent wording in the 'devving on Loomio' wiki around it.

gdpelican:
This looks great. I'll QA it at some point early next week; I think a cuke's probably slight overkill, but happy to be wrong.

spec/controllers/email_actions_controller_spec.rb
s01us: The should thing is mainly people following examples in surrounding tests i
think.
Perhaps i could try writing a script which brings all our tests up to spec

robguthrie: it'd be faster to just manually change them. I'm sure

spec/controllers/email_actions_controller_spec.rb
robguthrie: Actually, I prefer the FactoryGirl format.. It makes it really clear where the create method comes from.

spec/controllers/email_actions_controller_spec.rb
s01us: @gdpelican heads up that suggestion doesn't work. it generates 
```ruby
false.false?
```
and there's no `false?` defined for false.

```ruby
expect(DiscussionReader.for(discussion: @discussion, user: @user).following?).to eq false
``` 
works

gdpelican:
Hey this looks pretty good. A couple notes:

- I can't seem to see this link show up on plaintext emails; there might be a different template for them.
- Maybe out of scope for this PR, but a confusing thing happens when you click the link from an email intended for a different user. It goes:
  - Bob is signed into Loomio as bob@bob.com
  - He gets an email at bob@tom.com, inviting him to follow / unfollow a discussion. He clicks it.
  - He ends up on Loomio.org, signed in as bob@bob.com. It says 'You're following this discussion now!
  - But bob@bob.com not following the discussion, another user, bob@tom.com (the email recipient) is.

Not a huge deal probably, but it confused the crap out of me for a little while. We likely want a check to make sure the user that's signed in is the one coming from the email before we do any following.

robguthrie:
@gdpelican the "logged in as someone and using a follow link built for someone else" issue is kind of un avoidable as I understand. Do you want something done about it, or have a suggested change to make that better?

Note that we include identify in the follow link so that links will work when people are not logged in.

gdpelican:
We should be able to do some token-based auth for signing in / detecting if a user is signed in as another user. But it's way out of scope for this fix.

@s01us how would you feel about fixing the plaintext version? After that I think this is well ready to be merged.

rdbartlett:
I was thinking about a pop-up like "hey you came from an email to
rob@loomio.org but you're logged in as mix@loomio.org - do you want to log
out or continue as mix?"

robguthrie:
Id rather we require people to be logged in, and don't carry user identifiers in these links.

Moreover, I struggle to understand why someone is using the follow links on another persons emails. It just does not seem realistic to me.

robguthrie:
Hey @s01us and @gdpelican this is a sweet article reminding us all of the benefits of using crypto to manage stuff like this very issue. Please have a read of it and see how you think it might apply to us, generating tokens for users in the future.

https://neosmart.net/blog/2015/using-hmac-signatures-to-avoid-database-writes/

gdpelican:
Hey I'm going to open a new issue for this, because it's only tangentially
related to following via email. (And I don't think it should hold up this
PR.)

denjello:
@robguthrie > you're welcome. <https://www.loomio.org/comments/509619/> :v:

robguthrie:
Arr nice!
Do you have any thoughts about what values we would hash together to
produce links suitable for email actions?

denjello:
Sure:
 link to the profile of the person that wrote the comment or added you to the group. i.e.
```
<a href="https://www.loomio.org/u/hmbZxt1M/linus">
Linus Torvalds
</a> has added you to Roasting Kernels on Loomio.
```
I find it incredibly hard to find users, and this might help.

And this is not exactly an answer to your question, but the request following request came up in our "loomio integration meeting today": For users that only follow "what happened in the last 24h", it might be nice to have an option for them to get the whole thread mailed - because they are more likely to be the type that very seldom are to be found using the website in and of itself.

gdpelican:
Added plaintext follow / unfollow links; mergin' this in.

s01us:
Hey, sorry I was too busy to engage with this sooner

Issue #1647, opened 2014-08-26
Moving Discussions buggy


s01us:
after creating, I went into the "edit" for a discussion 
(https://www.loomio.org/d/T4QXGbgI/mikey-is-leaving-let-s-say-goodbye)

and tried to move it by selecting another group in the dropdown then clicking update.

i got a jolly yellow updated flash, but the discussion wasn't moved.

turns out I'd been offered some 'illegal' destination groups ...

rdbartlett:
@gdpelican this was just reported again in the Enspiral group [see here](https://www.loomio.org/d/YmF9wTPq/what-can-we-create-in-auckland#comment-473180) if you have access.

gdpelican:
I'm with @robguthrie on this one that we just need to relax the permissions, but there probably is a fix which matches the moving the discussion to what the frontend says you're able to do.

Happy to support someone with a fix on this one.

rdbartlett:
Cool. I would prefer an immediate fix because the UX sucks (the list offers
you illegal groups, then the flash messages tells you it worked, but it
doesn't) and then a redesign.

On 29 January 2015 at 12:30, James Kiesel <notifications@github.com> wrote:

> I'm with @robguthrie <https://github.com/robguthrie> on this one that we
> just need to relax the permissions, but there probably is a fix which
> matches the moving the discussion to what the frontend says you're able to
> do.
>
> Happy to support someone with a fix on this one.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/loomio/loomio/issues/1647#issuecomment-71939357>.
>



-- 

Richard D. Bartlett
Loomio co-founder
rich@loomio.org

<http://loomio.org/>

gdpelican:
Can't prioritize it over 1.0 sadly. Happy to support an open source
contributor for a fix though.

PR #1916, opened 2015-02-17
Fix move discussion for #1647

gdpelican:
This should fix the user being able to move discussions they're not able to.

Before, the UI would allow the author of a discussion to attempt to move a discussion, but it would fail on submit.

The logic (as I understand it should be): If a user is an admin of a group, they can move any discussion in that group to any other group they are a member of.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -154,8 +154,7 @@ class Ability
       end
     end
 
-    can [:destroy,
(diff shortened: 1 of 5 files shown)


gdpelican:
I fixed it! @robguthrie @HSalmon

robguthrie:
Thank you!!

PR #1923, opened 2015-02-18
redirect to dashboard at controller level

robguthrie:
We currently do this with JS in the marketing index. It's not necessary to do that because we don't do app level caching now. Redirect based on cookies will prevent the "flash of marketing page before redirect" issue we've put up with for ages.


gdpelican:
+10

PR #1924, opened 2015-02-18
translations table: remove unsed,  error prone idx

robguthrie:
and replace with hopefully optimising index on association fields


robguthrie:
@gdpelican what do ya think?

PR #1925, opened 2015-02-18
makes specs run quieter by changing .css.scss to .scss as asked

s01us:


gdpelican:
Hey @s01us, @robguthrie is doing a rebase of the master branch at the moment to cut down our repo size, which means PRs will need to be rebased against master again.

Should be quick and painless, but would be super curious to hear how your experience of doing it was.

robguthrie:
Yea sorry for the hassle, but I'm still working on what the correct
commands are.. hang on if you like and I'll get back to you.

PR #1929, opened 2015-02-19
Remove subgroup permissions option to create subgroups

### LINKED ISSUE #1637: Option to allow creation of subgroups within subgroups is confusing
Currently when setting up a new subgroup, you are shown the permissions tab, including the option to create subgroups. This doesn't make any sense in the context of setting up a subgroup, since you can't have sub-subgroups. Suggest removing this option when creating a subgroup to avoid confusion.

screen shot 2014-08-14 at 2 58 00 pm

HSalmon:
Fixes #1637


gdpelican:
Looks like a winner. I'll wait for the spec to pass and then merge. :hamster:

PR #1932, opened 2015-02-19
Generalize index action

gdpelican:
I paired with my good friend Jim Beem on this one, but I still reckon it turned out pretty well. :tropical_drink:


robguthrie:
That's nice!!

PR #1933, opened 2015-02-19
add transaction around membership check and creation

robguthrie:
We have a few places like this where the code is not threadsafe, but we run with threading enabled (indeed loomio.org would need a lot more resources if we did not)

The trick is, if there is a place where you are reading a value, then in another statement writing based on that result, the read and write need to be contained in a transaction so that the value does not get changed by another thread inbetween the read and the write.

In this case I think there is even a rails helper for find_or_create but I kinda remembered that being problematic when I wrote this ages ago, and this makes it super clear we are using a transaction.

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -373,10 +373,12 @@ class Group < ActiveRecord::Base
   end
 
   def find_or_create_membership(user, inviter)
-    membership = memberships.where(user_id: user.id).first
-    membership ||= Membership.create!(group: self,
-                                      user: user,
-                                      inviter: inviter)
+    Membership.transaction do
+      membership = memberships.where(user_id: user.id).first
+      membership ||= Membership.create!(group: self,
+                                        user: user,
+                                        inviter: inviter)
+    end
   end
 
   def user_membership_or_request_exists? user


robguthrie:
I'm going on a small rampage to find these over the next few days.

gdpelican:
It'd be nice to be able to go
```
def find_or_create_thing
  ...
end
transaction_method :find_or_create_thing
```

Similar to `handle_asynchronously`.

robguthrie:
I'm not sure I agree on the second option. If we call a transaction wrapping method at class level it means the code is harder to reason about, because part of what's going on is spatially separated from the rest. For me, It's easier to read and understand if it's all together. Especially when it's crucial to it working or not (we got an exception about this very method this morning). Maybe I'll change my mind if there is lots of these... can we wait and see on this?

Like I said, there is a rails method for find_or_create which will surely have a transaction within it, and I'm happy for someone to refactor this to use that, but this PR is mostly just about making dev's aware of what the cases where we need transactions look like.

gdpelican:
Happy to leave it for now, but I think I might write a `handle_in_transaction` gem, as that's something I'd use all over the place. I think the language makes it pretty self-explanatory what's going on.

gdpelican:
www.github.com/gdpelican/handle_in_transaction :monkey:

PR #1899, opened 2015-02-11
Buttons for Markdown with logic for inserting markup into textfield.

denjello:
For the record, here is code I am using in production for allowing users to insert markdown tags into the textfield in comments.

See the discussion at issue #1856 and over at [loomio](https://www.loomio.org/d/ag9nP5hG/markdown-love-it?proposal=SH04sUwe)

PR #1937, opened 2015-02-20
Buttons for Markdown with logic for inserting markup into textfield.

s01us:
cherry picked 2 commits + squashed  #1899


gdpelican:
Closing this; I think we're moving away from markdown support, and don't want to introduce an editing interface at this stage.

PR #1941, opened 2015-02-22
fix url_options exception

robguthrie:


gdpelican:
Whoa, why are you ripping out all of these tests for a tiny fix like this?

gdpelican:
And why include schema changes and a deploy script update?

robguthrie:
The tests are isolated.. and don't really do anything.

The schema changes I have no excuse for..

PR #1950, opened 2015-02-25
Reader volume

gdpelican:
Not for merging yet, but a solid first pass at converting the idea of 'following' to 'volume'


robguthrie:
Looks like a great start @gdpelican I'm super pleased with the direction it's taking.

robguthrie:
you might need to rebase off base though.. looks like my merge sqashing as confused your branch and you're trying to put in some old code.

app/models/discussion.rb
```diff
@@ -77,7 +77,7 @@ def followers
       active.
       joins("LEFT OUTER JOIN discussion_readers dr ON (dr.user_id = users.id AND dr.discussion_id = #{id})").
       joins("LEFT OUTER JOIN memberships m ON (m.user_id = users.id AND m.group_id = #{group_id})").
-      where('dr.following = TRUE OR (dr.following IS NULL AND m.following_by_default = TRUE)')
+      where('dr.volume = :email OR (dr.volume IS NULL AND m.volume = :email)', { email: DiscussionReader.volumes[:email] })
```
robguthrie: Are you super sure you want to call membership.volume that? I feel like it might make it clearer if its default_volume on membership

gdpelican:
Yeah, trying to patch that stuff up now.

gdpelican:
@robguthrie this is ready for a smoke testing.

gdpelican:
@robguthrie I added serializing out permitted params on here, which seems to work swimmingly. Also fixed the create new discussion bug by filtering out discussions without an id from the inbox.

(The bug was that we were calling `activeProposal()` -> `proposals()` on a discussion which had not been created yet, which means it's `setupViews()` hadn't been called yet, which means b0rkdom.)

app/models/discussion_reader.rb
```diff
@@ -19,20 +20,14 @@ def self.for(user: nil, discussion: nil)
     end
   end
 
-  def follow!
-    update_attribute(:following, true)
+  def set_volume_as_required!
+    if user.email_on_participation?
```
gdpelican: Nested unless / if :(

```
if user.email_on_participation? && !volume_is_email?
  set_volume! :email
end
```

app/models/motion.rb
```diff
@@ -226,11 +214,11 @@ def update_vote_counts!
   end
 
   def group_members_without_motion_author
-    group.users.where(User.arel_table[:id].not_eq(author.id))
+    group.members.without(author)
```
gdpelican: These are cool.

app/models/user.rb
```diff
@@ -114,17 +114,35 @@ class User < ActiveRecord::Base
...
+    where('memberships.email_new_discussions_and_proposals = ?', true) }
   scope :email_new_proposals_for, -> (group) { active.email_new_discussions_for(group) }
 
+  scope :without, -> (users) {
+    users = Array(users).compact
+
+    if users.size > 0
+      where('users.id NOT IN (?)', users)
+    else
+      all
```
gdpelican: I think you want `scoped` here, so it'll handle default scopes correctly.

app/queries/users_by_thread_volume_query.rb
```diff
@@ -0,0 +1,22 @@
+class UsersByThreadVolumeQuery
```
gdpelican: Any reason we're ditching the `Queries::UsersByThreadVolume` idiom for these?

app/queries/users_by_thread_volume_query.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def self.mute(discussion)
+    volume_and_discussion(:mute, discussion)
+  end
+
+  private
+  def self.volume_and_discussion(volume, discussion)
+    User.
+      active.
+      joins("LEFT OUTER JOIN discussion_readers dr ON (dr.user_id = users.id AND dr.discussion_id = #{discussion.id})").
```
gdpelican: Probably fine, but just to be more safest, `... dr.discussion_id = ?", discussion.id`

app/queries/users_to_email_query.rb
```diff
@@ -0,0 +1,40 @@
+class UsersToEmailQuery
+  def self.new_comment(comment)
+    discussion = comment.discussion
+
+    UsersByThreadVolumeQuery.email(discussion).
```
gdpelican: `.email(comment.discussion)` ?

app/views/thread_mailer/_footer.html.haml
```diff
@@ -11,10 +11,10 @@
   %br
   - if @following
     - unfollow_url = email_actions_unfollow_discussion_url(discussion_id: @discussion.id, unsubscribe_token: @recipient.unsubscribe_token)
-    = t :'email.thread_mailer.unfollow_discussion_html', unfollow_url: unfollow_url
-  - else 
+    = t :'email.thread_mailer.turn_off_notifications_html', url: unfollow_url
+  - else
     - follow_url = email_actions_follow_discussion_url(discussion_id: @discussion.id, unsubscribe_token: @recipient.unsubscribe_token)
-    = t :'email.thread_mailer.follow_discussion_html', follow_url: follow_url
+    = t :'email.thread_mailer.turn_on_notifications_html', follow_url: follow_url
```
gdpelican: follow_url -> url here as well?

app/queries/users_by_thread_volume_query.rb
robguthrie: rails forces its naming conventions hard. moving queries into app means you namespace on 2nd level depth. So Queries:: is basically wrong.. and it has to live in app/extras/queries just to work as is.

app/queries/users_by_thread_volume_query.rb
robguthrie: you can't do that in a join :(

app/views/thread_mailer/_footer.html.haml
robguthrie: oh, thanks

app/models/user.rb
robguthrie: scoped was deprecated in rails 4.0

PR #1952, opened 2015-02-25
Example discussion for top-level groups created in app

### LINKED ISSUE #1700: No example discussion in new group created from in app
image

HSalmon:
Fixes #1700

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -34,6 +34,9 @@ class GroupsController < GroupBaseController
       @group.add_admin! current_user
       @group.creator = current_user
       Measurement.increment('groups.create.success')
+      if @group.is_parent?
+        SetupGroup.create_example_discussion(@group)
+      end
       flash[:success] = t("success.group_created")
       redirect_to @group
     elsif @group.is_subgroup?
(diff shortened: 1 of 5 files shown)


features/step_definitions/start_new_group_steps.rb
```diff
@@ -86,7 +86,7 @@
   page.should have_content 'can\'t be blank'
 end
 
-Then(/^the example content should be created$/) do
+Then(/^the example content should be present$/) do
```
gdpelican: Why change created -> present? Should it be consistent with 'the example content should not be created'?

gdpelican:
Yeesssss...! :dancer:

app/controllers/groups_controller.rb
```diff
@@ -34,6 +34,9 @@ def create
       @group.add_admin! current_user
       @group.creator = current_user
       Measurement.increment('groups.create.success')
+      if @group.is_parent?
```
robguthrie: Not a show stopper but more an FYI.. this should be moved into the GroupService and out of the controller. I'm sure it will be moved real soon as we enable group creation in the angular interface

gdpelican:
Get in my repo! :open_mouth:

PR #1954, opened 2015-02-27
Search thrux

robguthrie:


app/queries/thread_search_query.rb
```diff
@@ -0,0 +1,129 @@
...
+     'comment_bodies'           => :D}
+  end
+
+  def self.motion_field_weights
+    {'name'        => :B,
+     'description' => :C}
+  end
+
+  def self.comment_field_weights
+    {'body' => :D} # So happy :D
```
elimisteve: :D

robguthrie: Yea cool eh? That's a small refactor on top of the amazing work that @gdpelican did to build this monster of a search engine. It really is a thing of beauty. 

Speaking of which.. I guess we should also include authors of the various records as well as proposal outcomes in the indices? I can imagine searching for someones name and some other content.

PR #1955, opened 2015-02-27
angular switcher

s01us:


robguthrie:
Cool.. but not quite right.. it should not redirect to the angular homepage.. it should just render the angular layout, so that the angular app knows to load the group/discussion the user wanted.

IE: preserve the url. If I load /d/1234  no matter what interface i use, i will see discussion 1234

robguthrie:
There is no angular homepage.
:shipit:

gdpelican:
There's some relevant code in `base_controller.rb`, around booting the angular ui; seems to be just a layout on the rails controller.

If I had a quick and dirty guess, I'd guess we'd want something like

```
before_filter :check_angular_ui

def check_angular_ui
  if current_user_or_visitor.angular_ui_enabled?
    @route = request.path # or something
    render 'layouts/angular', layout: false
  end # calling render in a before filter halts the action.
end
```

Then, in `layouts/angular.html.haml`:
```
window.Loomio.defaultRoute = "#{@route}"
```

Also, you could probably use the existing `use_angular_ui` boolean, as I don't think we're using that for anything at the moment.

robguthrie:
I dont get why you fellows think we need to pass the route.

robguthrie:
but otherwise you're both cool

robguthrie:
@s01us this piece of work does have to be done fairly urgently. Are you able to commit to finishing it by end of day Wedneday? If not I'll take it off your hands.

gdpelican:
I'm gonna grab this this morning. :tangerine:

PR #1958, opened 2015-03-01
Update help docs

HSalmon:
- Updated help docs to incorporate feedback from feedback discussion in Loomio Community.
- Added "Help" link to logged-in footer


gdpelican:
Can you please review these before you merge them in @robguthrie? This undid a chunk of the LoomioX work just merged in.

robguthrie:
ARRRG! Sorry! Maybe we should formalise this process to be something like
we both review, then the second reviewer merges?

It could be called the "2 keys for the vault" process.

gdpelican:
I think that works, although it makes us move a little slower, and forces more engagement with the PR list overall.

Can we say something like 'On first review, give it a thumbs up on the PR thread. If, after 24 hours, it hasn't been merged, feel free to merge it?'

PR #1963, opened 2015-03-03
working user locales for Angular

mixmix:
would love to discuss what I've learnt and am thinking 

there are several opinions coded in here:

- we should only serve one locale out to users where the angular app is initialized
- we should namespace the client.en.yml so we can pull the Rails cached translations that are already in memory

### DIFF
--- a/lineman/app/app.coffee
+++ b/lineman/app/app.coffee
@@ -17,10 +17,14 @@ angular.module('loomioApp', ['ngRoute',
 angular.module('loomioApp').config ($translateProvider) ->
   $translateProvider.
     useUrlLoader('/api/v1/translations/en').
-    preferredLanguage('en')
+    preferredLanguage('pt-BR')
+
+    # sends to /api/v1/translations/en?lang=pt-BR
 
 angular.module('loomioApp').run (Records, UserAuthService) ->
   if window? and window.Loomio?
     Records.import(window.Loomio.seedRecords)
     UserAuthService.currentUser = Records.users.find(window.Loomio.currentUserId)
 
+
+    console.log UserAuthService.currentUser.selectedLocale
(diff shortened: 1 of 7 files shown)


lineman/app/app.coffee
```diff
@@ -17,10 +17,14 @@ angular.module('loomioApp', ['ngRoute',
 angular.module('loomioApp').config ($translateProvider) ->
   $translateProvider.
     useUrlLoader('/api/v1/translations/en').
-    preferredLanguage('en')
+    preferredLanguage('pt-BR')
```
s01us: I'd like to know how we get `Records.users.find(window.Loomio.currentUserId).selectedLocale` or `UserAuthService.currentUser.selectedLocale` in here

s01us: If it's bad form to access current_user in here, we might have to fetch and set locale post - initilization

robguthrie: See inbox.coffee lines 6 and 7 for how to access current user.

Also I will show you how to move the translations into the layout.. the same way that other necessary data records are inserted in app/views/layouts/angular.html.haml

robguthrie:
Looking really good. If you get time, could you please read up on angular-translate the angular plugin

s01us:
yeah I have Rob. I've got into the docs and the source code and got advice from Mikey
that's my default these days unless I'm asking for a generic bearing to head off in

robguthrie:
great. awesome to hear it. It would be really awesome to see a test showing
that this approach to fallbacks works as intended.

s01us:
I'm not up to speed with testing the new interface. 

I would start by looking for examples of cucumbers which test the new interface and try replicating those.

s01us:
i.e. that's the plan, let me know if it's stupid please

gdpelican:
This is a great start, but we've unfortunately let it slide for a really long time, and will probably have to rewrite it. Closing this.

PR #1965, opened 2015-03-03
Add comment_replied_to mailer

HSalmon:


app/models/events/comment_replied_to.rb
```diff
@@ -3,10 +3,14 @@ class Events::CommentRepliedTo < Event
 
   def self.publish!(reply)
     return unless reply.parent.present?
-    create!(kind: 'comment_replied_to',
-            eventable: reply,
-            user: reply.author,
-            created_at: reply.created_at)
+    event = create!(kind: 'comment_replied_to',
```
gdpelican: Since you're aping other methods in the controller, this should stay as is, but FYI a more better way, with tap:

```
create!(kind: 'comment_replied_to', 
        eventable: reply,
        user: reply.parent.author, 
        created_at: reply.created_at).tap do |event|
  ThreadMailer.delay.comment_replied_to reply.parent.author, event
end
```

spec/models/events/comment_replied_to_spec.rb
```diff
@@ -9,16 +9,23 @@
...
   describe "::publish!" do
     let(:event) { double(:event, notify_users!: true) }
+    let(:mailer) { ThreadMailer.new }
     before { Event.stub(:create!).and_return(event); comment }
+    before { ThreadMailer.stub_chain(:delay, :comment_replied_to) }
 
     it 'creates an event' do
       Event.should_receive(:create!).with(kind: 'comment_replied_to',
                                           eventable: comment,
-                                          user: comment.author,
```
gdpelican: Why this change in functionality? I think it makes more sense on events for the user to be the 'actor', rather than the recipient of the notification.

app/models/events/comment_replied_to.rb
```diff
@@ -3,10 +3,14 @@ class Events::CommentRepliedTo < Event
...
+                    eventable: reply,
+                    user: reply.parent.author,
+                    created_at: reply.created_at)
+
+    ThreadMailer.delay.comment_replied_to(reply.parent.author, event)
```
gdpelican: reply.parent.author is right here, but for consistency if nothing else, we'll want to keep the 'user' of the event as the replyer (although I might be able to be convinced otherwise)

gdpelican:
This looks good to me :+1:

spec/models/events/comment_replied_to_spec.rb
```diff
@@ -9,16 +9,22 @@
 
   describe "::publish!" do
     let(:event) { double(:event, notify_users!: true) }
+    let(:mailer) { ThreadMailer.new }
```
gdpelican: Don't think you use this line?

gdpelican:
Thanks @HSalmon !

PR #1966, opened 2015-03-03
meta google site verification

simontegg:
add google site verification to activate Web Master Tools for better marketing knowledge


robguthrie:
Whoa @gdpelican this should have been in a ENV variable. We can't have other people knowing a site verification token.

gdpelican:
Even if it was in ENV, it'd just show up in the markup on the client side, so I don't think we can keep it a secret.

gdpelican:
Er. except the marketing page doesn't get rendered ever for third parties. So I'll waffle again and say it's good.

robguthrie:
yaaaa...ok

PR #1972, opened 2015-03-05
fix the added_to_group email readability sense

s01us:
a translator pointed out an email that didn't make sense : 

This translation:

```ruby
      content: "%{who} has added you to %{which_group} on Loomio. Click here to go to the group:"
```

was not followed by a link in the email. I've just changed the order of the email so it reads correctly


gdpelican:
Cool, thanks! :100:

PR #1973, opened 2015-03-06
make avatar_url be served more reliabley :fish: fixes #1807

### LINKED ISSUE #1807: Some avatar images are broken in "Yesterday on Loomio" email
I can't post a screenshot of the example I have because it contains confidential info, but about half the avatar images of users in the Yesterday on Loomio email are coming up as broken image links.

s01us:
YEAH!  :bomb: 

this should fix the mentioned issue - some avatars (type gravatar) not coming through in yesterday emails

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -8,7 +8,7 @@ class User < ActiveRecord::Base
   AVATAR_KINDS = %w[initials uploaded gravatar]
   IMAGE_SIZES = {
     large: 170,
-    medium_large: 70,
+    medlarge: 70,
     medium: 35,
     small: 25
   }.with_indifferent_access.freeze
@@ -22,11 +22,12 @@ class User < ActiveRecord::Base
 
   has_attached_file :uploaded_avatar,
     styles: {
-              large:    "#{IMAGE_SIZES.fetch(:large, :medium)}x#{IMAGE_SIZES.fetch(:large, :medium)}",
-              medlarge: "#{IMAGE_SIZES.fetch(:medlarge, :medium)}x#{IMAGE_SIZES.fetch(:medlarge, :medium)}",
-              medium:   "#{IMAGE_SIZES.fetch(:medium)}x#{IMAGE_SIZES.fetch(:medium)}",
-              small:    "#{IMAGE_SIZES.fetch(:small, :medium)}x#{IMAGE_SIZES.fetch(:small, :medium)}"
-            }
+              large:    "#{IMAGE_SIZES[:large]}x#{IMAGE_SIZES[:large]}",
+              medlarge: "#{IMAGE_SIZES[:medlarge]}x#{IMAGE_SIZES[:medlarge]}",
+              medium:   "#{IMAGE_SIZES[:medium]}x#{IMAGE_SIZES[:medium]}",
+              small:    "#{IMAGE_SIZES[:small]}x#{IMAGE_SIZES[:small]}"
+            },
+    default_url: "/avatars/:style/missing.png"
   validates_attachment :uploaded_avatar,
     size: { in: 0..User::MAX_AVATAR_IMAGE_SIZE_CONST.kilobytes },
     content_type: { content_type: /\Aimage/ },
@@ -306,7 +307,7 @@ class User < ActiveRecord::Base
   end
 
   def has_uploaded_image?
-    uploaded_avatar.url(:medium) != '/uploaded_avatars/medium/missing.png'
+    uploaded_avatar.url(:medium) != '/avatars/medium/missing.png'
   end
 
   def has_gravatar?(options = {})
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,10 +6,12 @@ class User < ActiveRecord::Base
   require 'digest/md5'
 
   AVATAR_KINDS = %w[initials uploaded gravatar]
-  LARGE_IMAGE = 170
-  MED_LARGE_IMAGE = 70
-  MEDIUM_IMAGE = 35
-  SMALL_IMAGE = 25
+  IMAGE_SIZES = {
+    large: 170,
+    medium_large: 70,
+    medium: 35,
+    small: 25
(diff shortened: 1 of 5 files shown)


robguthrie:
amazing. Do you have some kind of test case or something for this?

app/models/user.rb
```diff
@@ -287,7 +287,7 @@ def inactive_message
     I18n.t(:inactive_html, path_to_contact: '/contact').html_safe
   end
 
-  def avatar_url(size=nil,avatar_kind=nil)
+  def avatar_url(size=nil, kind=nil)
```
gdpelican: Barg, this whole method as currently constructed kinda pretty much is bad.

Can we do the following?

- create a constant hash of image size values
```
IMAGE_SIZES = {
  small: xx,
  medium: xx,
  large: xx,
  etc: xx
}.with_indifferent_access.freeze
```
- find our size in that hash and apply some sensible default
```
gravatar_url(size: IMAGE_SIZES[size || :medium])
```
- make it more explicit that avatar_kind is an attribute, and kind is a parameter
```
if kind || self.avatar_kind == 'gravatar'
```

robguthrie: That sounds great @gdpelican. 

I think kind should be removed from the arguments of the method. It seems like it's breaking a principal or two exposing the detail like that. 

I think I would classify this problem as resolved when there is a method that you can call without needing to pass any arguments, and it returns a usefully sized image, even if the user has not selected a profile image. IE: It should return some default image if there is no profile image.

Can I propose we use the following method signatures?

profile_image_url(size = :medium), which always returns a working image url.

and

has_profile_image? which returns `false` if profile_image_url will return the default image.

I think at this stage it's worth changing the name of the method and removing the buggy method, so that we can clean up the codebase properly.
 
Also.. it's considered funky and fresh, when accessing a hash with a default, to use #fetch. EG: IMAGE_SIZES.fetch(size, :medium).

s01us: I can do this.
Notice this has expanded considerably into a refactor, which is a pattern
which makes me feel like I'm rubbish collecting. (relevant? ? note for
reflection : general grumbling)

I don't have a test written but did find a test case that failed.

I'll write it up and get this sorted

gdpelican: I'm happy to take it.

s01us:
hey @gdpelican I was gonna get into this tmrw. 
Are you taking this over?

I don't want us doubling up on work. Also, don't want to distract from other important work on Angular

gdpelican:
Ah, cool. If you want to pick it up that's great, I just thought you were
frustated and done with it and I'd like to get it done.

Feel free to use / trash the commit I put in there. :bomb:

s01us:
yeah sorry, had  a chat with rob as well about the tone i was projecting because it came across unclear.
I'm happy to sort this out. <3 <3

app/models/user.rb
```diff
@@ -20,10 +22,10 @@ class User < ActiveRecord::Base
...
   has_attached_file :uploaded_avatar,
     styles: {
-              large: "#{User::LARGE_IMAGE}x#{User::LARGE_IMAGE}#",
-              medlarge: "#{User::MED_LARGE_IMAGE}x#{User::MED_LARGE_IMAGE}#",
-              medium: "#{User::MEDIUM_IMAGE}x#{User::MEDIUM_IMAGE}#",
-              small: "#{User::SMALL_IMAGE}x#{User::SMALL_IMAGE}#",
+              large:    "#{IMAGE_SIZES.fetch(:large, :medium)}x#{IMAGE_SIZES.fetch(:large, :medium)}",
+              medlarge: "#{IMAGE_SIZES.fetch(:medlarge, :medium)}x#{IMAGE_SIZES.fetch(:medlarge, :medium)}",
+              medium:   "#{IMAGE_SIZES.fetch(:medium)}x#{IMAGE_SIZES.fetch(:medium)}",
+              small:    "#{IMAGE_SIZES.fetch(:small, :medium)}x#{IMAGE_SIZES.fetch(:small, :medium)}"
```
s01us: this is not pretty 
I tried making a class method, but that was also ugly :fish_cake:

app/views/base/_avatar.html.haml
```diff
@@ -1,6 +1,6 @@
 -id ||= nil
--kind ||= user.avatar_kind
 -size ||= "medium"
+-kind ||= user.avatar_kind
```
s01us: we have to have kind for styling reasons

s01us: specifically: 

```ruby 
8: class: "#{size}-user-avatar #{kind}-user-avatar avatar-canvas"
```

app/models/user.rb
s01us: why are all my children ugly?

app/views/base/_avatar.html.haml
robguthrie: Mix.. I really don't like the image so I deleted it.

It's ok to have kind. I know it is needed. but you don't need to specify it when asking the object for a valid image url.

s01us:
this is passing. 
I'm going to squash it. 

let me know if there is anything else otherwise QA + deploy ?

app/models/user.rb
s01us: drop fetch
use medium_large for medlarge

supply missing avatar images

s01us:
discussed with rob. 

going to : 

[ ] write test for has_avatar?
[ ] trying and smartly set gravatar after user create if they have a gravatar
[ ] create has_avatar_image?  and use this to determine if serving image/ initials

robguthrie:
Mix, I've done a quick fix to remove kind from avatar_url method. I think the med-large -> medium_large stuff is worth keeping.. and forget everything else maybe?

s01us:
Did that issue get resolved.
This week has been a bit crazy, sorry for slowness

robguthrie:
Salright, yep. I've fixed the issue now. We can close this.

PR #1974, opened 2015-03-06
use FORCE_SSL to determine http(s) email links

s01us:
adds config for using https links in emails

### DIFF
--- a/spec/mailers/user_mailer_spec.rb
+++ b/spec/mailers/user_mailer_spec.rb
@@ -31,6 +31,17 @@ describe UserMailer do
     it 'assigns confirmation_url for email body' do
       @mail.body.encoded.should match("http://localhost:3000/g/#{@group.key}")
     end
+
+    it 'uses http links by default' do
+      expect(@mail.body.encoded).to match(/http:/)
+    end
+
+    it 'uses https links if action_mailer protocol is set to https' do
+      Loomio::Application.config.action_mailer.default_url_options[:protocol] = 'https'
+
+      @mail = UserMailer.group_membership_approved(@user, @group)
+      expect(@mail.body.encoded).to match(/https:/)
+    end
   end
 
   context 'sending email on being added to group' do
--- a/config/application.rb
+++ b/config/application.rb
@@ -53,7 +53,8 @@ module Loomio
     config.assets.version = '1.4'
 
     config.action_mailer.default_url_options = {
-      host: ENV['CANONICAL_HOST']
+      host:     ENV['CANONICAL_HOST'],
+      protocol: ENV['FORCE_SSL'] ? 'https' : 'http'
     }
 
     config.roadie.url_options = nil
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -33,7 +33,6 @@ Loomio::Application.configure do
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.default_url_options = { host: 'localhost', port: 3000 }
   config.action_mailer.delivery_method = :smtp
-  config.action_mailer.delivery_method = :smtp
 
   config.action_mailer.smtp_settings = { :address => "localhost", :port => 1025 }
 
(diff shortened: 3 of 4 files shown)


robguthrie:
mix. thank you for this.. it's reminded me that we already have
ENV['FORCE_SSL'] which we should use rather than introducing another ENV.

s01us:
Perfect.

I was wondering whether there was a canonical source which references all
the possible env.
I reckon.   .env-example could be a good start?

s01us:
ready

robguthrie:
No there is not but we should have one.

s01us:
workin on it

spec/mailers/user_mailer_spec.rb
```diff
@@ -29,7 +29,7 @@
     end
 
     it 'assigns confirmation_url for email body' do
-      @mail.body.encoded.should match("http://localhost:3000/g/#{@group.key}")
+      @mail.body.encoded.should match("https://localhost:3000/g/#{@group.key}")
```
gdpelican: I'm not sure that this test is actually testing your change, since you're just setting a default in the test environment and then seeing that it comes through.

```
ENV['FORCE_SSL'] = true
expect(@mail.body.encoded).to match /https/
```

s01us: I don't quite follow. I followed a red-green-refactor format. I wrote a
test which I that feels like it's of the form :

Given env is set to FORCE_SSL
When an email with link is generated
Then it has https links

I haven't written a test for the negative case, but i did see it behaving.
I don't know how to stub ENV, I did research it.

Can we talk about where to next on this

gdpelican: The testing environment usually comes with a blank ENV (unless you've defined a `test.env` file for dotenv, or set it in a different way). 
A couple different options for testing ENV-based behaviour [here](https://robots.thoughtbot.com/testing-and-environment-variables) (I personally much prefer caching and resetting the ENV value to stubbing the ENV object.)

```
it 'uses https when FORCE_SSL is set to true' do 
  ssl_env = ENV['FORCE_SSL']
  ENV['FORCE_SSL'] = true
  expect(@mail.body.encoded).to match /https:/
  ENV['FORCE_SSL'] = ssl_env
end

it 'uses http when FORCE_SSL is set to false' do 
  ssl_env = ENV['FORCE_SSL']
  ENV['FORCE_SSL'] = false
  expect(@mail.body.encoded).to match /http:/
  ENV['FORCE_SSL'] = ssl_env
end
```

s01us: assuming you mean I should change `config/environments/test.rb` so that it has the same line as production : 

```ruby
 protocol: ENV['FORCE_SSL'] ? 'https' : 'http'
```

then your proposed method doesn't work. I've tried it and you can change that env, but the app configuration is set at initializations (i think).

I started down a path of overwriting / stubbing the config (was looking at Loomio::Application.config.force_ssl , but that's the wrong thing)

found something. posting it up, but it doesn't feel pretty

spec/mailers/user_mailer_spec.rb
```diff
@@ -31,6 +31,17 @@
     it 'assigns confirmation_url for email body' do
       @mail.body.encoded.should match("http://localhost:3000/g/#{@group.key}")
     end
+
+    it 'uses http links by default' do
+      @mail.body.encoded.should match(/http:/)
+    end
+
+    it 'uses https links if env FORCE_SSL == true' do
+      Loomio::Application.config.action_mailer.default_url_options[:protocol] = 'https'
```
s01us: @gdpelican this is the best I could do. :disappointed:

robguthrie:
If the test is not testing the ENV, then just take it out.. and make this a one liner. I'm ok with that.

s01us:
I've changed the test to just reflect that we're testing action_mailer.

@robguthrie if you'd like stuff cut out can we either pair on this for 2 mins or can you just cut the bits out yourself. I'm low on time today and I don't have capacity to re-do things if I cut the wrong bits
:herb:

PR #1975, opened 2015-03-07
all the loomio referenced env

s01us:
pulled all the ENV loomio use using : 

```
grep -rIsho -e "ENV\.fetch\(.*\)"  >> temp
grep -rIsho -e "ENV\[.*\]"  >> temp
```

then sorting and uniq'ing with ruby.
use the option `-rIso` to show the list of files each is in. 

note e.g that `GEM_PATH` and `GEM_HOME` are spring related, so not needed ?


s01us:
@gdpelican  ^ regex ! 
could have been more complicated and used some `|` or but grep + ruby/node/script is more time effective

s01us:
please feel free to comment or commit your own edits to this. 

I'll need help sorting out the deprecated env & what some of the things are 
like what is `ALLOW_ROBOTS` :whale2:

s01us:
oh i see. crawlable

gdpelican:
Hey @elimisteve, we're gearing up for an internal alpha release in the quite near future, at which point we'll have a really solid checklist of angular-type issues to have a look at; I'll be keeping the issue tracker up to date, probably with some things that'll be pretty easy for outsiders to pick up.

Thanks!

elimisteve:
@s01us I've seen the Angular work only from a user's perspective.  Only 1 month away from v1.0?!  Nice!!

@gdpelican Sounds great, let me know when the alpha is "released" and I'll see what I can tackle.

robguthrie:
Thanks mix

PR #1977, opened 2015-03-08
bundle update!

s01us:
lets see if the tests pass 

want devise-i18n at the least for updated translations, but thought I'd have a look while i was here


Gemfile.lock
```diff
@@ -292,9 +292,9 @@ GEM
       nokogiri (~> 1.5, >= 1.5.11)
     font-awesome-sass (4.2.2)
       sass (~> 3.2)
-    foreman (0.77.0)
-      dotenv (~> 1.0.2)
-      thor (~> 0.19.1)
```
s01us: woah, it's dropping back foreman?

gdpelican:
Foreman rollback is weird, but I'm not too terribly worried about it; I don't think we actually use that gem for anything.

Green check and a smoke test is good enough for me here. :smoking:

PR #1979, opened 2015-03-09
provide minimum translation of markdown help

s01us:
check out this!

**the problem**: we're linking all users to an english-only markdown help doc

**the solution**: uses google page translate url instead of just serving the plain url

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -207,4 +207,12 @@ module ApplicationHelper
   def hide_beta_logo?
     current_user_or_visitor.belongs_to_manual_subscription_group?
   end
+
+  def markdown_help_url
+    if current_locale.to_s == 'en'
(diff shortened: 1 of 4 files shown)


s01us:
http://translate.google.co.nz/translate?hl=sl&u=https://help.github.com/articles/markdown-basics/&prev=search

robguthrie:
@s01us  I just looked at this.. and it seems like the translation breaks the formatting of the page.

robguthrie:
I I merged then quickly reverted.

But yea.. the translation service breaks the formatting of the markdown, making the page useless. I think you should just quickly throw together an in app version of the page that we can translate ourselves

s01us:
i got sign off from the slovenian translator that while it wasn't perfect it was better than nothing

robguthrie:
I don't think it's good enough.

PR #1991, opened 2015-03-11
markdown help page. needs translation applied to the strings

robguthrie:
The beginnings of a simple markdown tutorial.

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -7,4 +7,8 @@ class HelpController < ApplicationController
       render 'help'
     end
   end
+
+  def markdown
+
+  end
 end
--- a/app/views/comments/edit.html.haml
+++ b/app/views/comments/edit.html.haml
@@ -9,4 +9,4 @@
           = check_box_tag 'comment[uses_markdown]', 1, @comment.uses_markdown
           = label_tag :uses_markdown, 'Markdown'
           = surround '(', ')' do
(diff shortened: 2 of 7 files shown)


robguthrie:
@s01us i thought I'd offer to meet you half way on writing a markdown page. If you do the translations we're done.

s01us:
looks pretty accessible. 
i'll sort that over the weekend

s01us:
looks like this : 


could do with some more formatting work (I also had trouble with haml). 
I think the important thing is that there are some strings which are translatable which is better than having nothing.  

Note I've started a **help.en.yml** which I think we're going to want in the future.
once this is merged I can can get the help file started to be translated on Transifex

s01us:
this is ready @robguthrie 

the code being indented in the `pre` I think is a hard one to fix and is related to the layout indentation from that's being made by haml?

if we care about that (because currently users copying and pasting those examples will get bad markdown because of all the white space), we could : 

- tell haml not to indent as it compiles? 
- use some nasty JS on page : 
  - ```javascript
      $('pre')[0].textContent.replace(/^\s{4,}\n/gm, 'BLANK\n').replace(/^\s{4,}/gm, "").replace(/BLANK\n/gm, '\n')
    ```

mixmix:
Can we talk about this please.
I'm ok with stuff being dropped but I'm a bit upset about the lack of
clarity

PR #1992, opened 2015-03-11
add a few depenedent destroys for user..

robguthrie:


app/models/user.rb
```diff
@@ -96,10 +97,11 @@ class User < ActiveRecord::Base
 
   has_many :notifications, dependent: :destroy
   has_many :comments, dependent: :destroy
-  has_many :attachments
+  has_many :attachments, dependent: :destroy
```
s01us: this could wreck documents and seems strange considering you're leaving comments?

robguthrie: I'm not leaving comments. Deactivate is where we keep user data.. delete is where we do our best to remove everything and leave no dangling references. It's a work in progress, we do not do any user deletion right now

app/models/user.rb
```diff
@@ -96,10 +97,11 @@ class User < ActiveRecord::Base
...
 
   has_one :deactivation_response,
-          class_name: 'UserDeactivationResponse'
+          class_name: 'UserDeactivationResponse',
+          dependent: :destory
```
gdpelican: Typo!

robguthrie: My all time most common rails typo! Thanks

PR #1998, opened 2015-03-12
Add cukes for viewing marketing page

### LINKED ISSUE #1988: We could really do with a few more basic cukes
We need cucumber tests around viewing the front marketing page, and signing in though the sign in page.

These are pretty trivial to write but it would be if someone could do that for us

HSalmon:
Fixes #1988


gdpelican:
Looks like that new cuke you wrote failed, @HSalmon; any idea why? Looks like it should pass to me.

gdpelican:
Thanks @HSalmon :rocket:

PR #1999, opened 2015-03-12
ensure discussion readers are unique fixes #1997

### LINKED ISSUE #1997: BUG: Discussions appearing multiple times in dashboard!
image

Looks like multiple discussion readers are being created at some point! Yikes.

robguthrie:


app/models/discussion_reader.rb
```diff
@@ -11,9 +11,10 @@ class DiscussionReader < ActiveRecord::Base
 
   def self.for(user: nil, discussion: nil)
     if user.is_logged_in?
-      where(user_id: user.id, discussion_id: discussion.id).first_or_initialize do |dr|
-        dr.discussion = discussion
-        dr.user = user
+      begin
```
robguthrie: @gdpelican fyi: found this approach to making this thread safe:

http://www.rubydoc.info/docs/rails/ActiveRecord/Relation:find_or_create_by

See the bottom of the doc, above the comments

gdpelican: :+1:

PR #2000, opened 2015-03-12
Fixes Syntax Error on Group View

gbrego:
Hello, 

I have been keep up with Loomio Development.

I think I have found some syntax error on Master Code.

Please revise, comment and accept my pull request (if you want)!

### DIFF
--- a/app/views/discussions/_discussion_context.html.haml
+++ b/app/views/discussions/_discussion_context.html.haml
@@ -17,7 +17,7 @@
     - if can? :set_volume, @discussion
       - current_volume = @discussion_reader.volume
       .btn-group.dropdown.volume-dropdown
-        %button{type: 'button', class:"btn btn-left btn-default dropdown-toggle", 'data-toggle': "dropdown", 'aria-expanded': false}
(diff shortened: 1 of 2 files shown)


gdpelican:
Hey @gbrego, thanks for the PR?

Are you experiencing an error page when trying to view a group?

gbrego:
Yes, I'm experiencing syntax error on groups and discussions, as the screenshot you may see below

loomio_error

gbrego:
@gdpelican thanks answering.

Yes I'm experiencing errors on group and discussions (it can be seen below)
loomio_error

robguthrie:
Hi @gbrego, your patch is kind of interesting. We are not having problems with this part of the code, indeed if we were, our site would be very broken.

I cannot understand why you would be experiencing a problem with the code, or why your changes make any difference.

I'm going to close this pull request for now, we don't need this. But I do welcome your contribution, and thank you for taking the time to do this. I hope you enjoy using Loomio and maybe can contribute more in the future.

gbrego:
@robguthrie  

I've found the cause of the error.

I was using ruby-2.2.0-preview (installed using homebrew) when I changed to ruby-2.2.0 it solved itself.

thanks for answering my Pull Request

PR #2009, opened 2015-03-15
restore Intercom default messenger

rdbartlett:
I've put the default Intercom messenger back in because it makes it easier for people to get hold of us. Sorry for the hassle before, there was something wrong with my brain.


gdpelican:
:party:

PR #2012, opened 2015-03-16
Group max size

robguthrie:


robguthrie:
@gdpelican did we do anything stupid here?

app/controllers/application_controller.rb
```diff
@@ -29,6 +31,14 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  def show_max_size_warning?
+    ENV['SHOW_LOOMIO_ORG_MARKETING'] and @group and @group.approaching_max_size? and !@group.max_size_reached?
```
gdpelican: Don't use `and`, use &&. This will error if `@group` is nil, and will evaluate all of these possibilities each time.

app/models/group.rb
```diff
@@ -355,12 +355,31 @@ def discussion_private_default
...
+  def org_members_count
+    if is_subgroup?
+      parent.org_members_count
+    else
+      Membership.active.where(group_id: [id, subgroups.pluck(:id)].flatten).pluck(:user_id).uniq.count
     end
+  end
+
+  def approaching_max_size?
+    ENV['SHOW_LOOMIO_ORG_MARKETING'] && org_members_count > (org_max_size * 0.8)
```
gdpelican: Is 'SHOW_LOOMIO_ORG_MARKETING' essentially 'this is loomio.org, instead of some 3rd party install'? If so, is it possible to change it so it's a little more clear, before we start using it everywhere?

app/controllers/application_controller.rb
```diff
@@ -10,6 +10,8 @@ class ApplicationController < ActionController::Base
   helper_method :current_user_or_visitor
   helper_method :dashboard_or_root_path
   helper_method :subdomain
+  helper_method :show_max_size_warning?
+  helper_method :show_max_size_reached?
```
gdpelican: Does this need to be all the way up in ApplicationController? What about BaseController? Or separating out into a GroupInvitations concern or something? No need to pollute this namespace, I don't think.

db/migrate/20150315210731_change_group_max_size_default_again.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeGroupMaxSizeDefaultAgain < ActiveRecord::Migration
+  def change
+    change_column :groups, :max_size, :integer, default: 100
```
gdpelican: Why 100 here and 300 in the schema?

app/controllers/application_controller.rb
gdpelican: Ooops, looks like this might have been pulled out of the ruby spec recently. Ignore!

app/models/group.rb
rdbartlett: Yep, that's the correct interpretation. Any suggestion for what to call it?

db/migrate/20150315210731_change_group_max_size_default_again.rb
rdbartlett: Oops.

app/models/group.rb
gdpelican: `HOSTED_BY_LOOMIO`?
`THIRD_PARTY_INSTALL`, and take the inverse?

rdbartlett: Yeah I like THIRD_PARTY_INSTALL

rdbartlett:
@gdpelican I agree about the awkward name for SHOW_LOOMIO_ORG_MARKETING. I've switched it out for "hosted_by_loomio" in the positive case and "third_party_install" in the negative. Have updated the config vars on production too.

PR #2014, opened 2015-03-16
show the intercom widget everywhere

rdbartlett:
@robguthrie is there any reason not to do this?


gdpelican:
Definitely want to hold off on this at least until we get the 'Intercom looks way too much like a chat box' thing figured out.

I had a peek earlier and it's a little trickier than it should be.

PR #2019, opened 2015-03-17
add "turn off angular" link so stuck people can get out easily

robguthrie:
image

Had a few people venture into the angular interface and not know how to get out. So added this link


rdbartlett:
THANKS!

PR #2026, opened 2015-03-20
Dashboard thrux

gdpelican:
(:sunglasses:)

Various dashboard page upgrades:
- Add infinite scrolling for sorting by date
- Group by organization rather than group (include subgroups)
- Allow load more threads under groups
- Uses new-router
- Better controller structure for dashboard controller


rdbartlett:
Thanks James! :penguin:

PR #2032, opened 2015-03-24
Fix dropdowns

gdpelican:
Fixes angular dropdown menus on navbar

Main issue was that the new angular router requires controllers to return a value, otherwise b0rktown when trying to load controllers (since coffeescript returns the last value)
Secondary issue was missing 'dropdown-menu-right' css class


robguthrie:
whoo hoo!

gdpelican:
Mergin dat yo.

PR #2033, opened 2015-03-24
Fix attachments for the Olog(n)th time :)

gdpelican:


gdpelican:
Maybe we have a fear of attachment.

PR #2035, opened 2015-03-24
Dashboard mark as read

gdpelican:
Okay, lots of stuff here:

- Saves user preferences when changed
- More robust querying of client side db, so things don't slip through that shouldn't be there
- More better 'fire and forget' (things load faster and backfill later :zap:)
- Added mute and mark as unread buttons, which totally work
- Lots of cleanup and code cruft removal
- More reliably filter out muted discussions

Could likely do with a code review @robguthrie

### DIFF
--- a/app/controllers/api/discussion_readers_controller.rb
+++ b/app/controllers/api/discussion_readers_controller.rb
@@ -5,4 +5,16 @@ class API::DiscussionReadersController < API::RestfulController
     self.resource = DiscussionReader.for(user: current_user, discussion: @discussion)
   end
 
+  def mark_as_read
+    load_resource
+    resource.viewed! (discussion_event || @discussion).created_at
+    respond_with_resource
+  end
+
+  private
+
+  def discussion_event
+    Event.where(discussion_id: @discussion.id, sequence_id: params[:sequence_id]).first
+  end
+
 end
--- a/app/extras/group_discussions_viewer.rb
+++ b/app/extras/group_discussions_viewer.rb
@@ -2,7 +2,7 @@ class GroupDiscussionsViewer
   def self.for(group: nil, user: nil, filter: nil)
     groups = groups_displayed(group: group, user: user)
     Queries::VisibleDiscussions.new(groups: groups, user: user).tap do |discussions|
-      discussions = discussions.unread if filter == 'unread'
+      discussions = discussions.unread if filter == 'show_unread'
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -5,6 +5,9 @@ class User < ActiveRecord::Base
   require 'net/http'
   require 'digest/md5'
 
+  validates_inclusion_of :dashboard_sort,   in: %w[sort_by_group sort_by_date]
+  validates_inclusion_of :dashboard_filter, in: %w[show_all show_unread]
+
   AVATAR_KINDS = %w[initials uploaded gravatar]
   LARGE_IMAGE = 170
   MED_LARGE_IMAGE = 70
--- a/app/controllers/api/discussions_controller.rb
+++ b/app/controllers/api/discussions_controller.rb
@@ -27,12 +27,6 @@ class API::DiscussionsController < API::RestfulController
     respond_with_discussion
   end
 
-  def mark_as_read
-    event = Event.where(discussion_id: @discussion.id, sequence_id: params[:sequence_id]).first
-    discussion_reader.viewed! (event || @discussion).created_at
-    respond_with_discussion
-  end
-
   private
 
   def respond_with_discussion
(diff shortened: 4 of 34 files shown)


app/controllers/api/discussion_readers_controller.rb
```diff
@@ -5,4 +5,16 @@ def load_resource
     self.resource = DiscussionReader.for(user: current_user, discussion: @discussion)
   end
 
+  def mark_as_read
```
robguthrie: I reckon this should be on the discussions api. but not worries.

app/extras/group_discussions_viewer.rb
```diff
@@ -2,7 +2,7 @@ class GroupDiscussionsViewer
   def self.for(group: nil, user: nil, filter: nil)
     groups = groups_displayed(group: group, user: user)
     Queries::VisibleDiscussions.new(groups: groups, user: user).tap do |discussions|
-      discussions = discussions.unread if filter == 'unread'
+      discussions = discussions.unread if filter == 'show_unread'
```
robguthrie: show_unread is kinda a bad name.. maybe 'unread_only'?

app/models/user.rb
```diff
@@ -5,6 +5,9 @@ class User < ActiveRecord::Base
   require 'net/http'
   require 'digest/md5'
 
+  enum dashboard_sort:   [:sort_by_date, :sort_by_group]
+  enum dashboard_filter: [:show_all, :show_unread]
```
robguthrie: How are these enums feeling for you? I have a hunch that when we're not doing scales.. it's easier to read, debug, write as plain strings

app/serializers/discussion_reader_serializer.rb
```diff
@@ -9,5 +10,13 @@ class DiscussionReaderSerializer < ActiveModel::Serializer
              :read_items_count,
              :read_salient_items_count,
              :last_read_sequence_id,
-             :volume
+             :volume,
+             :is_read
+  def is_read
```
robguthrie: Yikes! this will cause an extra db query for every discussion. Can't we check unreadness on the client side?

robguthrie: Or with the query we're using to fetch the thing in the first place? VisibleDiscussions will give you read_counts etc

app/serializers/discussion_reader_serializer.rb
```diff
@@ -9,5 +10,13 @@ class DiscussionReaderSerializer < ActiveModel::Serializer
...
+    object.last_read_sequence_id == object.discussion.last_sequence_id
+  end
+
+  def group_id
+    object.discussion.group_id
```
robguthrie: and this.. I'm wondering what I'm going to find down the page as the reason why you had to put this here..

robguthrie:
This looks pretty good. Only concerns are that Enums look dodgy in this case, and a big problem with triggering single discussion queries from the serialiser. We have to find a way around that.. and it feels like it should not be too much effort to compute unreadness on the client.

app/serializers/discussion_reader_serializer.rb
gdpelican: Biggest thing is that I want to do `collection.find(unread: true)` for discussion readers in a loki query; joining on the client side's too much of a pain I think.

But I can just fiddle this onto the reader on the client side like we do with the discussion_id, probably

app/extras/group_discussions_viewer.rb
gdpelican: I waffled about this some; I think we're going to want to put more filters on down the line, and wanted a common name that would make sense for them ('show_muted', 'show_participating', 'show_authored_by_me', whatever)

app/models/user.rb
gdpelican: Enums are the sexy new thing!

I think I might replace them with a string and a model validation.

app/serializers/discussion_reader_serializer.rb
robguthrie: I don't think it should be a stored value, otherwise it will wind up being wrong. How about you do a loki where() and pass a function to filter unread. if you do it last it should not be very expensive.

robguthrie:
overall it's looking like a masterpiece. Please checkout https://trello.com/c/sIDtaIlz/415-latest-inbox-line-item-options-ideas  and let us know what you think on the card.

gdpelican:
where loads every model in the store and runs the function on it, which their docs discourage if you can avoid it. I'll try some fiddly things first, but we can use it if all else fails.
https://github.com/techfort/LokiJS/wiki/Indexing-and-Query-Performance

If I can muster up the motivation maybe I'll try benchmarking some things. :muscle:

robguthrie:
yep.. but if you've already applied filters for group and timeframe then
the `where` impact over the remaining set will be insignificant i reckon.

PR #2042, opened 2015-03-25
Thread position

robguthrie:
still work to do


app/controllers/api/events_controller.rb
```diff
@@ -19,4 +19,8 @@ def default_page_size
     50
   end
 
+  def params_reverse?
+    params[:reverse] == 'true'
```
gdpelican: heh, whoops. Silly...

gdpelican:
This looks great so far.

PR #2045, opened 2015-03-25
Group form

gdpelican:
This makes the group form workflow work again, including:

- Creating a new group (`/g/new`)
- Creating a new subgroup (`/g/new?parentId=xxx`)
- Editing an existing group (`/g/xxx/edit`)

todo before merge: 
- Privacy options (I've got a good idea of how to do this)
- Add some integration testing around group creation / editing
- Design and implement some buttons to make this an actual workflow, instead of just urls


robguthrie:
you are amazing

PR #2046, opened 2015-03-25
Accessibility improvements

rdbartlett:
I'll update this Pull Request throughout the day.

First commit:

* add text for screenreaders to the navigation menu using html `title`s and Bootstrap's `.sr-only` class
* hide icons and avatars using `aria-hidden: 'true'`

Second commit:

* Break pages into html `section`s, and add `role: 'region'` to make it easy for screenreaders to tab through the sections.
* when the region has a visible label, use `aria-labelledby: 'IDofLabel'`, then on the element #IDofLabel add `role: 'heading'`
* when the region doesn't have a visible label, use `aria-label: 'Some label'`


robguthrie:
@rdbartlett i've turned notification volume dropdown into a form and added a flash message to let you know it saved successfully. tested with screenreader it works well.. and aria-live is cool, I can easily imagine us having lots of sr-only flash messages, eg: for things like liking a comment, and maybe posting a comment too.

Would you mind fixing the styling of the button if you get a chance?

robguthrie:
FYI.. I'm going to keep this merged into master pretty tightly, it helps a lot for others wanting stuff to copy.

PR #2055, opened 2015-03-28
Angular membership page

gdpelican:
Plenty more to do here (notably some design work and a membership search filter), but this gets us off the ground and lets people do basic membership page things like removing members and making people coordinators.

Codewise, might be worth taking a look at LoadingService, which is a totally general and halfway clean solution for taking the cruftiness out of our current loading methods.


lineman/app/components/group_page/members_card/members_card.haml
```diff
@@ -9,12 +9,12 @@
   #members-card-title.card-title{ translate: 'group_page.members', role: 'heading' }
 
   .group-member-list
-    .group-member-list-item{ng-repeat: 'member in group.members()'}
+    .group-member-list-item{ng-repeat: 'member in groupPage.group.members() | limitTo:5'}
```
robguthrie: FYI: Limit to 20 would be better

lineman/app/services/loading_service.coffee
```diff
@@ -0,0 +1,9 @@
+angular.module('loomioApp').factory 'LoadingService', (Records) ->
```
robguthrie: When I was tired I did not get this. But now I think it's cool

lineman/app/models/membership_records_interface.coffee
```diff
@@ -2,18 +2,18 @@ angular.module('loomioApp').factory 'MembershipRecordsInterface', (BaseRecordsIn
...
+    fetchByGroup: (groupKey) ->
+      @restfulClient.getCollection {group_key: groupKey}
 
-    makeAdmin: (membership, success, failure) ->
+    makeAdmin: (membership) ->
       @restfulClient.postMember membership.id "make_admin"
 
-    removeAdmin: (membership, success, failure) ->
+    removeAdmin: (membership) ->
       @restfulClient.postMember membership.id "remove_admin"
```
robguthrie: I love restfulClient. We got that right I reckon.

lineman/app/components/memberships_page/memberships_page_controller.coffee
```diff
@@ -1,15 +1,27 @@
...
-    CurrentUser.isAdminOf($scope.group)
+  Records.groups.findOrFetchByKey($routeParams.key).then (group) =>
+    @group = group
+    @loadMore()
 
-  $scope.toggleMembershipAdmin = (membership) ->
+  @userIsAdmin = =>
+    CurrentUser.isAdminOf(@group)
+
+  @loadMore = =>
```
robguthrie: I don't think (bordering on sure that) you don't need fat arrow on the method definition line. Is there a reason or just playing safe?

PR #2056, opened 2015-03-28
change vote

rdbartlett:
I'll look at the rest of the vote form tomorrow


rdbartlett:
Something weird is happening on the votes table... occasionally the elements get out of sequence in the screenreader, so the votes don't make sense anymore. e.g. see this thread http://localhost:8000/d/TxOiVH7G and cycle thru all the positions

PR #2057, opened 2015-03-28
Proposal form

robguthrie:
remove datetime picker, using native date widget on browsers that support it, and simple text field on the rest.

Sill to do:
Show error styles
Show validator error messages

Think about polyfill date picker for browsers that don't support


robguthrie:
Safari (no date support)
image

robguthrie:
Chrome (date support)

image

PR #2060, opened 2015-03-30
Enabling translation of new help docs

HSalmon:
Hey @rdbartlett, @mixmix - not quite sure what needs to happen next with this (if anything). Thoughts?


rdbartlett:
wow great effort. I'm down a hole at the moment so I will leave it to Mix to QA and then we should be good to go

robguthrie:
Hannah I would say you should extract out helpers for titles to remove repetition and inconsistencies.
help_h3 'some_translation_name'

Happy to pair to try this out if you like.. but yea.. there are 2 or 3 patterns which I think should be extracted into methods, as we start to build up help logic.

robguthrie:
Wow.. you've done sooo well on this large job. Amazing!

robguthrie:
Hey.. I would not say that my feedback needs to be considered before merge. I'm happy to merge this right now unless there is a major

robguthrie:
Hey I'm pretty sure the right thing to do mid term is to not have the fragile connection between the document structure and the content. IE I think keeping two massive files in sync while making tiny changes that are hard to track is a place for improvement as soon as possible.

I'm sure we can write a helper function that allows us to put the content of the strings right into the document, but is still translatable. That would be a lot better eh?

``` 
%p = lmo_t(:some_translation_key, 'The content of the translation string', optional_args) 
```

lmo is the new prefix I like to give things, and lmo_t is a function that, in development mode will build you a .yml file called '/config/locales/generated/help.yml' at the same time it renders the page, and in production mode will just call `t :some_translation_key`. Then we just commit the help.yml into the source before deploy and we've got inline translations! 

Can we make this happen, @HSalmon, @mixmix?

HSalmon:
Hey @mixmix, do you wanna QA?

mixmix:
@HSalmon I've spun this up and skimmed the code. Looks really great! 
Awesome job, that's a total mammoth :elephant: of a piece of work.

:sunflower: 

**mergemerge merge!!!**

config/locales/help.en.yml
```diff
@@ -0,0 +1,737 @@
...
+        invited_into_group_html: '<a href="#invited-into-group">When you are invited into a Loomio group</a>'
+        joining_existing_group_html: '<a href="#joining-existing-group">When you want to join an existing Loomio group</a>'
+        single_sign_on_html: '<a href="#single-sign-on">Single sign-on</a>'
+      group_settings:
+        title_html: '<a href="#group-settings">Group settings</a>'
+        group_profile_html: '<a href="#group-profile">Group profile</a>'
+        group_privacy_html: '<a href="#group-privacy">Group privacy</a>'
+        group_permissions_html: '<a href="#group-permissions">Group permissions</a>'
+      coordination:
+        title_html: '<a href="#coordination">Coordinating your group</a>'
```
gdpelican: I guess I don't see why we need to include the anchor tags and their href's in these translations... seems like a hurdle for translators that doesn't need to be there.

Why not put the anchors in the haml, and the translations in the yml file?

```
%a{href: '#group-settings'} = t :"group_settings.title_html"
```

HSalmon:
@gdpelican, @mixmix, I've cleaned this up a bit. Is it good to merge d'you reckon?

robguthrie:
I reckon it's top notch, @HSalmon. Amazing work.

gdpelican:
I reckon this is pretty great.

PR #2065, opened 2015-03-30
Attempt at a dynamically maintained shortcut modal

gdpelican:
screenshot from 2015-03-30 16 14 50
This adds a dynamically updating modal with a list of all keyboard shortcuts registered in the app


gdpelican:
While very cool, there's too many holes in this solution to really be useful; it's less work to just maintain this list manually, I reckon.

robguthrie:
Good call. Thanks for giving it a go.

PR #2070, opened 2015-03-31
Add last read indicator

gdpelican:
Adds last read comment indicator.

Now with super special bonus lintel indicator as well!

Screenshot:
screenshot from 2015-03-31 11 21 16

### DIFF
--- a/lineman/app/components/thread_page/activity_card/activity_card.scss
+++ b/lineman/app/components/thread_page/activity_card/activity_card.scss
@@ -7,3 +7,13 @@
   line-height: 30px;
   span{text-decoration: underline;}
 }
+
+.activity-card-last-read {
+  margin: 10px 0;
+  padding-bottom: 0;
+  border-bottom: 1px solid $loomio-orange;
+  text-align: right;
+  @include fontTiny;
+  line-height: 9px;
+  color: $loomio-orange;
+}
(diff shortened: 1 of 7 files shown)


rdbartlett:
pretty cool! just checked it out locally. 

thoughts: 

* should not display if there are no unread items.
* I don't understand the rationale for adding the caret to the lintel? Lintel does one thing well, unread line does one thing well, smushing the two together makes me feel confused about both.

rdbartlett:
Also I think this somehow broke it because I can see the lintel when I'm on the dashboard.

gdpelican:
Hrm, never saw that dashboard lintel; weird. Anyway I took the caret out; I find it really useful (how much of this thing have I read so far?) but understand that it's a little difficult to explain what it is easily. Maybe we'll iterate it in at some point.

robguthrie:
I'm happy to consider iterating it at some point. Thanks James.

lineman/app/components/thread_page/activity_card/activity_card.scss
```diff
@@ -6,4 +6,14 @@
   @include metaText;
   line-height: 30px;
   span{text-decoration: underline;}
-}
\ No newline at end of file
+}
+
+.activity-card-last-read {
```
robguthrie: Hey James.. I'm trying to move the SCSS to pretty strict BEM syntax. Here Block is activity-card and Element is last-read.. so should be separated with double underscore. Just a note for next time.

robguthrie: I promise it gets nice to read after a while

PR #2071, opened 2015-03-31
standardise hover states on all dropdowns

rdbartlett:
All dropdowns work the same now, apart from the group privacy dropdown which is waiting to be turned into a form just like the volume dropdown.


rdbartlett:
In fact anywhere we use "dropdown-menu-with-labels" should be a form.

robguthrie:
Cool stuff. Thanks Rich!

PR #2074, opened 2015-03-31
Dashboard thread activity

gdpelican:
Not ready for merge yet!


app/controllers/api/discussions_controller.rb
```diff
@@ -1,20 +1,16 @@
...
-    @discussions = grouped inbox_threads.group_by(&:organization_id)
+  def dashboard_by_date
+    @discussions = page_collection dashboard_threads
     respond_with_discussions
   end
 
-  def inbox_by_group
-    @discussions = grouped inbox_threads.group_by(&:group_id)
+  def dashboard_by_group
+    @discussions = grouped dashboard_threads.group_by(&:organization_id)
```
robguthrie: Hey @gdpelican why do you group here and group on the client side too?

app/controllers/api/events_controller.rb
```diff
@@ -1,4 +1,15 @@
 class API::EventsController < API::RestfulController
+  include DashboardHelper
+
+  def dashboard_by_date
+    @events = page_collection dashboard_events
+    respond_with_collection
+  end
+
+  def dashboard_by_group
+    @events = grouped dashboard_events.group_by(&:organization_id)
```
robguthrie: I'm starting to think that these should be the same API call.. I don't understand why you group on the server side.

gdpelican: Grouping on the server side is the only way to guarantee we return the
right amount of threads for each group.

If I have threads A - Z, sorted desc by date, and groups 1-3, and threads
XYZ are all in group 3, I'm not going to return then with the API call
unless I do a grouping server side.

gdpelican:
Gonna put this stuff on hold for the moment until we have a better idea of if we really want this stuff on the dashboard.

PR #2083, opened 2015-04-03
Don't display slash when using search shortcut

gdpelican:


gdpelican:
I don't want to merge this yet; I think the behaviour's still a little erratic.

robguthrie:
hey did you see the link about html keybindings? I think it will cover this kind of case without JS

gdpelican:
Hm, I did have a look; I was a little sad that the way to activate access keys isn't consistent cross browser (firefox is CTRL+ALT+key, chrome is ALT+key, and we wouldn't be able to support not using modifier keys), so it's gotta be multiple key presses and different for different browsers.

Or I'm missing something maybe? I found the docs on the stuff kinda light.

robguthrie:
I did not look closely enough.. I'll take your word for it.

PR #2085, opened 2015-04-04
Locales help

mixmix:
@HSalmon look what you done!

also I pushed a button on transifex which allows it to use its _'translation memory'_ and tap into exact phrase matches .... 

all the empty translations just got updated a bit 
:crazy:


mixmix:
@HSalmon look people love it, and are translating it already: 


german, greek, hungarian, portuguese

mixmix:
@robguthrie  this is ready to go. 
I've enabled the translated help page, and taken the switch out of the help_controller that was serving two different versions. 

If you still want switching, a better option might be to check the `TRANSLATION_COVERAGE` for help docs in that particular language which can be accessed via this module: 

```ruby
module Loomio
  class I18n
    
...

    TRANSLATION_COVERAGE = (YAML.load_file Rails.root.join('.translation_coverage.yml')).with_indifferent_access
  end
end
```

app/controllers/help_controller.rb
```diff
@@ -1,10 +1,5 @@
 class HelpController < ApplicationController
```
mixmix: here's a bookmark to make it easier to find the help_controller:

note I renamed the views   :   
help > old_help
en_help > help

so that this controller just renders the help doc

robguthrie: Why are you doing this?

robguthrie: sorry, got confused. looks good

robguthrie:
I expected that that giving the translators more would increase engagement.. is that what's going on?

mixmix:
I think we just have some awesome super vigilant people

PR #2086, opened 2015-04-04
Add cancel reply option when replying to comment

gdpelican:


gdpelican:
I think this is the UI we want for this, although 'Cancel Reply' might not be the perfect copy for it.

PR #2089, opened 2015-04-05
Adds appropriate error messages for 403/404/500 responses

gdpelican:


robguthrie:
Hey @gdpelican I don' think it's right to change the page the user is on when these events happen. I think a error flash message would be better. Otherwise we are likely to lose their state and stuff.

gdpelican:
We don't change the page they're on at all.. it simply swaps out the
ng_viewport with an error pane.

robguthrie:
So If for some random reason my comment fails to post. How do I get my
comment back?

robguthrie:
Becasue if this swaps out the viewport, I can't use the back button or
anything to get it back.. it's lost. I think it's better to use an error
flash, then the user can copy their text out and do what they need to do.

gdpelican:
That's not covered by this; this only covers the case where visiting a
thread / group page results in a 403 / 404 / 500; we'll have to go through
and come up with sensible failure cases for all of the requests we make,
but this is just the low-hanging fruit of that work.

lineman/app/components/thread_page/thread_page_controller.coffee
```diff
@@ -9,6 +9,8 @@ angular.module('loomioApp').controller 'ThreadPageController', ($routeParams, $l
     @group = @discussion.group()
     $rootScope.$broadcast('viewingThread', @discussion)
     MessageChannelService.subscribeTo("/discussion-#{@discussion.key}", @onMessageReceived)
+  , (error) ->
```
gdpelican: We only broadcast the 'pageError' event when the initial 'Get me the thread / group' request fails. We'll have to broadcast other types of events for other types of failures, I'd imagine.

PR #2091, opened 2015-04-07
Set flash from rails if one exists

gdpelican:


robguthrie:
FYI: there will be notice, alert and maybe other cases.

gdpelican:
Kay, I went ahead and cleaned up the way we do flash messages, which was decently bad and is now pretty good.

Defines four levels:
```
FlashService.success
FlashService.info
FlashService.warning
FlashService.error
```

Which accepts a translation value and translation options (or just a simple string to display, in the case of symbols or things already translated on the server side)

I also switched it to a $broadcast / $on system, which is more reliable I think and more consistent with the way we do things in the rest of the app.

gdpelican:
:flashlight:!

lineman/app/components/flash/flash.coffee
```diff
@@ -3,4 +3,24 @@ angular.module('loomioApp').directive 'flash', ->
...
+      $timeout.cancel @pendingDismiss if @pendingDismiss?
+      @pendingDismiss = $timeout @dismiss, 2000
+
+    @modalIsVisible = ->
+      angular.element('.modal').hasClass('in')
+
+    @display = => @flash and (@modal == @modalIsVisible())
+    @dismiss = => @flash = null
+
+    FlashService.success window.Loomio.flash.success if window.Loomio.flash.success?
```
robguthrie: Maybe this init block should be moved to app.coffee in the future? Anyway this whole thing looks great. Thanks @gdpelican.

gdpelican:
Yeah! I tried putting it there, but had that code run before the
`$scope.$on 'flashMessage'` event had been defined, so it didn't catch it.
Wonder if there's a better place, or a way to switch up the order in which
things get loaded?

PR #2095, opened 2015-04-08
New group setup flow

rdbartlett:
Ready for QA


rdbartlett:
Needs smoke test then please note the time of deploy for experiment purposes pls.

app/controllers/start_group_controller.rb
```diff
@@ -0,0 +1,51 @@
+class StartGroupController < ApplicationController
```
gdpelican: Why inherit from ApplicationController instead of BaseController?

app/controllers/start_group_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      render :new
+    end
+  end
+
+  def create_signed_out
+    @group = Group.new(permitted_params.group)
+    @email = params[:email]
+    @name =  params[:name]
+    @errors = []
+    @errors << 'name' if @name.blank?
```
gdpelican: Whoa.. shouldn't these be happening as validations on the model?

app/models/permitted_params.rb
```diff
@@ -1,4 +1,4 @@
-class PermittedParams < Struct.new(:params, :user)
+class PermittedParams < Struct.new(:params)
```
gdpelican: Heh, I didn't know you could do `PermittedParams < Struct.new`; I'd always seen it as `PermittedParams = Struct.new`, cool.

app/views/start_group/new.html.haml
```diff
@@ -0,0 +1,59 @@
...
+          .col-sm-9
+            - if @errors.any?
+              .form__error= t :"start_group_form.form_error"
+
+        - unless user_signed_in?
+          .form-group
+            .col-sm-3
+              %label.control-label{for: 'name'}= t :"start_group_form.name"
+            .col-sm-9
+              - if @errors.include? 'name'
```
gdpelican: Hm.. Rails should be doing all of this for you, I'm not sure I quite understand why we're handling errors so manually here.

app/controllers/start_group_controller.rb
robguthrie: There is no name and email on the group model.

app/controllers/start_group_controller.rb
robguthrie: Because base controller is for signed in things only.

app/views/start_group/new.html.haml
robguthrie: It's easier than trying to get rails to do this, because name and email are not part of group model.

app/controllers/start_group_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    @errors << 'email' unless /^[^@]+@[^@]+\.[^@]+$/.match @email
```
gdpelican: We must have a better email validation laying around than this? Makes me really really nervous to have this one-off validation on the controller as well.

gdpelican: These should happen on the Invitation model, I reckon. I'd feel a lot better if this controller method looked more like

```
@group = Group.new(permitted_params.group)
@invitation = Invitation.new(name: params[:name], email: params[:email]) # xtra credit for permitted_params.invitation
if @group.valid? && @invitation.valid? # perform email / name validations here
  StartGroupService.start_group(group: @group, admin_invitation: @invitation)
else
  @errors = {}.extend @group.errors, @invitation.errors
end
```

gdpelican: Okay, I feel a little left out of this process here; this code is not good, and I don't think this was urgent enough to warrant an off-handed 'it's fine', and deploy.

PR #2098, opened 2015-04-09
Updates on Translations

gbrego:
Hello Loomio Team, 

This is a simple commit for a message that was not included on languages YML files.

Thanks in advance.

### DIFF
--- a/app/controllers/discussions_controller.rb
+++ b/app/controllers/discussions_controller.rb
@@ -27,7 +27,7 @@ class DiscussionsController < GroupBaseController
                                 params: permitted_params.discussion,
                                 actor: current_user)
       current_user.update_attributes(uses_markdown: @discussion.uses_markdown)
-      flash[:notice] = 'Discussion was successfully updated.'
+      flash[:notice] = t("success.discussion_updated")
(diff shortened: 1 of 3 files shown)


gdpelican:
Hey, thanks @gbrego! Surprised we haven't caught this one before.

Hey @mixmix, is it ok to merge the pt-BR translations like this, or will that muck with Transifex somehow?

robguthrie:
@mixmix really need you to reply to us.

mixmix:
awesome catch @gbrego 

---

Hey this is safe to merge. 
the pt-BR.yml file will get over-written when we update, it will need to be changed in Transifex. 
Respectfully, we need to all understand this simple idea so that I'm not a hold-up when I'm away for a conference / having a weekend. :heart:

robguthrie:
For clarity @gdpelican: Yes, any updates to non 'en' locales will be overwritten.

gdpelican:
That's fine, I just wanted to make sure that we wouldn't have to deal with
a merge conflict or something down the line for committing to non-en files.
I'll put it in the knowledge bank.

PR #2101, opened 2015-04-10
add is_referral to groups

rdbartlett:
I think this is right but I have been a bit distracted so it needs a proper QA before deploy

### DIFF
--- a/app/controllers/start_group_controller.rb
+++ b/app/controllers/start_group_controller.rb
@@ -20,6 +20,8 @@ class StartGroupController < ApplicationController
 
     # check for valid name and email
     if @group.valid? and @errors.empty?
+      @group = Group.new(permitted_params.group)
+      @group.is_referral = true
       StartGroupService.start_group(@group)
       @group.add_admin!(current_user)
       redirect_to @group
@@ -30,6 +32,7 @@ class StartGroupController < ApplicationController
 
   def create_signed_out
     @group = Group.new(permitted_params.group)
+    @group.is_referral = false
     @email = params[:email]
     @name =  params[:name]
     @errors = []
--- a/db/migrate/20150410025815_add_is_referral_to_group.rb
+++ b/db/migrate/20150410025815_add_is_referral_to_group.rb
@@ -1,0 +1,6 @@
+class AddIsReferralToGroup < ActiveRecord::Migration
+  def change
+    add_column :groups, :is_referral, :boolean
+    Group.reset_column_information
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150324073159) do
+ActiveRecord::Schema.define(version: 20150410025815) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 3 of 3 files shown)


db/migrate/20150410025815_add_is_referral_to_group.rb
```diff
@@ -0,0 +1,9 @@
+class AddIsReferralToGroup < ActiveRecord::Migration
+  def change
+    add_column :groups, :is_referral, :boolean
+  end
+
+  Group.parents_only.where("created_at < '2015-04-9 1:00:00'").each do |g|
```
robguthrie: Group.parents_only.where("created_at < '2015-04-9 1:00:00'").where(group_request_id: nil).update_all(is_referral: true)

This will do all groups in one query on the db, so no long runnning migrations. Please test.

db/migrate/20150410025815_add_is_referral_to_group.rb
```diff
@@ -0,0 +1,9 @@
+class AddIsReferralToGroup < ActiveRecord::Migration
+  def change
+    add_column :groups, :is_referral, :boolean
+  end
+
```
robguthrie: add Group.update_column_information

app/controllers/start_group_controller.rb
```diff
@@ -22,6 +22,7 @@ def create_signed_in
     if @group.valid? and @errors.empty?
       StartGroupService.start_group(@group)
       @group.add_admin!(current_user)
+      @group.update_attribute(:is_referral, true)
```
robguthrie: Please just do
````
@group = Group.new(permitted_params.group)
@group.is_referral = true
````
at the top of the method

File: app/controllers/start_group_controller.rb [hunk 4]
```diff
@@ -44,6 +45,7 @@ def create_signed_out
       StartGroupService.invite_admin_to_group(group: @group,
                                               name: @name,
                                               email: @email)
+      @group.update_attribute(:is_referral, false)
```
robguthrie: isn't the default false?

robguthrie: so can remove the line?

rdbartlett: I wasn't sure how to handle defaults because of the stupid subgroups

db/migrate/20150410025815_add_is_referral_to_group.rb
rdbartlett: you mean reset_column_information?

app/controllers/start_group_controller.rb
rdbartlett: i.e. I think it is actually good to use true, false, and nil because I want a subgroup.is_referral to return nil

db/migrate/20150410025815_add_is_referral_to_group.rb
rdbartlett: group_request_id is not a column on Group though

rdbartlett:
run on deploy:

```    
non_referral_group_ids = GroupRequest.pluck(:group_id).compact;
Group.parents_only.where('groups.id NOT IN (?)', non_referral_group_ids).update_all(is_referral: true)
``

rdbartlett:
I've tested locally, should be good to go. Let's let Travis run first though. xx

PR #2102, opened 2015-04-12
Dashboard refactor

gdpelican:
Not ready for merge.

A refactor of the way the dashboard page works. I think this is better but could be wrong.

### DIFF
--- a/lineman/app/components/dashboard_page/dashboard_page.haml
+++ b/lineman/app/components/dashboard_page/dashboard_page.haml
@@ -71,4 +71,4 @@
             %loading{ng-show: 'dashboardPage.loadMoreExecuting'}
           .dashboard-page-sort-date{ng-if: 'dashboardPage.sort() == "sort_by_group"'}
             .dashboard-page-group{ng-repeat: 'group in dashboardPage.dashboardGroups() | orderBy:dashboardPage.groupName'}
-              %dashboard_page_thread_collection{group: 'group', threads: 'dashboardPage.dashboardDiscussions', filter: 'dashboardPage.inGroup(group)'}
+              %dashboard_page_group_collection{group: 'group', threads: 'dashboardPage.dashboardDiscussions'}
--- a/lineman/app/components/dashboard_page/dashboard_page.haml
+++ b/lineman/app/components/dashboard_page/dashboard_page.haml
@@ -62,16 +62,13 @@
             %span{translate: 'dashboard_page.no_threads'}>
             %a{ng-show: 'dashboardPage.filter() != "show_all"', translate: 'dashboard_page.view_recent', href: '#', ng-click: 'dashboardPage.changePreferences({dashboard_filter: "show_all"})'}>
           .dashboard-page-sort-group{ng-if: 'dashboardPage.sort() == "sort_by_date"'}
-            %dashboard_page_thread_collection{name: 'Older', threads: 'dashboardPage.dashboardDiscussions()', filter: 'dashboardPage.older'}
+            %dashboard_page_thread_collection{name: 'today', threads: 'dashboardPage.dashboardDiscussions', filter: 'dashboardPage.today'}
+            %dashboard_page_thread_collection{name: 'yesterday', threads: 'dashboardPage.dashboardDiscussions', filter: 'dashboardPage.yesterday'}
+            %dashboard_page_thread_collection{name: 'this_week', threads: 'dashboardPage.dashboardDiscussions', filter: 'dashboardPage.thisWeek'}
+            %dashboard_page_thread_collection{name: 'this_month', threads: 'dashboardPage.dashboardDiscussions', filter: 'dashboardPage.thisMonth'}
+            %dashboard_page_thread_collection{name: 'older', threads: 'dashboardPage.dashboardDiscussions', filter: 'dashboardPage.older'}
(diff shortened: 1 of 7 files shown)


lineman/app/components/dashboard_page/dashboard_page_controller.coffee
```diff
@@ -2,39 +2,29 @@ angular.module('loomioApp').controller 'DashboardPageController', ($rootScope, R
...
   @sort   = -> CurrentUser.dashboardSort
   @filter = -> CurrentUser.dashboardFilter
 
+  @loadedCount = =>
+    @loaded[@sort()] = @loaded[@sort()] or {}
+    @loaded[@sort()][@filter()] = @loaded[@sort()][@filter()] or 0
+
+  @updateLoadedCount = =>
+    current = @loadedCount()
+    @loaded[@sort()][@filter()] = current + @perPage[@sort()]
```
gdpelican: I haven't reintegrated the loadedCount() into the interface yet; it's primarily designed to limit us displaying a bunch of threads we've loaded in other places on the dashboard (which I much prefer to control; if you've been navigating around the app a bunch and land on the dashboard, I'd rather there be 25 threads there rather than 500, and just load more as you need em.)

gdpelican: (The primary thing I don't like is going to the dashboard and seeing this:

```
Today:
Thread 1
Thread 2
Thread 4

Yesterday:
Thread 5
```
Then loading more and having that thread 3 slide into 'today' unannounced; makes certain things difficult to find.

lineman/app/components/dashboard_page/dashboard_page_group_collection.haml
```diff
@@ -0,0 +1,9 @@
+%section.dashboard-page-threads-container{ng-show: 'any()', role: 'region', aria-label: "{{ name | translate }}"}
+  .dashboard-page-group-name{aria-hidden: 'true'}
+    %img.selector-list-item-group-logo.pull-left{src: "{{group.logoUrl()}}", aria-hidden: 'true'}>
+    %h2.dashboard-page-group-name
+      %a{href: '/g/{{group.key}}'} {{group.name}}
```
rdbartlett: This is an accessibility anti-pattern that I introduced that I would like to squelch ASAP. Hiding the H2 was a supremely dumb decision I made while my brain was on the shelf. A better version (needs test) would be:

```
%section.dashboard-page-threads-container{ng-show: 'any()', role: 'region', aria-labelledby: "{{ name }}"}
  .dashboard-page-group-name
    %img.selector-list-item-group-logo.pull-left{src: "{{group.logoUrl()}}", aria-hidden: 'true'}>
    %h2.dashboard-page-group-name{id: '{{ name }}'}
      %a{href: '/g/{{group.key}}'} {{group.name}}

```

[See here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) for explanation of aria-labelledby.

lineman/app/components/dashboard_page/dashboard_page_thread_collection.haml
```diff
@@ -0,0 +1,4 @@
+%section.dashboard-page-threads-container{ng-show: 'any()', role: 'region', aria-label: "{{ name | translate }}"}
```
rdbartlett: Same as above

lineman/app/components/dashboard_page/dashboard_page.haml
```diff
@@ -62,35 +62,13 @@
...
-            %section.dashboard-page-threads-container{ng-if: 'dashboardPage.anyThisMonth()', role: 'region', aria-label: "{{ 'dashboard_page.threads_from.this_month' | translate }}"}
-              .dashboard-page-date-range{translate: 'common.date_range.this_month', aria-hidden: 'true'}
-              .dashboard-page-threads
-                %dashboard_page_thread{discussion: 'discussion', ng-repeat: 'discussion in dashboardPage.dashboardDiscussions() | filter:dashboardPage.thisMonth'}
-            %section.dashboard-page-threads-container{ng-if: 'anyOlder()', role: 'region', aria-label: "{{ 'dashboard_page.threads_from.older' | translate }}"}
-              .dashboard-page-date-range{translate: 'common.date_range.older', aria-hidden: 'true'}
-              .dashboard-page-threads
-                %dashboard_page_thread{discussion: 'discussion', ng-repeat: 'discussion in dashboardPage.dashboardDiscussions() | filter:dashboardPage.older'}
+            %dashboard_page_thread_collection{name: 'today', threads: 'dashboardPage.dashboardDiscussions', filter: 'dashboardPage.today'}
```
robguthrie: I had been thinking that you would just pass 'today' into the component... and the component itself would setup a lokijs view of the right section of records. Then you don't need to use an expensive filter and the "what to display" logic is kept within the component, and you don't need to tell the component how to do it's job.

What do you think about that?

gdpelican:
I have hesitations about doing that way, because I think it's an anti-pattern to start increasing the number of requests in order to load the page; making each component shoot out a request on page load instantly increases the number of requests needed to load the by_date dashboard from 1 to 5, and it's even worse for by_group, since a user will have N groups.

gdpelican:
In any case, I'd like to shelf this and see what comes out of the design sessions for dashboard before continuing.

PR #2106, opened 2015-04-13
Convert group privacy dropdown to form

gdpelican:
Converts group privacy dropdown to a form.


rdbartlett:
We'll sort this out in another accessibility pass soon, but FYI @gdpelican the .sr-only stuff currently seems to be a bit of an anti-pattern too. Seems that JAWS totally ignores it, at least in the first test. Solution coming soon!

robguthrie:
I think the only way we'll communicate the right a11y patterns is by
ensuring that we keep all the code we can upto date. Might be a good idea
to process the a11y card soon.

rdbartlett:
Yep agreed. I reckon it's next week's job for me

PR #2107, opened 2015-04-13
Use ng-src for attachment/user images

gdpelican:
This fixed all of the image 404's in angular for me locally; problem was using the `src` attribute instead of `ng-src`

There's likely some lingering issues around things like default group logos, I'll check that out next.


gdpelican:
Yeah, can confirm we're not loading the default group images correctly; do we have images for these hanging around somewhere @rdbartlett @robguthrie?

PR #2108, opened 2015-04-13
Messaging for empty discussions list on group page.

gdpelican:
Please don't merge, I really need to write e2e tests for this.

### DIFF
--- a/app/controllers/angular_support_controller.rb
+++ b/app/controllers/angular_support_controller.rb
@@ -1,5 +1,5 @@
 class AngularSupportController < ApplicationController
-  before_filter :prevent_production_destruction
+  around_filter :ensure_testing_environment
 
   USER_PARAMS = {name: 'Patrick Swayze',
                  email: 'patrick_swayze@loomio.org',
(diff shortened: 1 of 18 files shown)


gdpelican:
Sigh, it seems like every time I try to spin up the e2e tests, they take forever to run and (mostly) fail, which makes writing new tests really tough. (And this is something I would really rather write tests for than just shrug and hope it works.)

I think I'm going to put some time in either this weekend or early next week to figure out why I'm having these issues, and what we can do to make writing and running these tests quicker and less painful.

robguthrie:
I put a bit of time into the tests earlier this week. I personally found it
really rewarding.. did not take a lot to get a bunch working again, and the
BEM syntax + protractor felt really right. They don't all pass but it
should be pretty trivial to fix the rest of them. I don't find it a big
problem to run them.. they're faster than the rails cukes, i really hope
someone buys you a new computer soon.. Also Its 3rd on my list to get these
working on travis, It's aparently possible to run both ruby and node in a
single travis project. Anyway.. write the tests even without the
implementation, maybe I can fill them in for you.

gdpelican:
Here's a log of the errors I'm getting running the tests on a fresh install (and a new compy):
https://gist.github.com/gdpelican/b2a0fcef9aeef810f0b2

gdpelican:
Okay, this should be good to go.

Fixes in here:
- Allows API to properly load discussions for a single group
- Allows API to properly load discussions for groups the user is not a member of
- Adds a loading widget to the group page, the same as we do for thread page
- Added tests around messaging for empty discussion list (I've also hand-tested all the use cases)
- Some speed improvements for e2e testing

PR #2109, opened 2015-04-13
Hacky fix for default group image 404s

gdpelican:


gdpelican:
This isn't hacky anymore; hooray!

PR #2112, opened 2015-04-14
Focus inner anchor when highlighting a search result

gdpelican:


gdpelican:
Think this should do it; can confirm that calling `focus()` there sets `document.activeElement` to that search result anchor

PR #2118, opened 2015-04-21
locales update

robguthrie:


mixmix:
@robguthrie  this is ready to merge.

just updated OS - having trouble spinning up Loomio. Looks like you made a large change to the frontpage translations?

PR #2119, opened 2015-04-21
Update FAQ & add to contact page

HSalmon:


HSalmon:
contact-us

robguthrie:
Looking good. Can we please change 'get in touch to talk pricing' to just 'Contact us if you're interested'.

Also can we change the group size limit text? I'm not quite sure what to say.. Something like we have a limit of 5000 members but if your group outgrows this we'll be sure to get in contact with you. 

Do we need to think about the page in terms of non loomio.org versions? What should it do then? redirect to sign_in?

@rdbartlett any thoughts?

gdpelican:
Best IMO is:

- render 'help/faq' and social media links if hosted_by_loomio?
- Move the contact email (contact@loomio.org) into an ENV variable, so
folks can specify where they'd like their contact mail sent.

HSalmon:
@robguthrie, @rdbartlett, @gdpelican - should I proceed with the suggestion above made?

robguthrie:
yea, i think that's good.

rdbartlett:
**What size groups work well on Loomio?**

Loomio currently works well for groups up to about 1000 people.

HSalmon:
@robguthrie, @gdpelican - does this need any further modifications?

robguthrie:
nope. let's do it!

gdpelican:
:+1: The best!

rdbartlett:
Nice on Salmon :fish:

PR #2127, opened 2015-04-29
try to be sure that we delete the comment votes when deleting the comment

robguthrie:


app/services/comment_service.rb
```diff
@@ -43,6 +43,8 @@ def self.create(comment:, actor:, mark_as_read: true)
 
   def self.destroy(comment:, actor:)
     actor.ability.authorize!(:destroy, comment)
+    # paranoid destroy_all because comment votes seem to be dangling around.
+    comment.comment_votes.destroy_all
```
gdpelican: Can we do `delete_all` so it doesn't potentially take forever?

robguthrie: yea.. good idea..

robguthrie: actually .. because comment_votes have events which have notifications.. it
might be a bit involved to delete_all manually for everything concerned.

PR #2128, opened 2015-04-30
Fix headings, add labels to sections

HSalmon:


app/views/help/_contents.html.haml
```diff
@@ -1,141 +1,142 @@
...
-  .row
-    .col-sm-3
-      %ul
+%section{"role" => "region", "aria-labelledby" => "table-of-contents"}
+  #contents
+    %h2#table-of-contents= t :"new_help.contents.title"
+    .row
+      .col-sm-3
+        %ul
+          %p.single-item
```
rdbartlett: Suggest that all these `.single-item` and `.top-level` elements should be `%li`'s with `h3`'s inside them. So like:

```
         %li.single-item
           %h3
             %a{href: "#how-loomio-works"}= "How Loomio works"
         %li.top-level
           %h3
             %a{href: "#getting-started"}= t :"new_help.contents.getting_started.title_html"
         %li
           %a{href: "#starting-new-group"}= t :"new_help.contents.getting_started.starting_new_group_html"
```

Also I wonder if you need the class selector at all, or if you can use style rules like `#contents %ul %li %h3{}` but I don't really have a strong view one way or the other on that.

rdbartlett:
Looks great Hannah. Passes my code review at least :) Are you happy for this to be deployed?

PR #2129, opened 2015-04-30
Fix urls

robguthrie:


robguthrie:
This is going to take another hour or two to fix up. It fixes a bug whereby discussions would not be linked to the correct subdomain.

robguthrie:
latest test shows that updating a discussion is broken under this.. need to investigate.

gdpelican:
What's the status on this @robguthrie? Did this make it into master in another form already?

robguthrie:
nope.. there is one bug to fix - updating discussions is broken

PR #2131, opened 2015-04-30
Preserve votes on open proposals when deleting membership

HSalmon:


HSalmon:
@robguthrie, I think this is ready to go?

robguthrie:
Yea.. I guess so. Do you think it's the behaviour we want?\

rdbartlett:
Yes I think this is the right thing. Deleting votes seems dumb and bad and I would need to hear a good reason for doing it.

PR #2132, opened 2015-04-30
locales update

mixmix:
ready to go @robguthrie !


mixmix:
wow ... i think that's nati totally killing it with the help docs

robguthrie:
thanks mixy!

PR #2133, opened 2015-04-30
Add remote db rake tasks

gdpelican:


gdpelican:
Whoops, don't merge me just yet, I don't quite work.

robguthrie:
this feels like it should be a separate gem to me.

gdpelican:
I guess, but I'm not going to write a gem for it, I'm just going to use these commands. :)

robguthrie:
We should be writing a function to create a safe database for developers to use, rather than using the production db. 3 Devs have production and I don't think any more should, so I won't accept this. Hope you agree about that in principal. Maybe we should fast track the sanitise production function?

PR #2135, opened 2015-05-01
Remove restrictions on group size

HSalmon:


HSalmon:
@robguthrie, I think this is also good to go?

robguthrie:
Thank you, Hannah!!

PR #2137, opened 2015-05-04
Dashboard refactor

gdpelican:
Refactor? I hardly know her! :bomb:


app/controllers/api/discussions_controller.rb
```diff
@@ -2,22 +2,16 @@ class API::DiscussionsController < API::RestfulController
...
 
-  def inbox_by_date
-    load_and_authorize_group if params[:group_id]
+  def discussions_for_dashboard
     @discussions = page_collection inbox_threads
     respond_with_discussions
   end
 
-  def inbox_by_organization
+  def discussions_for_inbox
```
rdbartlett: I am happy to be ignored as a pedant but: is there a reason we're mixing 'threads' and 'discussions' in these methods?

gdpelican: There's a bit of a fuzzy line. I'm trying to keep things on the angular
side called 'threads', while everything in existing loomio is a
'discussion'. Happy to look at moving the name transition line a little
(maybe no calling things 'threads' once we get to the controller) but fact
is we're always going to have a bit of an odd spot where we do something
like

```
def get_threads
  Discussion.find_stuff
end
```

rdbartlett: Yeah fair enough. Just FYI, my preference is to use 'threads' as much as possible, because I assume the threads of the future will have lots of non-discussion components. But I have no desire to argue the case. Carry on :)

gdpelican: We had/have a similar situation with 'motions' vs 'proposals', and kinda
settled on calling them 'proposals' in the angular app, and still 'motions'
on the rails side because changing all the names in the existing app is
time-consuming, error-prone, and doesn't really do anything necessarily. (I
tried a full-on rename of Motion => Proposal for a bit a while back, and
nah not worth it.)

robguthrie: We also call them proposals in the rails app in places too. My advise is to
use threads where ever easy and possible, otherwise discussions. I'm happy
for it to be a dumb confusing thing as long as we keep pushing in the right
direction, ie: using threads where possible.

robguthrie: i find the names inbox and dashboard are confusing together.... i think
where you say dashboard, maybe you should say recent. But it's nbd rn.

gdpelican:
Closing this for now.

PR #2140, opened 2015-05-05
Print discussion

HSalmon:
screen shot 2015-05-05 at 9 35 06 pm

screen shot 2015-05-05 at 9 36 04 pm

### DIFF
--- a/app/controllers/discussions_controller.rb
+++ b/app/controllers/discussions_controller.rb
@@ -183,6 +183,16 @@ class DiscussionsController < GroupBaseController
     redirect_to @version.reify()
   end
 
+  def print
+    @discussion = Discussion.find_by_key! params[:id]
+    @group = @discussion.group
+    if @discussion.motions.present?
+      @motions = @discussion.motions.sort_by {|m| m.created_at }
+    end
+    @activity = @discussion.items if @discussion.items.present?
+    render :layout => "print"
+  end
+
   private
 
   def we_dont_serve_images_here_google_bot
--- a/app/views/discussions/print.html.haml
+++ b/app/views/discussions/print.html.haml
@@ -1,0 +1,96 @@
+.print-container
+  .row.title
+    .col-md-9
+      %h1= @discussion.title
+      .groups-title
+        - if @group.is_subgroup?
+          %h3= @group.parent.name
+          = content_tag :span, "\u25B6", class: 'name-separator'
+        %h3=@group.name
+  .row.context
+    .col-md-9
+      = @discussion.description
+    .col-md-2.col-md-offset-1
+      .discussion-details
+        = t("discussion_context.started_print",
+            when: @discussion.created_at.strftime("%d %B %Y"),
+            who: @discussion.author.name)
+  - if @motions
+    .row
+      .col-md-12
+        %h2= t :proposals
+      - @motions.each do |motion|
+        .col-md-9
+          %h3= motion.title
+          .proposal-description
+            = motion.description
+          .proposal-details
+            %ul
+              - if motion.voting?
+                %li= t(:closing) + ": " + motion.closing_at.strftime("%d %B %Y, %I.%M %p %Z")
+              - else
+                %li= t(:closed_at) + ": " + motion.closed_at.strftime("%d %B %Y, %I.%M %p %Z")
+              %li= t :proposed_by_who_on, when: motion.created_at.strftime("%d %B %Y, %I.%M %p %Z"), who: motion.author_name
+            - if motion.has_outcome?
+              %strong= t(:outcome) + ": "
+              %span
+                = motion.outcome
+        .col-md-9
+          .motion-graph.clearfix
+            %canvas.pie{id: "graph_#{motion.id}", height: 150, width: 150}
+            .legend
+              .item.yes
+                = t(:'position_verbs.yes')
+                (#{motion.yes_votes_count})
+              .item.abstain
+                = t(:'position_verbs.abstain')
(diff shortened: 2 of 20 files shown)


HSalmon:
screen shot 2015-05-05 at 9 36 04 pm

app/controllers/discussions_controller.rb
```diff
@@ -4,7 +4,7 @@ class DiscussionsController < GroupBaseController
   before_filter :we_dont_serve_images_here_google_bot
   before_filter :authenticate_user!, :except => [:show, :index]
   before_filter :load_resource_by_key, except: [:new, :create, :index, :update_version]
-  authorize_resource :except => [:new, :create, :index, :add_comment]
+  authorize_resource :except => [:new, :create, :index, :add_comment, :print]
```
robguthrie: Woha!
Big red alert!

You must authorise the resource the same way that the show action does.
Don't merge this until this is fixed. Cheers!

rdbartlett: Maybe something for the next iteration: would be cool if the pie chart made
sense in grayscale.

HSalmon:
@robguthrie, @gdpelican, would one of you mind taking a look at this to see whether things need changing before I squash? Thanks dewds!

app/controllers/discussions_controller.rb
```diff
@@ -183,6 +183,16 @@ def update_version
     redirect_to @version.reify()
   end
 
+  def print
+    @discussion = Discussion.find_by_key! params[:id]
```
gdpelican: This is way too many queries just to access discussion relations. Can we use `includes` to minimize this?

```
@discussion = Discussion.includes(:group, :motions, :items).find_by_key(params[:id])
``` 

Then you can safely do `@discussion.items` within the view without incurring another db query.

config/routes.rb
```diff
@@ -235,6 +235,8 @@
     end
   end
 
+  get 'discussions/:id/print' => 'discussions#print', as: :print_discussion
```
gdpelican: Can we slide this into the existing discussion resources like so?

```
resources :discussions do
  get :print, on: :member # this route will automagically be named 'print_discussion' 
end
```

gdpelican: Actually, in our case, we'd want it like this I think.

```
resources :discussions, path: 'd' do
 ...
  member do
    ...
    get :print
  end
end

File: app/views/discussions/print.html.haml [hunk 4]
```diff
@@ -0,0 +1,101 @@
...
+          - filter_discussion_events(@activity).each do |event|
+            - if event.kind == 'new_comment'
+              %li
+                = render 'avatar', user: event.eventable.user
+                %span
+                  %strong= event.eventable.user_name
+                %span= event.eventable.body
+                %span.timestamp= event.eventable.created_at.strftime("%d %B, %I.%M %p")
+            - else
+              - item = DiscussionItem.new(event)
```
gdpelican: I'm a bit confused by this bit; what's going on here? Seems pretty scary be initializing DiscussionItems in the view like this.

app/views/discussions/print.html.haml
```diff
@@ -0,0 +1,101 @@
...
+                    percent: motion.percent_voted,
+                    numerator: (motion.group_size_when_voting - motion.members_not_voted_count),
+                    denominator:  motion.group_size_when_voting
+            = render('motions/motion_pie_javascript', motion: motion)
+        .col-md-9
+          %h3= t :positions
+          %ul.positions
+            - if motion.has_votes?
+              - motion.grouped_unique_votes.each do |vote|
+                - icon = display_proposal_activity_icons_for_print(vote)
```
gdpelican: Don't think we have to declare this as a variable, since we're just using it once.

```
image_tag proposal_activity_icon_for(vote)
```

app/views/discussions/print.html.haml
```diff
@@ -0,0 +1,101 @@
...
+        .col-md-9
+          - if motion == @discussion.current_motion
+            %h3= t :current_decision
+          %h3= motion.title
+          .proposal-description
+            = motion.description
+          .proposal-details
+            %ul
+              - if motion.voting?
+                %li= t(:closing) + ": " + motion.closing_at.strftime("%d %B, %I.%M %p %Z")
```
gdpelican: Hm, this seems like a common thing we're running into: we're duplicating passing parameters because we have to get the right translation key.

Some possible options:
- Switch on the translation key in the view directly
```
t motion.voting? ? :closing : :closed_at, closing_at: motion.closing_at.strftime
```
- Write a helper method for each one
```
t closing_key_for(motion, closing_at: motion.closing_at)

# in helper
def closing_key_for(motion)
  motion.voting? ? :closing : :closed_at
end
```

Writing a helper method which wraps t and allows passing a true/false value and a condition (my favourite)
```
t_ternary(when_true: :closing, when_false: :closed_at, condition: motion.voting?, closed_at: motion.closed_at)

# in helper
def t_ternary(when_true:, when_false:, condition:, *args)
  t condition ? when_true : when_false, args)
end
```

gdpelican:
Great stuff Hannah! I've left a few things, but overall looks great. :rabbit:

HSalmon:
screen shot 2015-05-07 at 12 42 42 pm

rdbartlett:
Loving this. Can you show me what it looks like when there are multiple
proposals, one of which is still open?

HSalmon:
Does this have sign-off @rdbartlett, @robguthrie?

robguthrie:
yes!

PR #2141, opened 2015-05-05
Dashboard refactor deux

gdpelican:
This is a nice, error-free version of the current dashboard refactor. Ready for a git pull & look, but not ready to merge just yet.


app/controllers/api/discussions_controller.rb
```diff
@@ -2,19 +2,13 @@ class API::DiscussionsController < API::RestfulController
...
   end
 
-  def inbox_by_organization
-    @discussions = grouped inbox_threads.group_by(&:organization_id)
-    respond_with_discussions
-  end
-
-  def inbox_by_group
-    @discussions = grouped inbox_threads.group_by(&:group_id)
+  def discussions_for_inbox
```
gdpelican: These queries are finally starting to converge, although I'm not quite ready to collapse them just yet.

app/controllers/api/discussions_controller.rb
```diff
@@ -55,37 +49,19 @@ def discussion_params
...
 
   def visible_records
-    load_and_authorize_group
-    groups = if @group
-               VisibleGroupsQuery.expand(group: @group, user: current_user)
-             else
-               current_user.groups
-             end
-    Queries::VisibleDiscussions.new(user: current_user, groups: groups).order(last_activity_at: :desc)
+    Queries::VisibleDiscussions.new(user: current_user, groups: current_user.groups)
```
gdpelican: Much cleaner when we don't need to worry about grouping

config/locales/client.en.yml
```diff
@@ -16,8 +16,8 @@ en:
     date_range:
       today: 'Today'
       yesterday: 'Yesterday'
-      this_week: 'This Week'
-      this_month: 'This Month'
+      thisweek: 'This Week'
+      thismonth: 'This Month'
```
gdpelican: got fed up with needing to care about camel vs snake case for these.

db/migrate/20150504075221_populate_participating.rb
```diff
@@ -0,0 +1,16 @@
+class PopulateParticipating < ActiveRecord::Migration
+  def up
+    progress_bar = ProgressBar.create( format: "(\e[32m%c/%C\e[0m) %a |%B| \e[31m%e\e[0m ",
+                                       progress_mark: "\e[32m/\e[0m",
+                                       total: Discussion.count )
+    Discussion.reset_column_information
```
gdpelican: Okay, so in plaintext:

For each discussion
Each discussion reader for that discussion, which has a user that either
a) has an event in discussion.events
b) is the author of the discussion
Shall be set to participating in the discussion.

This took ~20min on my machine running a recent prod dump.

lineman/app/components/base.scss
```diff
@@ -6,7 +6,7 @@ body {
 }
 
 h1, h2, h3, h4{
-  @include fontReset;
+  @include fontReset; /* Y U SO RESET? */
```
gdpelican: This was weird. Why did we do this?

lineman/app/components/dashboard_page/dashboard_page.haml
```diff
@@ -2,95 +2,64 @@
   .dashboard-container
     %section.dashboard-options{aria-label: "{{ 'dashboard_page.options.aria_label' | translate }}"}
       .lmo-btn-group.lmo-btn-group-right
-          .lmo-btn-wrapper
```
gdpelican: Okay, so complete reset of the dashboard view basically. It's better. (kawaii!)

lineman/app/components/dashboard_page/dashboard_page_controller.coffee
```diff
@@ -1,97 +1,60 @@
...
+  @perPage = 25
+  @loaded =
+    show_all:           0
+    show_muted:         0
+    show_proposals:     0
+    show_participating: 0
+  @filter    = -> CurrentUser.dashboardFilter
+  @nowLoaded = -> @loaded[@filter()]
+
+  @timeframes =
```
gdpelican: The dashboard can handle two different types of views for any filter, one for timeframes and one for groups.

I didn't implement the current / past proposals thing yet... it strikes me as likely to get designed away soon, heh. >.>

lineman/app/components/dashboard_page/dashboard_page_controller.coffee
```diff
@@ -1,97 +1,60 @@
...
+  @timeframeNames = _.map @timeframes, (timeframe, name) -> name
+  @timeframeQueryFor = (name) -> @timeframes[name].view
+
+  @groupThreadLimit = 5
+  @groups = -> CurrentUser.parentGroups()
+  @groupName = (group) -> group.name
+  @groupQueryFor = (group) -> @["group#{group.id}"]
+  @moreForThisGroup = (group) -> @groupQueryFor(group, { filter: @filter() }).length() > @groupThreadLimit
+
+  @updateQueries = ->
```
gdpelican: We now calculate these views on large page events (changing a filter, loading more records, navigating to the page, etc.), and they (the _views_, not the collections) are static until the next large page event. These queries are the objects we pass to the underlying thread collection 

Timeframes are stored in
```
this.timeframes.today.view
this.timeframes.yesterday.view
etc
```

Groups are stored in
```
this.group45
this.group600
```
...but that's not great, so I'm going to namespace em in my next pass through this.
(I'm probably going to change this to group.key too, instead of group.id)

lineman/app/components/groups_page/groups_loaded.coffee
```diff
@@ -0,0 +1,15 @@
+angular.module('loomioApp').directive 'groupsLoaded', ($timeout) ->
+  restrict: 'A'
+  replace: true
+  link: (scope, element) ->
+    $timeout ->
+      leftHeight = rightHeight = 0
```
gdpelican: Hey I really like this chunk of code. Yeah it's a little javascript, but it's super light and allows the 2-column group page design to be possible in a general case, by dropping groups into a column based on which column is shorter at the time.

lineman/app/components/home/home.coffee
```diff
@@ -1,3 +0,0 @@
-angular.module('loomioApp').controller 'HomeController', () ->
```
gdpelican: I know it looks like a big PR (it is), but a lot of it is carving out old components we're not using at all.

lineman/app/components/navbar/navbar.coffee
```diff
@@ -3,5 +3,13 @@ angular.module('loomioApp').directive 'navbar', ->
...
-  link: (scope, element, attrs) ->
+  controller: ($scope, Records, ThreadQueryService) ->
+    $scope.$on 'currentComponent', (el, component) ->
+      $scope.selected = component
+
+    $scope.unreadThreadCount = ->
+      ThreadQueryService.unreadQuery().length()
+
+    if !$scope.inboxLoaded
+      Records.discussions.fetchInbox().then ->
```
gdpelican: Make sure we're just loading the inbox once, then later we'll do time-sensitive queries for (what should be) tiny live updates to the inbox.

lineman/app/components/navbar/navbar_user_options.coffee
```diff
@@ -3,5 +3,7 @@ angular.module('loomioApp').directive 'navbarUserOptions', ->
   restrict: 'E'
   templateUrl: 'generated/components/navbar/navbar_user_options.html'
   replace: true
-  controller: 'NavbarController'
+  controller: ($scope, DiscussionFormService) ->
```
gdpelican: This was broken for me (it still is..). I don't know why. I think it's a bower package thingy, but haven't investigated it yet.

lineman/app/models/user_model.coffee
```diff
@@ -11,6 +11,9 @@ angular.module('loomioApp').factory 'UserModel', (BaseModel) ->
     setupViews: ->
       @membershipsView = @recordStore.memberships.belongingTo(userId: @id)
       @notificationsView = @recordStore.notifications.belongingTo(userId: @id)
+      @groupsView = @recordStore.groups.belongingTo(id: { $in: @groupIds() })
```
gdpelican: Woo views.

Also, 'belongingTo' is a bit of a misnomer, since it'll make a view out of whatever params you pass to it, foreign key thingy or not.

lineman/app/services/thread_query_service.coffee
```diff
@@ -0,0 +1,64 @@
...
+                     options['filter'] or 'show_unread',
+                     options['limit'] or 5
+
+    timeframeQuery: (options = {}) ->
+      threadQueryFor createTimeframeView(options['name'], options['timeframe']),
+                     options['filter'] or 'show_all',
+                     options['limit'] or 100
+
+    threadQueryFor = (view, filter) ->
+      view: view
```
gdpelican: Here's our query object... which will likely make its way to being a full fledged model at some point.

It's basically a db view (which stays the same) + a filter (which can change) + helper methods for data access goodness.

I've had a great time with this setup so far.

spec/models/events/new_comment_spec.rb
```diff
@@ -17,6 +17,11 @@
       Events::NewComment.publish!(comment)
     end
 
+    it 'marks the discussion reader as participating' do
```
gdpelican: I basically tested participating enough to ensure that when someone
- creates a discussion
- creates a motion
- creates a comment
- creates a vote

then they are participating in the associated discussion.

spec/models/events/new_vote_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    it 'marks the discussion reader as participating' do
+      Events::NewVote.publish!(vote)
+      expect(DiscussionReader.for(user: vote.author, discussion: discussion).participating).to be_truthy
+    end
+
+    it 'returns an event' do
+      expect(Events::NewVote.publish!(vote)).to eq event
+    end
+  end
+end
```
gdpelican: Woo!

lineman/app/components/inbox_page/inbox_page_controller.coffee
```diff
@@ -0,0 +1,16 @@
+angular.module('loomioApp').controller 'InboxPageController', ($rootScope, Records, CurrentUser, LoadingService, ThreadQueryService) ->
+  $rootScope.$broadcast('currentComponent', 'inboxPage')
```
gdpelican: Hey so this whole controller is 100% duplicated at this point... there's nothing this controller can do at this point that DashboardPageController can't handle by itself.

I'm keeping it around because I think the designs will change to give the inbox some more custom functionality though.

app/controllers/api/discussions_controller.rb
```diff
@@ -55,37 +49,19 @@ def discussion_params
...
-  def grouped(discussions)
-    discussions.map { |g, discussions| discussions.first(Integer(params[:per] || 5)) }.flatten
+  def discussions_for_preview(filter = params[:filter])
+    case filter
+    when 'show_proposals'     then visible_records.not_muted.with_active_motions
+    when 'show_participating' then visible_records.not_muted.participating
+    when 'show_muted'         then visible_records.muted
+    when 'show_unread'        then visible_records.not_muted.unread
+    else                           visible_records.not_muted
+    end.recent.sorted_by_latest_motions
```
robguthrie: This looks really great. We must know what we're doing..

I'm just confused as to why there is still sorted_by_latest_motions on there..

app/extras/queries/visible_discussions.rb
```diff
@@ -49,12 +49,24 @@ def with_active_motions
...
+    self
+  end
+
   def unread
     join_to_discussion_readers
     @relation = @relation.where('dv.last_read_at IS NULL OR (dv.last_read_at < discussions.last_activity_at)')
     self
   end
 
+  def muted
```
robguthrie: looks like you've left off discussions in muted groups.

robguthrie: ie: you need to also select any where m.volume = :mute also

app/models/discussion_reader.rb
```diff
@@ -24,6 +24,10 @@ def set_volume_as_required!
     end
   end
 
+  def participate!
+    update(participating: true) unless self.participating
```
robguthrie: is update the same as update_attribute?

config/locales/client.en.yml
robguthrie: is there a risk that this is introducing another problem rather than solving anything?

db/migrate/20150504075221_populate_participating.rb
robguthrie: ok. We'll need to pull the data migration out of the schema migration. Can you throw this into the DiscussionService or something please?

robguthrie:
Looks great James. Nice work - exciting progress

rdbartlett:
I'm going to do a quick styling pass on this one. Ping me when you're back online James :)

rdbartlett:
Just pushed up my work, not sure if I will get any more done today. I think when I changed all the namespaces to BEM format, I introduced a bug where the proposal pie chart is not rendering on the dashboard. Weird behaviour: one pie chart renders, then if I visit a thread and go back to the dashboard, all the rest of the pies render too. !?

TODO: 

- only display proposal closing at when it is less than 24 hrs
- figure out how to make thread previews truncate properly on mobile
- mobile navbar
- groups page tidy up
- a11y pass

app/controllers/api/discussions_controller.rb
gdpelican: Yep, whoops!

app/extras/queries/visible_discussions.rb
gdpelican: Yep; thanks!

app/models/discussion_reader.rb
gdpelican: Ah, I did want update_attribute here.

Difference is that `update_attribute` isn't subject to validations or before/after hooks (not that we have them here, but traditionally boolean flags like to use update_attribute)

lineman/app/components/base.scss
robguthrie: Headings are necessary to make the page as accessible as possible, but we don't want the same styles between the headings on different pages.

lineman/app/components/dashboard_page/dashboard_page_controller.coffee
robguthrie: can you explain why you keep a record of loaded count?

rdbartlett:
@gdpelican styles are now in a pretty good state. Let's see if those specs pass - hopefully I haven't broken anything by renaming css classes.

I took an axe to the Groups page - that design needs rethinking before we do anything else to it I reckon.

Assuming Travis agrees, I reckon we should merge this before doing anything further.

PR #2142, opened 2015-05-05
Don't error when certain query parameters are passed when logged out

gdpelican:
TDD is dead. Long live TDD.

### DIFF
--- a/spec/controllers/groups_controller_spec.rb
+++ b/spec/controllers/groups_controller_spec.rb
@@ -11,7 +11,17 @@ describe GroupsController do
 
       it "show" do
         get :show, :id => group.key
-        response.should be_success
+        expect(response).to be_success
+      end
+
+      it 'should not error on sifting unread' do
+        get :show, id: group.key, unread: "true"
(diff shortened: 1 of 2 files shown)


spec/controllers/groups_controller_spec.rb
```diff
@@ -11,7 +11,17 @@
 
       it "show" do
         get :show, :id => group.key
-        response.should be_success
+        expect(response).to be_success
+      end
+
+      it 'should not error on sifting unread' do
+        get :show, id: group.key, unread: true
```
robguthrie: @gdpelican did you know there is a real dumb thing about controller tests like this?

passing unread: true like this passes a ruby TrueClass, which is impossible to actually pass in a real request. It's really misleading and causes you to write buggy code.

You need to pass "true" as a string and test that the string is "true" on the other side.. otherwise you're going to have a bug.

gdpelican: Whoa, that is a dumb thing, heh. Thanks!

PR #2148, opened 2015-05-11
Group metrics with ahoy

robguthrie:


app/controllers/api/discussions_controller.rb
```diff
@@ -8,8 +8,8 @@ def inbox_by_date
     respond_with_discussions
   end
 
-  def inbox_by_organization
-    @discussions = grouped inbox_threads.group_by(&:organization_id)
+  def inbox_by_organisation
```
gdpelican: This is coming out anyway. Would you mind reverting so there's no merge conflict?

PR #2149, opened 2015-05-11
Dashboard fetch latest

gdpelican:
Okay @robguthrie here's a first pass for today at loading only the latest records for things.

Notably missing from this implementation is
- Loading activity in a discussion
- Loading notifications

Yes, these are probably the two most important use cases for this. But figured I'd share and see what you think of this tack, which is very minimalist.


config/routes.rb
```diff
@@ -46,8 +46,7 @@
     resources :events, only: :index
 
     resources :discussions, only: [:show, :index, :create, :update, :destroy] do
-      get :discussions_for_dashboard, on: :collection
-      get :discussions_for_inbox, on: :collection
+      get :dashboard, on: :collection
```
gdpelican: These finally collapsed into a single query.

lineman/app/models/discussion_records_interface.coffee
```diff
@@ -3,13 +3,18 @@ angular.module('loomioApp').factory 'DiscussionRecordsInterface', (BaseRecordsIn
     model: DiscussionModel
 
     fetchByGroup: (options = {}) ->
-      @restfulClient.getCollection
+      @fetch
```
gdpelican: I never liked that we were calling @restfulClient directly from the child classes, and turns out this is why: we want everything to go through fetch so behaviour can be as consistent as possible.

app/controllers/api/discussions_controller.rb
```diff
@@ -2,13 +2,8 @@ class API::DiscussionsController < API::RestfulController
...
 
-  def discussions_for_dashboard
-    @discussions = page_collection discussions_for_preview
-    respond_with_discussions
-  end
-
-  def discussions_for_inbox
-    @discussions = discussions_for_preview('show_unread')
+  def dashboard
+    instantiate_collection { |collection| discussions_for_preview(collection) }
```
gdpelican: This might be a little confusing at first glance, but we're passing a filtering block to instantiate collection here.

Then, when we instantiate the collection it looks a little like

- Get visible records
- filter those records if we're filtering things
- timeframe those records if we've got a timeframe
- page those records if we're paging things
- return the filtered, timeframed, and/or paged collection

I think it works rather nicely.

app/controllers/api/discussions_controller.rb
```diff
@@ -39,29 +34,20 @@ def respond_with_discussions
            root: 'discussion_wrappers'
   end
 
-  def discussion_params
```
gdpelican: I'm 99.9% positive this is vestigial.

app/controllers/api/restful_controller.rb
```diff
@@ -96,13 +98,25 @@ def instantiate_resouce
...
+      collection.within(params[:since], params[:until], params[:timeframe_for])
     else
-      visible_records
+      collection
     end
-    self.collection = page_collection(records).to_a
+  end
+
+  def parse_date_parameters
+    API_DATE_PARAMETERS.each { |field| params[field] = DateTime.parse(params[field]) if params[field] }
```
gdpelican: This kinda sucks, but it'll blow up if it gets passed a straight string (because of the timezone format on Moment.toDate(), apparantly), or if we pass in an empty string.

app/controllers/api/discussions_controller.rb
```diff
@@ -39,29 +34,20 @@ def respond_with_discussions
...
-  end
-
   def visible_records
     Queries::VisibleDiscussions.new(user: current_user, groups: current_user.groups)
   end
 
   private
 
-  def discussions_for_preview(filter = params[:filter])
+  def discussions_for_preview(collection, filter = params[:filter])
```
robguthrie: rename to filter_collection maybe?

lineman/app/models/base_records_interface.coffee
```diff
@@ -45,8 +46,20 @@ angular.module('loomioApp').factory 'BaseRecordsInterface', (RestfulClient, $q)
     fetchByKey: (key) ->
       @restfulClient.getMember(key)
 
-    fetch: (params) ->
-      @restfulClient.getCollection(params)
+    fetch: ({params, path, cacheKey}) ->
```
gdpelican: Collapsed this into a single method which I actually found really nice to write fetch methods with (in coffee at least):

```
@fetch
  path: '/my_custom_path'
  params:
    param1: value1
    param2: value2
  cacheKey 'myCustomCacheKey'
```

lineman/app/models/notification_records_interface.coffee
```diff
@@ -2,6 +2,13 @@ angular.module('loomioApp').factory 'NotificationRecordsInterface', (BaseRecords
   class NotificationRecordsInterface extends BaseRecordsInterface
     model: NotificationModel
 
+    fetchMyNotifications: ->
+      @fetch
+        params:
+          from: 0
```
gdpelican: It might do to have a chat about paging with this system, which will be a bit wonky; I don't think we want to be combining timeframe and page too much. (Give me the 25th - 50th record created in the last day? Blerg. Even 'give me the last 20 records created in the last timeframe' doesn't work, because what if someone comes back after an hour and 30 things have been created? Still have a hole.)

robguthrie: nope.. paging does not makes sense in this context. We can only grab most
recent records with this..

robguthrie:
is this ready for merge?

gdpelican:
I think so. I'll merge it up if the tests all pass.

PR #2150, opened 2015-05-12
Thread preview actions

rdbartlett:
screen shot 2015-05-12 at 2 54 35 pm

I designed this in the browser so I figured I'd open a PRQ to share my CSS styles with you @gdpelican -full specs for the feature are [on the roadmap](https://trello.com/c/TBFeAulf/500-dashboard-actions).


gdpelican:
What's a star?

rdbartlett:
Ignore stars for right now. Another job.

AaronThornton00:
Is the obvious text there for accesability?  If so do the screen readers
work with just the title attribute?

rdbartlett:
The text is there so you know what the button does.

AaronThornton00:
I would argue thats what the icon is there for. The title and or tooltip
spell it out for newbies.  Once you do it a couple of times you don't need
the clutter of the text.
Its not hard to imagine the pollution of a list view of five unread items
in a row.
I think icons are way less grating and distracting in index views and
lists.

rdbartlett:
It only shows the action buttons when you hover over the thread though, so you never see more than one at a time. 

I started with icons only, displayed persistently on all threads, with a tooltip on hover:

screen shot 2015-05-12 at 6 52 53 pm

This version shows just the 'mark as read' icon on the unread threads, and just the 'mute' icon on the read threads. I built it and didn't like how the one icon feels like it turns into another one after you click it - I found that quite confusing. The other option would be to have both icons exposed all the time, but then once again it feels cluttered - filling the screen up with icons many people will seldom use.

So I opted instead to only show the actions on hover (like Google Inbox). Then it felt like I could use as much space as I needed to just make the action explicit with a label, rather than hiding that behind another hover. I didn't like hiding the explanation behind a hover, because I can imagine nervous new users never going near enough to the icon to learn what it means.

Incidentally, Inbox uses no labels and no tooltips, just a title. But I feel like their target audience is a bit more technically literate and prepared to experiment than ours. I'm not dead set on that one though.

AaronThornton00:
Nice. Very cool then.  I didn't get the hover thing from the image I saw.

AaronThornton00:
Have you noticed Inbox shows all the icons on the last one you clicked.

rdbartlett:
Yeah it is quite nice - it's displayed whenever you hover, focus, or expand a thread.

gdpelican:
Whoa, this basically already worked anyway.

I patched it up a little, and added starring :star:

robguthrie:
looks great james

rdbartlett:
Woohoo! Amazing!

PR #2162, opened 2015-05-14
Reload dashboard filter on home click

gdpelican:
Don't merge me!

Angular router has a bug, which will be fixed in the next release, around injecting $scope into component controllers. ([linky linky](http://stackoverflow.com/questions/30167898/how-can-we-watch-expressions-inside-the-controller-in-angular-1-4))

I don't think it's a good idea to start tracking the new router from master, so I'd rather just wait until the next angular router release before we merge this (it's a weeny bug anyway).


gdpelican:
Okay, got a working branch of the new angular router working on #2170, so this will be good to go soon after.

rdbartlett:
nice one JK

PR #2165, opened 2015-05-14
Sort discussions by latest activity

gdpelican:
I was getting some really really old threads coming through the dashboard query; turns out we're not sorting discussions at all server side, which means we're not getting the right discussions back.


robguthrie:
oh... this makes a lot of sense. Thanks

PR #2171, opened 2015-05-19
Represent deleted comments in discussions

HSalmon:
screen shot 2015-05-19 at 3 51 55 pm

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -1,6 +1,7 @@
 class Comment < ActiveRecord::Base
   include Twitter::Extractor
   include Translatable
+  include Archivable
 
   has_paper_trail
   acts_as_tree
@@ -78,6 +79,10 @@ class Comment < ActiveRecord::Base
 
   def is_reply?
     parent.present?
+  end
+
+  def archived?
+    archived_at.present?
   end
 
   def liker_names(max: 3)
--- a/app/models/concerns/archivable.rb
+++ b/app/models/concerns/archivable.rb
@@ -1,0 +1,12 @@
+module Archivable
+  extend ActiveSupport::Concern
+
+  included do
+    scope :archived, -> { where('archived_at is not null') }
+    scope :published, -> { where(archived_at: nil) }
+  end
+
+  def archive
+    update_attributes(archived_at: Time.now)
+  end
+end
--- a/app/services/comment_service.rb
+++ b/app/services/comment_service.rb
@@ -41,6 +41,11 @@ class CommentService
     event
   end
 
+  def self.archive(comment:, actor:)
+    actor.ability.authorize!(:archive, comment)
+    comment.archive
+  end
+
   def self.destroy(comment:, actor:)
     actor.ability.authorize!(:destroy, comment)
     # paranoid destroy_all because comment votes seem to be dangling around.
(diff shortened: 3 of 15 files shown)


HSalmon:
@robguthrie, maybe I could get your eyes on this when you've got a mo'?

app/models/comment.rb
```diff
@@ -80,6 +81,10 @@ def is_reply?
     parent.present?
   end
 
+  def archived?
```
robguthrie: this method should be in the concern.

app/models/concerns/archivable.rb
```diff
@@ -0,0 +1,12 @@
+module Archivable
+  extend ActiveSupport::Concern
+
+  included do
+    scope :archived, -> { where('archived_at is not null') }
+    scope :published, -> { where(archived_at: nil) }
```
robguthrie: might need to rethink this.. published might be defined by the host model. in which case it might clash to define published like this. Maybe this should be not_archived, when can then be chained in to published as defined by the host.

app/services/comment_service.rb
```diff
@@ -41,6 +41,11 @@ def self.create(comment:, actor:, mark_as_read: true)
     event
   end
 
+  def self.archive(comment:, actor:)
+    actor.ability.authorize!(:archive, comment)
+    comment.archive
```
robguthrie: i think the archive method should have a !

gdpelican: :exclamation:

app/models/concerns/archivable.rb
robguthrie: Arrr also.. I just realised that we won't be considering archived when pulling a list of comments, because we want to include archived comments in the feed. So insted we should have a body method that returns blank if the comment is archived. We'll need this so that the CommentSerializer does not return the comment body for angular users.

app/models/concerns/archivable.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+  def archive
+    update_attributes(archived_at: Time.now)
```
gdpelican: I think slightly better is `update_attribute` here.

Also, I'm thinking we don't want to update the time if for some reason we end up trying to archive an already-archived model.

```
update_attribute :archived_at, Time.now unless self.archived_at
```

config/routes.rb
```diff
@@ -246,6 +246,7 @@
   resources :comments , only: [:destroy, :edit, :update, :show] do
```
gdpelican: Do we need to keep the `destroy` route around?

app/models/concerns/archivable.rb
robguthrie: I thought that too.. but that is not thread safe. It might not matter too much, but I like to draw attention to all the unthreadsafe things we do, to develop awareness of the issue.

gdpelican: Ah, I didn't realize `update_attribute` isn't threadsafe, thanks! I guess most of the places you'd use it (updating boolean flags), it wouldn't matter, but here it slightly could.

rdbartlett:
Hey @gdpelican we've just had a chat about this and agreed to put it on ice for now. This was intended as an easy quick improvement but the implementation work has revealed hidden complexity. So I'm going to close the PR for now, and we can pull back and design it out properly when we need to.

PR #2172, opened 2015-05-20
Allow custom OAuth2 providers and hide unused OAuth2 providers

soundasleep:
For example, if GOOGLE_KEY isn't set up, then the Google OAuth2 provider shouldn't be displayed in the sign up/login screens.

Adds new environment variables, `CUSTOM_OAUTH2_*`, that allow an arbitrary OAuth2 provider to be supported (in our client's case, provided by the Drupal OAuth2 server module).

Also adds `ENABLE_PERSONA` and `ENABLE_EMAIL_PASSWORD` to allow Persona and email/password registration to be selectively enabled/disabled.

Also adds support for `config/local_env.yml` so environment variables can be loaded at init time.


gdpelican:
Thanks @soundasleep! Give me a few minutes and I'll give it a peek through.

.gitignore
```diff
@@ -28,3 +28,4 @@ bin/
 /lineman/app/css/modules.scss
 .powenv
 /lineman/app/css/modules.scss
+config/local_env.yml
```
gdpelican: We already use `dotenv` for this purpose (loading local env variables for dev/test), so I don't think we need another method.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# EditorConfig helps developers define and maintain consistent
```
gdpelican: Can we take this file out?

app/views/devise/registrations/new.html.haml
```diff
@@ -23,16 +23,26 @@
...
+                      = t :sign_up_with_google
+                  - if has_twitter?
+                    = link_to user_omniauth_authorize_path(:twitter), class: 'btn btn-default btn-block btn-lg login', id: 'twitter-login-btn' do
+                      = t :sign_up_with_twitter
+                  - if has_persona?
+                    = link_to user_omniauth_authorize_path(:browser_id), class: 'btn btn-default btn-block btn-lg login', id: 'persona-login-btn' do
+                      = t :sign_up_with_persona
+                  - if has_custom_oauth2?
+                    = link_to user_omniauth_authorize_path(:custom_oauth2), class: 'btn btn-default btn-block btn-lg login', id: 'custom_oauth2-login-btn' do
+                      = t :sign_up_with_custom_oauth2, :name => Rails.application.secrets.custom_oauth2_name
```
gdpelican: I like to encourage the new `key: value` (over :key => :value) syntax wherever I can; I find it more readable and would like to have it as consistent as possible throughout the codebase.

config/application.rb
```diff
@@ -6,6 +6,14 @@
 
 module Loomio
   class Application < Rails::Application
+    # Load any ENV values from config/local_env.yml, if it exists
+    config.before_configuration do
```
gdpelican: Yeah I'd like to avoid doing this if possible. `dotenv` should handle this for you with values in the `.env` file. Probably a thing we should put in the README, eh?

config/initializers/devise.rb
```diff
@@ -205,10 +207,19 @@
...
   # config.omniauth :github, 'APP_ID', 'APP_SECRET', :scope => 'user,public_repo'
-  config.omniauth 'google_oauth2', Rails.application.secrets.google_key, Rails.application.secrets.google_secret, name: 'google'
-  config.omniauth :facebook, Rails.application.secrets.facebook_key, Rails.application.secrets.facebook_secret
-  config.omniauth :browser_id
-  config.omniauth :twitter, Rails.application.secrets.twitter_key, Rails.application.secrets.twitter_secret
+  config.omniauth 'google_oauth2', Rails.application.secrets.google_key, Rails.application.secrets.google_secret, name: 'google' if has_google?
+  config.omniauth :facebook, Rails.application.secrets.facebook_key, Rails.application.secrets.facebook_secret if has_facebook?
+  config.omniauth :browser_id if has_persona?
+  config.omniauth :twitter, Rails.application.secrets.twitter_key, Rails.application.secrets.twitter_secret if has_twitter?
+  config.omniauth :custom_oauth2, Rails.application.secrets.custom_oauth2_key, Rails.application.secrets.custom_oauth2_secret,
```
gdpelican: Hey so I'm trying to figure out what happens when the day comes that we want to sign in with Github, and Dropbox, and Trello, and anyone else that has an OAuth signin. What are the challenges around adding support for N custom providers, instead of just the one?

config/secrets.yml
```diff
@@ -16,6 +16,16 @@ default: &default
   aws_access_key_id: <%= ENV['AWS_ACCESS_KEY_ID'] %>
   aws_secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
   aws_bucket: <%= ENV['AWS_BUCKET'] %>
+  enable_persona: <%= ENV['ENABLE_PERSONA'] || true %>
```
gdpelican: Don't think we want the secrets file to have opinions about defaults for the ENV values, unless I'm missing something here?

robguthrie: Where else would we put them?

app/strategies/custom_oauth2.rb
```diff
@@ -0,0 +1,52 @@
...
+  # or as a URI parameter). This may not be possible with all
+  # providers.
+
+  # We *need* a UID or else omniauth won't be able to persist name/email fields at registration
+  uid{ raw_info['id'] || raw_info['result']['uid'] || fail("Could not find uid from #{raw_info.to_json}") }
+
+  info do
+    {
+      # Support both normal OAuth2 and Drupal OAuth2
+      :name => raw_info['name'] || raw_info['result']['field_full_name']['und']['item']['value'] || raw_info['result']['name'],
```
gdpelican: Looking at this makes me a little nervous just because I don't know how confident to be that the raw_info will have a 'field_full_name' field, which has a 'und' field, which will have an 'item' field... etc., and this'll blow up if the structure isn't perfect. Are we sure these will come through correctly for every request?

app/views/devise/registrations/_form.html.haml
```diff
@@ -2,9 +2,10 @@
   =f.error_notification
   =f.input :name
   =f.input :email
-  -unless omniauth_authenticated_and_waiting?
-    =f.input :password
-    =f.input :password_confirmation
+  - if has_email_password?
```
gdpelican: Combine for maximum succinctnicity!

`if has_email_password? && !omniauth_authenicated_and_waiting?` :smile:

config/secrets.yml
gdpelican: `<anything> || true`, is always going to be true, so we can't default it to true like that. (And `<anything> || false` will just give the value of anything if it exists, and a falsey value otherwise, so that' not that useful either.)

For defaults, any logic in yml files is a little smelly to me; maybe there's a separate helper / settings file which wraps `Rails.application.secrets`? I just remember when we tried to add defaults a while back, we kinda found we didn't need any.

robguthrie: well put. I agree. I've been wanting to have a CONFIGURATION.md  file which
explains all the ENV settings one can use to configure the app. (and
supporting that.. ensuring that all configuration is done via ENV)

.editorconfig
soundasleep: Sure - I do recommend it for open source projects though, to keep line endings and whitespace consistent :)

app/views/devise/registrations/new.html.haml
soundasleep: Done :)

config/application.rb
soundasleep: Yup - have removed.

config/initializers/devise.rb
soundasleep: Because of OAuth2, you need custom strategies for each and every OAuth2 provider - which Omniauth provides with gems. Therefore I think to support N providers you have to include N extra gems...

app/strategies/custom_oauth2.rb
soundasleep: I've fixed this though it looks a bit messy, I don't know if there's a cleaner way to do it in Ruby?

config/secrets.yml
robguthrie: So should this ENV really be PERSONA_DISABLED so the default is false?

gdpelican: Yeah that makes sense to me.

robguthrie:
Hey @gdpelican can you (pretty please) recap what we need to do to merge this?

gdpelican:
Yep, sure.
- change ENV vars that are defaulting to true to ones that default to false (PERSONA_ENABLED => PERSONA_DISABLED)
- Add a #todo: clean this up comment on the `extract_stuff` methods in `custom_oauth2.rb` I'm not too averse to merging it like that, but I very much don't want someone else coming along and copying it. At some point, if it comes up again, I'll be partial to implementing a simplistic version of Hashie to handle that case.
- merge master, resolve conflicts, and squash
- I'd love a smoke test of this from someone (could be me), maybe @soundasleep could tap up some instructions on how to do that?

gdpelican:
I'm gonna pick this up as a chore in this sprint.

gdpelican:
Closing this in favor of #2305

PR #2173, opened 2015-05-20
Fix usernames with spaces

gdpelican:


gdpelican:
@robguthrie I cannot for the life of me figure out where these users with spaces in their usernames are filtering into the system... I looked at the Omniauth, admin, and good 'ol regular registration flows, and couldn't reproduce.

The before_save hook for generate_username looks solid in this regard at least (although it should likely be before_validation, rather than before_save)... are we sure we're still letting people register usernames with spaces somehow?

PR #2175, opened 2015-05-20
Fix subdomains

gdpelican:
Pulling against the new router stuff.

This is a WIP, be happy to hear thoughts on it.

Stuff I found out:
`$router.navigate(url)` doesn't work currently (WTF new router). Have to figure out what the actual way to say 'hey go to this route now' in NewRouterLand is.
Navigating to a different subdomain pretty much guarantees reloading the app, which I think is a good thing in our case.
Angular currently requires an underlying route in rails in order to function... I don't really know why? So navigating to '/home' won't work unless you have a '/home' route defined in rails. I think this is pretty :-1: and want to find a way around it in the near future.


gdpelican:
This is super merge-b0rked and I want to change the way we're doing this anyway. Closing.

Issue #2120, opened 2015-04-22
Updating the outcome of a proposal after it is closed


niceking:
1. After a proposal had closed and I had published an outcome, a month later I tried to change the outcome of the proposal i made 
2. Page refreshed, got a flash message about the outcome being updated, however I scrolled down and the outcome was now empty
3. Repasted in my text and tried to publish the outcome again, I *think* i got a slightly different success message saying that the outcome has been published. This time text was persisted
4. Did this email everyone who responded? Unclear to me 
5. Have not tried to replicate in case it emails everyone each time

Attached is a screenshot showing that the outcome was updated, and the republished. Also a link to the loomio for those with access to this group 

https://www.loomio.org/d/zL4cBXu1/rails-girls-at-enspiral-space?proposal=8dGKF3iO

![screen shot 2015-04-22 at 12 13 46 pm](https://cloud.githubusercontent.com/assets/4241125/7265147/55ead446-e8e9-11e4-87bf-8a7ada866084.png)

zacharystenger:
I was able to replicate and fix this bug on my development environment.  
In the MotionsController, the update_outcome method makes this call 
```ruby
if MotionService.update_outcome(motion: @motion, params: params, actor: current_user) 
```
which passes params that look something like this (scroll to middle to see motion & outcome objects):
```ruby
{"utf8"=>"✓", "_method"=>"patch", "authenticity_token"=>"XvDLKFeyGSBoYn+FYMeiL2oXinKUndQYJIUmaegZq0pBrng5TjkXXmom+NeN2ftvow0JUhVydxBvffaXL1B/pg==", "motion"=>{"outcome"=>"updated outcome goes here"}, "commit"=>"Publish outcome", "controller"=>"motions", "action"=>"update_outcome", "id"=>"fTcLrP4k"}
```
The problem is the outcome object is inside the motion object so I changed this to pass permitted_params.motion instead like it does in the create_outcome method
```ruby
if MotionService.update_outcome(motion: @motion, params: permitted_params.motion, actor: current_user)
```
which passes params that look something like this:
```ruby
{"outcome"=>"updated outcome goes here"}
```
And it worked (as far as I can tell, but I'm new to Loomio...)  

Another possible fix would be to change the MotionService update_outcome instead. Currently it does this:
```ruby
motion.outcome = params[:outcome]
```
But this sets the outcome to nil because params doesn't immediately have an outcome object, it is within the motion object.  To fix it change that line to this:  
```ruby
[...]

PR #2176, opened 2015-05-20
Fix for issue #2120, update outcome of a proposal

zacharystenger:
Changed ``params: params`` to ``params: permitted_params.motion`` in ``MotionsController#update_outcome``  
I've tested it and it works.


gdpelican:
Great, thanks @zacharystenger 

Would you mind writing a test for this? Should be something in `motions_controller_spec.rb` along the lines of

```
context "updating a motion's outcome" do
  it "updates the motion's outcome" do
    # ensure that the outcome is actually updated when we patch :update_outcome
  end
end
```

Thanks!

zacharystenger:
Sure thing.

gdpelican:
So ya know, the last step will be to squash up the commits on your branch into a single commit; there's a couple of methods for doing that here:

https://github.com/loomio/loomio/wiki/How-to-squash

Thanks again for your contribution :cool::balloon:

zacharystenger:
Awesome, thanks for the tip.

zacharystenger:
I wrote a test that fails with this bug and passes with my fix. But, I feel like I'm only testing that ``MotionService#update_outcome`` is called correctly and am not checking that the outcome is actually updated.  Any feedback would be great, I'm pretty new to Ruby on Rails and RSpec, thanks!  
In ``motions_controller_spec.rb`` I added this with the other let commands
``  let(:permitted_params) { stub_model(PermittedParams, motion: {outcome: 'new outcome'}) }``  

And added this test:
```
context "updating a motion's outcome" do
  it "updates the motion's outcome" do
    MotionService.should_receive(:update_outcome).with(:motion => motion, :params => permitted_params.motion, :actor => user)
    post :update_outcome, :id => motion.key, :motion => permitted_params.motion
  end
end
```

gdpelican:
Hm, I don't see these changes in the current PR; could you push them up?

As you note, the other tests in that area are just testing that the service is called correctly. So perhaps we want another test in `motion_service_spec.rb` to match the `describe '.create_outcome` test?

zacharystenger:
I created the service test and pushed both.  
Reading up on squashing now. My apologies for the messy PR, future ones will be cleaner.  Thanks for your patience and help!

PR #2178, opened 2015-05-20
send more params to intercom

rdbartlett:
I feel like this is fine to merge but I haven't tested it


rdbartlett:
@robguthrie when you get a chance :)

PR #2184, opened 2015-05-21
locales update

mixmix:
here you go!


mixmix:
@robguthrie  :whale:

robguthrie:
thanks a billion mixy

PR #2186, opened 2015-05-21
Generate username method cleanup

gdpelican:
Clean up the generate_username method on user, and run it before validation

### DIFF
--- a/app/extras/username_generator.rb
+++ b/app/extras/username_generator.rb
@@ -1,0 +1,26 @@
+UsernameGenerator = Struct.new(:user) do
+
+  def generate
+    begin
+      i ||= 0
+      username = "#{base_username}#{i if i > 0}"
+      i += 1
+    end while conflicts.include? username
+    username
+  end
+
+  private
+
+  def base_username
+    (user.name || user.email).split('@').first      # all characters before '@'
+                             .parameterize          # convert ascii chars
+                             .downcase              # all lower case
+                             .gsub(/[^a-z0-9]/, '') # alphanumeric
+                             .slice(0,18)           # first 18 characters
+  end
+
+  def conflicts
+    UsernameConflictsQuery.conflicts_for(base_username).pluck(:username)
+  end
+
+end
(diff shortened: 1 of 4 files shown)


app/extras/username_generator.rb
```diff
@@ -0,0 +1,27 @@
+UsernameGenerator = Struct.new(:user) do
```
robguthrie: What's the reason for not using class here?

app/extras/username_generator.rb
```diff
@@ -0,0 +1,27 @@
...
+    @base_username ||=
+      (user.name || user.email).split('@').first      # all characters before '@'
+                               .parameterize          # convert ascii chars
+                               .downcase              # all lower case
+                               .gsub(/[^a-z0-9]/, '') # alphanumeric
+                               .slice(0,18)           # first 18 characters
+  end
+
+  def conflicts
+    @conflicts ||= UsernameConflictsQuery.conflicts_for(base_username).pluck(:username)
```
robguthrie: This looks really great.. however in both cases I don't understand why you memoize. Seems like a source of problems without benefits

robguthrie:
Really lovely work. Merge when you please but I'd love to know why you prefer struct and why you memoize.

app/extras/username_generator.rb
gdpelican: I'm a Struct fan in instances where the initialize method is just boilerplate.

So having to write

```
attr_accessor :some_thing, :some_other_thing, :some_third_thing
def initialize(some_thing, some_other_thing, some_third_thing)
  @some_thing = some_thing
  @some_other_thing = some_other_thing
  @some_third_thing = some_third_thing
end
```
 is kinda verbose and unnecessary and I try to avoid it. (here it's only one thing, but I still like Structs for it)

Maybe it'd be useful to write a thing which made this work:

```
class UsernameGenerator
  initialize :some_thing, :some_other_thing, :some_third_thing
end
```

app/extras/username_generator.rb
gdpelican: Er, yeah, I don't need to memoize these, thanks. Leftover from an earlier iteration where I was calling 'conflicts' and 'base_username' multiple times.

robguthrie:
yay. go with the structs. this seems great James.

PR #2193, opened 2015-05-25
Scroll to first h1 on page load

gdpelican:


lineman/app/app.coffee
```diff
@@ -21,12 +21,12 @@ angular.module('loomioApp', ['ngNewRouter',
...
 
-angular.module('loomioApp').controller 'AppController', ($scope, $filter, $rootScope, $router, KeyEventService) ->
+angular.module('loomioApp').controller 'AppController', ($scope, $filter, $rootScope, $router, KeyEventService, ScrollService) ->
   $scope.currentComponent = 'nothing yet'
 
-  $scope.$on 'currentComponent', ->
+  $scope.$on 'currentComponent', (event, options = {}) ->
     $scope.pageError = null
-    angular.element(document.querySelector('.main-container'))
+    ScrollService.scrollTo(options['scrollTo'] or 'h1:first')
```
gdpelican: By default, we'll scroll to the first 'h1' on the page. If we need to it be something different, we can broadcast a scrollTo option on the page controller, like this:

```
$rootScope.$broadcast 'currentComponent', { page: 'thisPage', scrollTo: '#myScrollAnchor' }
```

robguthrie:
looks good.

PR #2194, opened 2015-05-25
Add Model Locator class for loading via params

gdpelican:
Hey I did this over the weekend because it was bugging me.

The API controllers are quite well tested, so it's a pretty low risk refactor imho.


gdpelican:
screenshot from 2015-05-25 12 13 11
Kawaii! :raised_hands:

robguthrie:
nice one!

PR #2197, opened 2015-05-26
Serverside event loading for threads

gdpelican:
WIP; don't merge me.

### DIFF
--- a/app/models/event_stream.rb
+++ b/app/models/event_stream.rb
@@ -1,0 +1,38 @@
+class EventStream
+
+  PAGESIZE = 30
+  ROLLBACK = 2
+
+  def initialize(discussion_reader:, focused: nil)
+    @discussion_reader = discussion_reader
+    @discussion = discussion_reader.discussion
+    @focused = focused
+  end
+
+  def initial_events
+    Event.where(discussion: @discussion)
+         .where('sequence_id > ?', load_from_event)
+         .order(sequence_id: :asc)
+         .limit(PAGESIZE)
+  end
+
+  def focused_sequence_id
+    @focused ||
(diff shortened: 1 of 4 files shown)


gdpelican:
Hey @robguthrie I threw this together as a solution for having to wait for events to load on the thread page.

Main thing missing at the moment is getting the focused sequence id (ie 15 in /d/slug/:id#15) server side, since the hash part of the url doesn't get passed. Might look at doing a url like /d/slug/:id/15 instead.

app/models/event_stream.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def initial_events
+    Event.where(discussion: @discussion)
+         .where('sequence_id > ?', load_from_event)
+         .order(sequence_id: :asc)
+         .limit(PAGESIZE)
+  end
+
+  def focused_sequence_id
```
gdpelican: I may just take this out, and have the client worry about focusing the right element.
The server will still need to know what sequence_id is being targeted in the anchor, though, so it can load the right events.

robguthrie:
hey I'm not sure that this will solve the problem james. Are you free for a chat?

robguthrie:
What's the reason for moving this logic to the server side?

app/models/event_stream.rb
robguthrie: can you pleeease come online?

robguthrie: Thing is, we want to preload the events _before_ the user clicks on a discussion. Because we already have the discussion locally so it loads instantly.. it would be best if we already had the events too.

So I imagine that we fetch discussions as per usual, then we grab relevant events for the top few discussions in the current view of the dashboard.

PR #2198, opened 2015-05-27
Ensure page load thread position

gdpelican:
- Ensure page scroll happens after events & proposals are loaded
- Take out support for location hashes for now
- Ensure record is in clientside db before scrolling

### DIFF
--- a/lineman/app/components/thread_page/activity_card/activity_card.coffee
+++ b/lineman/app/components/thread_page/activity_card/activity_card.coffee
@@ -47,7 +47,6 @@ angular.module('loomioApp').directive 'activityCard', ->
     $scope.threadItemVisible = (item) ->
       addSequenceId(item.sequenceId)
       $scope.discussion.markAsRead(item.sequenceId)
-      # $location.hash(_.min(visibleSequenceIds))
       $scope.loadEventsForwards($scope.lastLoadedSequenceId) if $scope.loadMoreAfterReading(item)
 
     $scope.loadEvents = ({from, per}) ->
--- a/lineman/app/components/thread_page/activity_card/activity_card.coffee
+++ b/lineman/app/components/thread_page/activity_card/activity_card.coffee
@@ -8,29 +8,23 @@ angular.module('loomioApp').directive 'activityCard', ->
     $scope.pageSize = 30
     $scope.firstLoadedSequenceId = 0
     $scope.lastLoadedSequenceId = 0
-    $scope.newActivitySequenceId = $scope.discussion.reader().lastReadSequenceId + 1
+    $scope.lastReadSequenceId = $scope.discussion.reader().lastReadSequenceId
     visibleSequenceIds = []
-    rollback = 2
 
     $scope.init = ->
       $scope.discussion.markAsRead(0)
 
-      target = _.parseInt($location.hash())
-      if target >= $scope.discussion.firstSequenceId and target < $scope.discussion.lastSequenceId 
-        # valid sequence id is specified in url
-        $scope.initialLoaded  = _.max [target - rollback, 0]
-        $scope.initialFocused = target
-      else if $scope.discussion.isUnread()
-        # discussion is unread
-        $scope.initialLoaded  = _.max [$scope.discussion.reader().lastReadSequenceId - rollback, 0]
-        $scope.initialFocused = $scope.initialLoaded + rollback
-      else
-        # discussion is read
-        $scope.initialLoaded  = _.max [$scope.discussion.lastSequenceId - $scope.pageSize + 1, 0]
-        $scope.initialFocused = _.max [$scope.discussion.lastSequenceId - rollback, 0]
+      $scope.loadEventsForwards($scope.initialLoadSequenceId()).then ->
+        $rootScope.$broadcast 'threadPageEventsLoaded'
 
-      $scope.loadEventsForwards($scope.initialLoaded).then ->
-        $rootScope.$broadcast 'threadPageEventsLoaded', $scope.initialFocused
+    $scope.initialLoadSequenceId = ->
+      if $scope.discussion.isUnread()
+        $scope.discussion.reader().lastReadSequenceId - 1
+      else
+        $scope.discussion.lastSequenceId - $scope.pageSize + 1
+
+    $scope.hasNewActivity = ->
+      $scope.discussion.isUnread()
 
     $scope.beforeCount = ->
       $scope.firstLoadedSequenceId - $scope.discussion.firstSequenceId
@@ -53,7 +47,7 @@ angular.module('loomioApp').directive 'activityCard', ->
     $scope.threadItemVisible = (item) ->
       addSequenceId(item.sequenceId)
       $scope.discussion.markAsRead(item.sequenceId)
-      $location.hash(_.min(visibleSequenceIds))
+      # $location.hash(_.min(visibleSequenceIds))
       $scope.loadEventsForwards($scope.lastLoadedSequenceId) if $scope.loadMoreAfterReading(item)
 
     $scope.loadEvents = ({from, per}) ->
--- a/lineman/app/components/thread_page/activity_card/activity_card.haml
+++ b/lineman/app/components/thread_page/activity_card/activity_card.haml
@@ -6,7 +6,7 @@
   %loading.activity-card__loading.page-loading{ng-show: 'loadEventsBackwardsExecuting'}
   %ul.activity-card__activity-list
     %li.activity-card__activity-list-item{ng_repeat: 'event in discussion.events() track by event.id', in-view: '($inview&&threadItemVisible(event)) || (!$inview&&threadItemHidden(event))', in-view-options: '{debounce: 100}'}
-      .activity-card__new-activity{ng-if: 'event.sequenceId == newActivitySequenceId', translate: 'activity_card.new_activity'}
       .activity-card__last-item{ng-if: '$last'}
       .activity-card-content{id: 'sequence-{{event.sequenceId}}', ng-include: "'generated/components/thread_page/activity_card/'+ event.kind + '.html'"}
+      .activity-card__last-read-activity{ng-if: 'event.sequenceId == lastReadSequenceId', ng-class: '{"activity-card__new-activity": hasNewActivity()}', translate: 'activity_card.new_activity'}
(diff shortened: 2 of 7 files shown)


lineman/app/components/thread_page/activity_card/activity_card.coffee
```diff
@@ -10,24 +10,23 @@ angular.module('loomioApp').directive 'activityCard', ->
...
+      # if target >= $scope.discussion.firstSequenceId and target < $scope.discussion.lastSequenceId 
+      #   # valid sequence id is specified in url
+      #   $scope.initialLoaded  = _.max [target - rollback, 0]
+      #   $scope.initialFocused = target
+      if $scope.discussion.isUnread()
         # discussion is unread
-        $scope.initialLoaded  = _.max [$scope.discussion.reader().lastReadSequenceId - rollback, 0]
-        $scope.initialFocused = $scope.initialLoaded + rollback
+        $scope.initialLoaded  = _.max [$scope.discussion.reader().lastReadSequenceId - 1, 0]
+        $scope.initialFocused = $scope.discussion.reader().lastReadSequenceId + 1
```
robguthrie: There is a chance that lastReadSequenceId + 1 is not rendered in the thread. Can you focus on the new activity marker instead?

lineman/app/components/thread_page/activity_card/activity_card.coffee
```diff
@@ -10,24 +10,23 @@ angular.module('loomioApp').directive 'activityCard', ->
...
       else
         # discussion is read
         $scope.initialLoaded  = _.max [$scope.discussion.lastSequenceId - $scope.pageSize + 1, 0]
-        $scope.initialFocused = _.max [$scope.discussion.lastSequenceId - rollback, 0]
+        $scope.initialFocused = $scope.discussion.lastSequenceId
```
robguthrie: Again.. last sequence id might not be rendered.. we can just use lastReadSequenceId here right?

lineman/app/components/thread_page/thread_page_controller.coffee
```diff
@@ -1,24 +1,43 @@
...
-    @setFocusMode()
-    ScrollService.scrollTo('.thread-context') if @focusMode == 'context'
-  , (error) ->
+  @performScroll = ->
+    focusedElement = @focusedElement()
+    if focusedElement && !@scrolledAlready
+      ScrollService.scrollTo focusedElement
+      @scrolledAlready = true
+
+  @focusedElement = ->
```
robguthrie: name suggestion: elementToFocus

lineman/app/components/thread_page/thread_page_controller.coffee
```diff
@@ -1,24 +1,43 @@
...
+    if $location.hash().match(/^proposal-\d+$/) and Records.proposals.find(@focusedProposalId)
+      "#proposal-#{@focusedProposalId}"
+    else if @discussion.lastSequenceId == 0 or @discussion.reader().lastReadSequenceId == -1
+      ".thread-context"
+    else if Records.events.findByDiscussionAndSequenceId(@discussion, @focusedSequenceId)
```
robguthrie: Yea.. again can we focus on the new activity indicator? I think that would be real nice.

Is it easy to explain/tell me why you check for a record here?

robguthrie: Oh now I see. focusedElement can return false in which case you don't scroll. Maybe add a final `else false` and a little comment? (if i understand)

lineman/app/components/thread_page/thread_page_controller.coffee
```diff
@@ -1,24 +1,43 @@
...
+      MessageChannelService.subscribeTo "/discussion-#{@discussion.key}"
+      @performScroll()
+
+  @init Records.discussions.find $routeParams.key
+  Records.discussions.findOrFetchByKey($routeParams.key).then @init, (error) ->
     $rootScope.$broadcast('pageError', error)
 
-  @onMessageReceived = (event) ->
-    console.log 'discussion update received', event
+  $scope.$on 'threadPageEventsLoaded',    (event, sequenceId) =>
```
robguthrie: Worst case (quick user, slow network) you might get this signal for another thread (if you navigate to another thread before the previous threads events have loaded.) you might want to confirm discussion key before proceeding (ie: return unless discussionKey = @discussion.key)

lineman/app/models/event_records_interface.coffee
```diff
@@ -9,6 +9,12 @@ angular.module('loomioApp').factory 'EventRecordsInterface', (BaseRecordsInterfa
           from: options['from']
           per: options['per']
 
+    findByDiscussionAndSequenceId: (discussion, sequenceId) ->
+      @collection.chain()
+                 .find(discussionId: discussion.id)
+                 .find(sequenceId: sequenceId)
+                 .data()[0] # so, turns out finding by multiple parameters doesn't actually work..
```
robguthrie: yea nah, that's right. gotta use a result set ..

robguthrie:
This is looking really great james. awesome work.

lineman/app/components/thread_page/activity_card/activity_card.coffee
gdpelican: The  newActivitySequenceId actually has the same problem at the moment:

```
  $scope.newActivitySequenceId = $scope.discussion.reader().lastReadSequenceId + 1
```

```
  .activity-card__new-activity{ng-if: 'event.sequenceId == newActivitySequenceId'}
```

lineman/app/components/thread_page/activity_card/activity_card.coffee
gdpelican: The new activity marker has the same problem at the moment, actually.

```
  $scope.newActivitySequenceId = $scope.discussion.reader().lastReadSequenceId + 1
```
```
  .activity-card__new-activity{ng-if: 'event.sequenceId == newActivitySequenceId'}
```

lineman/app/components/thread_page/thread_page_controller.coffee
gdpelican: Cool, changed that.

lineman/app/components/thread_page/thread_page_controller.coffee
gdpelican: Yep, that's why. I'll make it more explicit.

lineman/app/components/thread_page/activity_card/activity_card.coffee
gdpelican: Mmmight have to write a little algorithm that's like 'get me the smallest sequence id in the record store greater than the lastReadSequenceId'?

robguthrie: either that or you pass a selector (rahter than a sequence id) to
initialFocused and place dom elements into the thread when you pass
lastReadSequenceId

lineman/app/components/thread_page/activity_card/activity_card.haml
```diff
@@ -6,7 +6,7 @@
   %loading.activity-card__loading.page-loading{ng-show: 'loadEventsBackwardsExecuting'}
   %ul.activity-card__activity-list
     %li.activity-card__activity-list-item{ng_repeat: 'event in discussion.events() track by event.id', in-view: '($inview&&threadItemVisible(event)) || (!$inview&&threadItemHidden(event))', in-view-options: '{debounce: 100}'}
-      .activity-card__new-activity{ng-if: 'event.sequenceId == newActivitySequenceId', translate: 'activity_card.new_activity'}
       .activity-card__last-item{ng-if: '$last'}
       .activity-card-content{id: 'sequence-{{event.sequenceId}}', ng-include: "'generated/components/thread_page/activity_card/'+ event.kind + '.html'"}
+      .activity-card__new-activity{ng-if: 'event.sequenceId == lastReadSequenceId', translate: 'activity_card.new_activity'}
```
gdpelican: If the new activity indicator is rendered after the record, then we can pop straight to it without worrying about whether the sequenceId after it is rendered or not.

lineman/app/components/thread_page/thread_page_controller.coffee
```diff
@@ -1,24 +1,48 @@
...
+      else
+        "#sequence-#{@sequenceIdToFocus}"
+    else
+      false # our record isn't there yet
+
+  @init = (discussion) =>
+    if discussion and !@discussion?
+      @discussion = discussion
+      @group = @discussion.group()
+      @sequenceIdToFocus = @discussion.reader().lastReadSequenceId # or location hash when we put it back in.
```
gdpelican: I can't help but notice that we're scrolling to discussionReader.lastReadSequenceId... pretty much no matter what. I could be a fan of saying 'just scroll to the bottom of the last read thing, always.'

robguthrie: Yea I totally agree... It's a nice way to sidestep the sequence Id problems too..because the thread ng-repeat can be responsible for placing the element we're going to scroll+focus on, and we just reference that rahter than an ID.

The only thing I'd like to add is that we focus on the new activity indicator.. or if there is no new activity then focus on the last item. (i suggest we add a 'last item' class). I reckon it will be important for a11y to focus on the new activity indicator.. that way the screenreader can say stuff like "5 new comments since you last read this discussion" and they can continue on happy and knowing what's going on.

robguthrie: I think it might be nice to add a last item element.. that does the same kinda thing as new activity but is not visible.

gdpelican: I split the difference here and added a 'last-activity' item, which also has a 'new-activity-item' class if the discussion is unread.

Now we can just always scroll to that.

lineman/spec/directives/activity_card_spec.coffee
```diff
@@ -12,32 +12,3 @@ describe 'Activity Card Component', ->
     prepareDirective @, 'activity_card', (parent) =>
       parent.discussion = @factory.create 'discussions', title: 'hi mom'
     expect(@$scope.discussion.title).toBe('hi mom')
-
-  it 'scrolls to unread when the discussion is unread', ->
```
gdpelican: We refactored out all of these tests, heh.

PR #2199, opened 2015-05-28
Add placeholders to group page

gdpelican:
Added placeholders to the group page, with some specs around it.

### DIFF
--- a/app/models/discussion.rb
+++ b/app/models/discussion.rb
@@ -193,6 +193,10 @@ class Discussion < ActiveRecord::Base
     end
   end
 
+  def example_discussion?
+    self.author_id == User.loomio_helper_bot.id
+  end
+
   private
   def set_last_activity_at_to_created_at
     update_attribute(:last_activity_at, created_at)
--- a/app/models/discussion.rb
+++ b/app/models/discussion.rb
@@ -193,10 +193,6 @@ class Discussion < ActiveRecord::Base
     end
   end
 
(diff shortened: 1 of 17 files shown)


app/models/discussion.rb
```diff
@@ -193,6 +193,10 @@ def inherit_group_privacy!
     end
   end
 
+  def example_discussion?
+    self.author_id == User.loomio_helper_bot.id
```
gdpelican: This pretty much sucks, and I think we probably want to refactor and add an 'example' column to Discussion so we don't have to bother the poor helper bot 1000s of times for each discussion we load.

lineman/app/components/group_page/group_page.haml
```diff
@@ -27,6 +27,7 @@
         .col-sm-8
           %section.group-page__description{aria-labelledby: 'description-card-title'}
             %h2#description-card-title.lmo-card-heading{ translate: 'group_page.description'}
+            .group-page__description-placeholder.group-page-placeholder{translate: 'group_page.description_placeholder', ng-if: 'groupPage.showDescriptionPlaceholder()' }
```
rdbartlett: no biggie but I was thinking the utility class for this should be .lmo-hint or something? I expect it will be repeated throughout the app

gdpelican: Yeah makes sense. Can I kick that to you for a future styling pass?

PR #2200, opened 2015-05-28
trying to get travis to run angular specs/e2es

robguthrie:


.travis.yml
```diff
@@ -10,16 +20,15 @@ branches:
...
+  # start rails s for cuckes to hit 
+
+#script: "bundle exec rake travis"
+script: cd lineman && 
+
 addons:
   postgresql: "9.4"
-  code_climate:
-    repo_token:
-      secure: "RQcqTqhx9hhcWO97ib8s8I8m/ufPbBuKRfEEh4RHNIsBEMpqekaPNHtK+T2ym2pfbCpD+L+0bZ+nMlSCrYz0gklfNnYJlPH4/g1f0SX2OEdq2eUCi9xlRObi9ayXsz7Ujtk14mncNGx+e02TnAtEjtFBT8uyvxn8Ql4djvOTsCg="
```
gdpelican: I still use codeclimate occasionally, and it shows up as a badge on our README, so I don't think we want to get rid of it just yet.

robguthrie: ok cool. however is it fine to leave the token in the source code? Is there
another place we can put it maybe?

gdpelican: https://coderwall.com/p/5mtq6q/encrypt-your-code-climate-repo-token-for-public-repositories-on-travis-ci

Looks like that's the :ticket:

robguthrie:
This is almost working now!!!

PR #2201, opened 2015-05-28
Remove discussions#index

gdpelican:
Considering this has in prod for a while now and we haven't seen it come through errbit, I'm pretty sure it's safe to take out.


gdpelican:
Attack of the Merge:monkey:!

Issue #2039, opened 2015-03-25
default group logo broken


rdbartlett:
the default group logo (on the group page) is pointing to a resource that doesn't exist:
https://www.loomio.org/images/default-logo-medium.png

baschtl:
@rdbartlett Is this still the case? When I look at https://www.loomio.org/g/CY61pwAE/llengues-d-arago-catala the default logo shows up.

rdbartlett:
Thanks for checking in @baschtl - yep it is still broken for new groups.

baschtl:
Ah, I was able to verify this locally. I will look into it.

PR #2202, opened 2015-05-28
Fixes default group logo and cover photo broken

baschtl:
This PR addresses #2039.

The changes I reverted were introduced by @robguthrie on 26th February. We should double check with his help why this was introduced.

Somebody: Please review.

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -178,11 +178,11 @@ class Group < ActiveRecord::Base
 
   has_attached_file    :cover_photo,
                        styles: { desktop: "970x200#", card: "460x94#"},
-                       default_url: '/images/default-cover-photo.png'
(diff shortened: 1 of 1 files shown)


gdpelican:
Thanks for this, @baschtl. This looks good to me, but I want to wait until Rob has a peek to merge to ensure I'm not missing something.

robguthrie:
If this fixes it for loomio beta that's great. Thanks for this @baschtl. I think longer term I'd like to remove the defaults from the models entirely and have the views pull in a default image if there is no image defined. 

Defaults are really difficult to get right when you have two UI's (our default image in Loomio 1.0 will probably be different to this one) and it seems to make more sense for the UI/view code to manage what image to use in case of no user defined image.

Thanks again for fixing this.

PR #2208, opened 2015-05-29
Don't disappear new activity marker after read

gdpelican:


gdpelican:
I should've caught this in testing. I am a shame :panda_face:.

PR #2209, opened 2015-06-01
add subdomain support to LmoUrlService w/ specs

gdpelican:
Don't merge me just yet

This should add support for swapping around subdomains when navigating... will be interesting to test it out a little more in practice to see what further needs to be done.

### DIFF
--- a/lineman/app/models/group_model.coffee
+++ b/lineman/app/models/group_model.coffee
@@ -16,6 +16,12 @@ angular.module('loomioApp').factory 'GroupModel', (BaseModel) ->
     organisationIds: ->
       _.pluck(@subgroups(), 'id').concat(@id)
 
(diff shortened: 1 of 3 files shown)


gdpelican:
This is squashed and should be good to go, @robguthrie

lineman/app/models/group_model.coffee
```diff
@@ -22,6 +22,12 @@ angular.module('loomioApp').factory 'GroupModel', (BaseModel) ->
       else
         []
 
+    pathSubdomain: ->
```
robguthrie: pathSubdomain? OrganisationSubdomain?

gdpelican: Noted and changed.

PR #2213, opened 2015-06-03
try to fix incorrect voters count

robguthrie:
not ready


gdpelican:
It'd be super cool to have a failing spec for this issue, to isolate where exactly the mismatch is.

robguthrie:
I totally agree.. we have failing case in the census2018 group, need to find out the detail and replicate when I get time.

mhjb:
Any update on this?

PR #2215, opened 2015-06-03
Spring cleaning

gdpelican:
Some quick refactoring I did on the plane wee! (Don't merge me yet)

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -15,31 +15,23 @@ class Users::SessionsController < Devise::SessionsController
 
   # and this
   def create_parse_user_if_needed
-    return unless need_to_create_parse_user?
-    Group.find(ENV['PARSE_GROUP_ID']).add_member! parsed_user
-    sign_in :user, parsed_user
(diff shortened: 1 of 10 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -15,23 +15,31 @@ class Users::SessionsController < Devise::SessionsController
 
   # and this
   def create_parse_user_if_needed
-    if ENV['PARSE_ID'] and request.subdomain == ENV['PARSE_SUBDOMAIN']
```
robguthrie: This is one thing I don't think you should ever touch. We'll probably remove it soon.. and it was only added for a single group.. and there are no tests.

robguthrie: that aside.. looks like a nice refactor

gdpelican:
I want to hold off on this for now until I get a better chance to test out the discussion cache refactor

gdpelican:
Hey @robguthrie how do you feel about merging this after the tests go green?

robguthrie:
good!

PR #2221, opened 2015-06-08
locales update

mixmix:
shazam


mixmix:
here you go @robguthrie

robguthrie:
Thanks a billion mixmix

robguthrie:
Mix this is deployed now.

PR #2223, opened 2015-06-09
Angular proposal outcome

HSalmon:


app/controllers/api/motions_controller.rb
```diff
@@ -11,6 +11,22 @@ def close
...
+  def create_outcome
+    load_and_authorize(:motion, :create_outcome)
+    MotionService.create_outcome(motion: @motion,
+                                 params: permitted_params.motion,
+                                 actor:  current_user)
+    respond_with_resource
+  end
+
+  def update_outcome
+    load_and_authorize(:motion, :update_outcome)
```
robguthrie: Interesting. The MotionService already authorises (and raises if action not permitted). Might be cool to have a chat about where we want authorisation to happen. (I've been thinking that keeping it in the service relieves controllers of the task and prevents it being forgotten).

lineman/spec/models/discussion_model_spec.coffee
```diff
@@ -61,3 +62,30 @@ describe 'DiscussionModel', ->
...
+        permittedParams:
+          discussion: ['title', 'author_id']
+
+    it 'copies all of the attributes of the object being cloned', ->
+      expect(discussion.clone().title).toBe(discussion.title)
+
+    it 'copies all of the methods of the object being closed', ->
+      expect(discussion.clone().authorName()).toBe(discussion.authorName())
+
+    it 'does not create a new record', ->
```
robguthrie: 'does not add record to recordstore' seems more correct

lineman/spec/models/discussion_model_spec.coffee
```diff
@@ -61,3 +62,30 @@ describe 'DiscussionModel', ->
   describe 'reader', ->
     it "returns the discussion reader associated with this discussion", ->
       expect(discussion.reader()).toBe(discussionReader)
+
+  describe 'clone()', ->
```
robguthrie: this effing rules

app/controllers/api/motions_controller.rb
gdpelican: Ah right. I suppose we'll go back to just loading it there for now, although it's a little confusing to separate loading and authorizing like that.. I totally agree about needing to auth in the service though.

app/serializers/permitted_params_serializer.rb
```diff
@@ -5,7 +5,7 @@ def object
     PermittedParams.new
   end
 
-  %w[user vote motion membership_request membership
+  %w[user vote motion proposal membership_request membership
```
gdpelican: We tripped on this thing again, heh. I never would have imagined how many times it would crop up and be a  pain :sheep:

lineman/app/models/base_model.coffee
```diff
@@ -17,6 +17,13 @@ angular.module('loomioApp').factory 'BaseModel', ->
     initialize: (data) ->
       @updateFromJSON(data)
 
+    clone: ->
+      attrs = @baseSerialize()[@constructor.singular] or {}
+      clone = @recordsInterface.initialize attrs
+      clone.id = @id
```
gdpelican: `_.merge clone, { id: @id, key: @key }` maybe?

lineman/app/models/group_model.coffee
```diff
@@ -87,4 +90,5 @@ angular.module('loomioApp').factory 'GroupModel', (BaseModel) ->
         '/img/default-cover-photo.png'
 
     archive: ->
+      # is this broken (group null right?)
```
gdpelican: That looks very broken :smiley:

lineman/app/models/user_model.coffee
```diff
@@ -57,6 +57,17 @@ angular.module('loomioApp').factory 'UserModel', (BaseModel) ->
       proposal.isActive() and
       (@isAdminOf(proposal.group()) or @isAuthorOf(proposal))
 
+    canCreateOutcomeFor: (proposal) ->
```
gdpelican: This seems like it can be unduplicated some... and it seems like proposal.isClosed() should be on both, no?

```
proposal.hasOutcome() and @canSetOutcomeFor(proposal)
```
and
```
!proposal.hasOutcome() and @canSetOutcomeFor(proposal)
```

lineman/spec/models/discussion_model_spec.coffee
gdpelican: I have to note that a couple of these are failing at the moment, because we aren't currently simulating the bit where we post to the server and the updating values get imported back into the record store.

lineman/spec-e2e/proposal_spec.coffee
```diff
@@ -1,87 +1,62 @@
 describe 'Proposals', ->
 
   DiscussionPage = require './helpers/discussion_page.coffee'
+  proposalsHelper = require './helpers/proposals_helper.coffee'
   page = new DiscussionPage
```
robguthrie: would be good if DiscussionPage had a 'new' at the beginning of it.. so we end up with discussionPage = require ... and no need for new DiscissionPage here

lineman/app/models/base_model.coffee
```diff
@@ -17,6 +17,13 @@ angular.module('loomioApp').factory 'BaseModel', ->
     initialize: (data) ->
       @updateFromJSON(data)
 
+    clone: ->
+      attrs = @baseSerialize()[@constructor.singular] or {}
+      clone = @recordsInterface.initialize attrs
```
robguthrie: I don't trust this. BaseRecordsInterface:initialize inserts into the colleciton. I think you want to be calling new Model.. so maybe something like new @constructor(attrs)

app/controllers/api/motions_controller.rb
gdpelican: How about making the service smart enough to accept an id, so the controller doesn't have to know about a Motion at all?

robguthrie: oh yea.. that sounds alright eh. motion_id: rather than motion:

gdpelican:
Alright, I just gave this a quick cleanup and a smoke test, and am satisfied with it. Let's dooo it!

robguthrie:
amazing!

PR #2225, opened 2015-06-10
Angular start button

gdpelican:
:rocket:

### DIFF
--- a/lineman/app/components/mixins.scss
+++ b/lineman/app/components/mixins.scss
@@ -114,7 +114,7 @@ $navbar-height: 45px;
 #turn-off-angular {
   z-index: 100;
   position: fixed;
-  right: 10px;
+  left: 10px;
   bottom: 10px;
   width: 100px;
   background-color: pink
--- a/lineman/app/components/start_menu/start_menu.coffee
+++ b/lineman/app/components/start_menu/start_menu.coffee
@@ -1,0 +1,12 @@
+angular.module('loomioApp').directive 'startMenu', ->
+  scope: {}
+  restrict: 'E'
+  templateUrl: 'generated/components/start_menu/start_menu.html'
+  replace: true
+  controller: ($scope, $modal) ->
+
+    $scope.$on 'modalOpened', ->
+      $scope.startMenuOpen = false
+
+    $scope.toggleStartMenu = ->
+      $scope.startMenuOpen = !$scope.startMenuOpen
--- a/lineman/app/components/start_menu/start_menu.haml
+++ b/lineman/app/components/start_menu/start_menu.haml
@@ -1,0 +1,9 @@
+.start-menu
+  .start-menu__modal-backdrop.modal-backdrop{ng-show: 'startMenuOpen', modal-window: true}
+  %ul.start-menu__options.start-menu__over-modal{ng-if: 'startMenuOpen'}
+    %start_menu_option{text: 'invite_people', icon: 'user', action: 'invitePeople'}
+    %start_menu_option{text: 'start_group', icon: 'group', action: 'startGroup'}
+    %start_menu_option{text: 'start_thread', icon: 'pencil', action: 'startThread'}
+  %a.start-menu__start-button.start-menu__menu-button.start-menu__over-modal{href: '', ng-click: 'toggleStartMenu()'}
+    %i.fa.fa-plus.fa-2x{ng-if: '!startMenuOpen'}
+    %i.fa.fa-times.fa-2x{ng-if: 'startMenuOpen'}
--- a/lineman/app/components/start_menu/start_menu.scss
+++ b/lineman/app/components/start_menu/start_menu.scss
@@ -1,0 +1,50 @@
+.start-menu {
+  position: absolute;
+  bottom: 10px;
+  right: 10px;
+  text-align: right;
+}
+
+.start-menu__modal-backdrop {
+  z-index: 500;
+}
+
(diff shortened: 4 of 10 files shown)


lineman/app/components/mixins.scss
```diff
@@ -114,7 +114,7 @@ $navbar-height: 45px;
 #turn-off-angular {
   z-index: 100;
   position: fixed;
-  right: 10px;
+  left: 10px;
```
gdpelican: I just bumped the turn off angular link to the other side so it doesn't block the start button

lineman/app/components/start_menu/start_menu.coffee
```diff
@@ -0,0 +1,12 @@
+angular.module('loomioApp').directive 'startMenu', ->
+  scope: {}
+  restrict: 'E'
+  templateUrl: 'generated/components/start_menu/start_menu.html'
+  replace: true
+  controller: ($scope, $modal) ->
+
+    $scope.$on 'modalOpened', ->
+      $scope.startMenuOpen = false
```
gdpelican: We want to know when the modal opens, so we can close the menu.

It strikes me there may be other cases where globally notifying the app that a modal is opening might be useful.

lineman/app/components/thread_page/discussion_form/discussion_form_controller.coffee
```diff
@@ -1,5 +1,5 @@
-angular.module('loomioApp').controller 'DiscussionFormController', ($scope, $controller, $location, discussion, UserAuthService, CurrentUser) ->
-  $scope.discussion = discussion
+angular.module('loomioApp').controller 'DiscussionFormController', ($scope, $controller, $location, discussion, UserAuthService, CurrentUser, Records) ->
+  $scope.discussion = discussion or Records.discussions.initialize()
```
gdpelican: If we don't get a discussion passed, initialize a new blank one.

lineman/app/services/modal_service.coffee
```diff
@@ -0,0 +1,25 @@
+angular.module('loomioApp').factory 'ModalService', ($modal, $rootScope) ->
+  new class ModalService
```
gdpelican: I was sad with how we're doing modals. I think this is much better.

lineman/app/services/modal_service.coffee
```diff
@@ -0,0 +1,25 @@
...
+    openModal: (name, options = {}) ->
+      return console.log "Invalid modal name: #{name}" unless @modals[name]
+      $rootScope.$broadcast 'modalOpened'
+      $modal.open
+        templateUrl: @modals[name].template
+        controller:  @modals[name].controller
+        resolve:     @setLocals name, options
+
+    setLocals: (name, options) ->
+      locals = {}
```
gdpelican: I think I need to revisit this when we start passing objects to the modal.

It doesn't affect the start menu, so I'm not bothering with it too hard right now.

The intended effect is to pass resolve objects to the modal based on an options hash.
The code I want to write:
```
ModalService.open('discussionForm', { discussion: $scope.discussion })
```

lineman/app/services/modal_service.coffee
```diff
@@ -0,0 +1,25 @@
...
+      _.each @modals[name].locals, (local) ->
+        locals[local] = options[local]
+      locals
+
+    modals:
```
gdpelican: A list of modals with their associated templates and controllers.

This will hopefully replace DiscussionFormService, ProposalFormService, and the myriad of places where we're duplicating modal-calling code.

gdpelican:
This is CR-ready @robguthrie

lineman/app/components/start_menu/start_menu.haml
```diff
@@ -0,0 +1,9 @@
+.lmo-start-menu
```
robguthrie: Right now the convention we're following is we only prefixing lmo onto utility classes. Can we follow that here? or there something clashing with start-menu?

gdpelican: Nah, I'll change it. I guess I was thinking 'start-menu' seems pretty clash-able, but adding lmo to the beginning of it doesn't really help that any.

lineman/app/services/modal_service.coffee
robguthrie: I think this looks pretty good!  May I suggest that it seems like a case for inheritance so it can follow open/closed principal (Closed for modification, open for extension). 

InvitePeopleModal and StartThreadModal inherit from BaseModal, and their unique configs live in the sub classes.

lineman/app/services/modal_service.coffee
robguthrie: yay thanks for trying to get on top of this. I think you're on track but I don't think this list of modal config options in here is right. it seems weird that the modal service would need to be added to every time you want a new kind of modal.

lineman/app/services/modal_service.coffee
gdpelican: Yeah that sounds good.

lineman/app/components/start_menu/start_menu.scss
```diff
@@ -0,0 +1,50 @@
+.start-menu {
+  position: absolute;
```
robguthrie: should be position: fixed I'm pretty sure

robguthrie:
wonderful. I'm sure that the start button needs to be position fixed but we can patch that later. DEPLOY!

PR #2226, opened 2015-06-11
Angular vote modal

robguthrie:
not ready but I reckon it's not far off!


app/serializers/group_serializer.rb
```diff
@@ -13,6 +13,7 @@ class GroupSerializer < ActiveModel::Serializer
              :members_can_raise_proposals,
              :members_can_vote,
              :memberships_count,
+             :members_count,
```
gdpelican: What's the difference between this and memberships_count?

robguthrie: memberships_count is rubbish and should not be trusted. It's just a counter_cache: true which does not take into account suspended, archived.

gdpelican: Okay, but I'd prefer to then throw it away and reimplement it, rather than
putting two different implementations in there. I think we use
memberships_count somewhere, and I'm about to use it in the invite people
modal.

robguthrie: yes I reckon we throw it away too.

PR #2228, opened 2015-06-11
Add host info to window.Loomio

gdpelican:


gdpelican:
I put this in the last PR. Then I didn't push it up. :see_no_evil:

PR #2229, opened 2015-06-11
Add start group workflow

gdpelican:
Includes adding the group form modal, and the group form modal.

Cut from `angular-start-button`; I'm planning on demoing this branch tomorrow.


robguthrie:
Hey James, I think we should be creating a StartGroupForm rather than changing the GroupForm.

robguthrie:
I've cherry picked this into angular-vote-modal. Thanks James

PR #2231, opened 2015-06-11
Prefer request.domain over request.host for hostInfo

gdpelican:


gdpelican:
Subdomains!? We don't need no subdomains! (╯°□°)╯︵ ┻━┻

PR #2232, opened 2015-06-11
Angular invitations

gdpelican:
WIP, no merge yet!


robguthrie:
@gdpelican I've freshened this up against master and resolved some merge conflicts. Also I sneaked a change in .. moving uses_markdown to the discussion_form and comment_form as hidden inputs so one did not need to remember to set them manuall when calling build etc.

gdpelican:
Did that work? I had some issues with that approach where angular was
ignoring me and my hidden inputs.

robguthrie:
xxx

PR #2234, opened 2015-06-14
Fix runaway vote percentage

gdpelican:


robguthrie:
as discussed please just round to nearest whole number

PR #2235, opened 2015-06-14
Fix update vote

gdpelican:


lineman/app/components/thread_page/proposal_positions_panel/proposal_positions_panel.coffee
```diff
@@ -6,7 +6,8 @@ angular.module('loomioApp').directive 'proposalPositionsPanel', ->
   controller: ($scope, Records, CurrentUser, ModalService, VoteForm) ->
 
     $scope.changeVote = ->
-      ModalService.open VoteForm, proposal: -> $scope.proposal
+      ModalService.open VoteForm, 
+        proposal: -> $scope.proposal.clone()
```
robguthrie: why clone proposal here? (we operate on the vote) Does no harm I can see though.

lineman/app/components/mini_proposal_pie_chart/mini_proposal_pie_chart_controller.coffee
```diff
@@ -16,12 +16,10 @@ angular.module('loomioApp').controller 'MiniProposalPieChartController', ($scope
   refreshPieChartData = ->
     return unless $scope.proposal
     counts = $scope.proposal.voteCounts
-    hasAnyVotes = counts.yes + counts.abstain + counts.no + counts.block > 0
     $scope.pieChartData[0].value = counts.yes
     $scope.pieChartData[1].value = counts.abstain
     $scope.pieChartData[2].value = counts.no
     $scope.pieChartData[3].value = counts.block
-    $scope.pieChartData[4].value = if hasAnyVotes then 0 else 1
+    $scope.pieChartData[4].value = if _.some [counts.yes, counts.abstain, counts.no, counts.block] then 0 else 1
```
robguthrie: seems like a refactor should be happening

spec/controllers/api/votes_controller_spec.rb
```diff
@@ -23,58 +24,66 @@
...
-        #get :my_votes, discussion_id: discussion.id, format: :json
-        #json = JSON.parse(response.body)
-        #expect(json.keys).to include *(%w[votes])
-        #motions = json['votes'].map { |v| v['id'] }
-        #expect(motions).to include my_vote.id
-        #expect(motions).to_not include my_older_vote.id        
-      #end
-    #end
-  #end
+  describe 'index' do
```
robguthrie: nice one. Thank you

lineman/app/components/thread_page/proposal_positions_panel/proposal_positions_panel.coffee
gdpelican: Yep, I was just being paranoid, I think.

I've been having a look at the Modal Service and keep feeling like there's a better way to handle resolves, specifically that the form should know about what resolves it's expecting and have a sensible default (Records.proposals.build()) for when it doesn't receive one

Something like
```
# Edit Proposal:
ModalService.open ProposalForm, { proposal: $scope.proposal }

# New proposal:
ModalService.open ProposalForm, { discussion_id: $scope.discussion.id }

# ProposalForm:
$scope.proposal = options['proposal'] or Records.proposals.build(discussion_id: options['discussion_id'])
```

With some element to pass the options through in between
(It's still a half-baked idea, sorry)

lineman/app/components/mini_proposal_pie_chart/mini_proposal_pie_chart_controller.coffee
gdpelican: Yeah, there's likely a fancy elegant `_.map` solution here with `$scope.proposal.voteCounts -> $scope.pieChartData` but I wasn't up to puzzling it out at the time.

lineman/app/components/thread_page/proposal_positions_panel/proposal_positions_panel.coffee
robguthrie: FYI: In both those examples you need to wrap the resolve value in a function or I don't think they'll work.

We could add a key to the modal/form: "requiredArguments" and list resolves that must be supplied or raise an exception.

lineman/app/components/mini_proposal_pie_chart/mini_proposal_pie_chart_controller.coffee
robguthrie: even just naming this _.some block with a method name would fix it for me

gdpelican:
@HSalmon let's write an e2e for this tomorrow, then we can merge this in.

lineman/app/models/base_model.coffee
```diff
@@ -5,13 +5,14 @@ angular.module('loomioApp').factory 'BaseModel', ->
...
 
-    constructor: (recordsInterface, data) ->
+    constructor: (recordsInterface, data, postInitializeData = {}) ->
       @errors = {}
       @processing = false
       Object.defineProperty(@, 'recordsInterface', value: recordsInterface, enumerable: false)
       Object.defineProperty(@, 'recordStore', value: recordsInterface.recordStore, enumerable: false)
       Object.defineProperty(@, 'restfulClient', value: recordsInterface.restfulClient, enumerable: false)
       @initialize(data)
+      _.merge @, postInitializeData
```
gdpelican: Here it wasn't setting up views for the clone because it had no id at the time, so have to set the id on the model before we finish the initialize method.

PR #2236, opened 2015-06-14
Proposal lintel

gdpelican:
Start / view proposal from anywhere on thread page

- Display start proposal in lintel when start proposal button is not visible
- Display proposal info in lintel when proposal is not visible
- Add click events which scroll to the appropriate parts of the thread


lineman/app/components/proposal_form/proposal_form.coffee
```diff
@@ -0,0 +1,15 @@
+angular.module('loomioApp').factory 'ProposalForm', ->
+  templateUrl: 'generated/components/proposal_form/proposal_form.html'
+  controller: ($scope, $modalInstance, proposal, FlashService) ->
+    $scope.proposal = proposal
+
+    $scope.submit = ->
+      proposal.save().then ->
+        $modalInstance.close()
```
robguthrie: you can also call $scope.$close() and save injecting the instance I think.

robguthrie:
This looks amazing, James. Shall I merge?

gdpelican:
Not just yet; looking through the diff I've got a little cleanup to do (injecting things I don't need, etc.)

lineman/app/components/proposal_form/proposal_form.coffee
gdpelican: Ah, that's so great.

lineman/app/components/start_proposal_button/start_proposal_button.coffee
```diff
@@ -0,0 +1,12 @@
+angular.module('loomioApp').directive 'startProposalButton', ->
+  scope: { discussion: '='}
+  restrict: 'E'
+  templateUrl: 'generated/components/start_proposal_button/start_proposal_button.html'
+  replace: true
+  controller: ($scope, Records, ModalService, ProposalForm, CurrentUser, AbilityService) ->
+
+    $scope.canStartProposal = ->
+      AbilityService.canStartProposal $scope.discussion
```
gdpelican: What do you think of this AbilityService, @robguthrie?

My discomforts with current permissions:
- Injecting a single instance from the record store as a dependency is a little odd; feels much more natural to have a service
- There's a ton of permissions, and having them on the user model will bloat it quite a bit
- It feels odd to be able to go 'someOtherUser.canCreateProposals()' or something like this; I think we only need to know about permissions for our current user, no?

gdpelican:
Also the mini pie chart in the lintel isn't showing up at the moment, because chart.js had another pain in the butt up its sleeve.

robguthrie:
I'm thinking we should consider writing our own pie chart. Also really into
AbilityService. Awesome jawsome. You're super.

PR #2237, opened 2015-06-15
locale update - greek frontpage

mixmix:


mixmix:
@robguthrie  good to go !

PR #2243, opened 2015-06-16
fix lintel and star alignment

rdbartlett:
My dev environment is broken so this is untested but I'm pretty confident in these little css tweaks


robguthrie:
thanks rich

PR #2246, opened 2015-06-16
Disappear fab background on click

gdpelican:


lineman/app/components/start_menu/start_menu.haml
```diff
@@ -1,5 +1,5 @@
 .start-menu
-  .start-menu__modal-backdrop.modal-backdrop{ng-show: 'startMenuOpen', modal-window: true}
+  .start-menu__modal-backdrop.modal-backdrop{ng-show: 'startMenuOpen', modal-window: true, ng-click: 'toggleStartMenu() ||'}
```
gdpelican: This is a bit effed; the 'modal-window: true' bit appends a 'close($event)' ng-click event automagically, so this looks like 

`ng-click: 'toggleStartMenu() || close($event)'`

I tried adding a $scope.close method to the controller, but didn't work. Too easy a fix to spend too much time on, I figure... this can be one of those things the :alien: are baffled by in 300 years.

robguthrie: another indication we should ditch bootstrap's modal stuff. cool thanks

robguthrie:
merge?

gdpelican:
Merge!

PR #2248, opened 2015-06-16
refactor start-proposal-card to follow bem and fix broken e2e

robguthrie:


robguthrie:
@gdpelican fyi I refactored this.. only as far as I could be bothered. I found that calling a bunch of elements a "button" at the component level was super confusing. To me a button element is the button. Anything more than that is a panel or card.

I think we'll want to pull this back out into a card again, but I resisted doing that as it looked like you may have had trouble with in-view lintel stuff when this was in a seperate component, which may have been why we ended up with the button component in the first place.

Anyway.. this fixes the last of the broken e2e tests and makes stuff follow BEM a little bit more closely.

gdpelican:
Yeah I just wanted to make sure that the element in the lintel and the
element in the thread did the same thing without duplication... I imagine
there will be other cases where we have a button or action out of context
and it will want to behave the same as the action in the context... I
thought of it as more of a directive and less of a component, although I
understand that those two things live pretty close together here.

robguthrie:
oh I had not realised it was used more than once... now I understand a bit better, thanks.

robguthrie:
I'm sorry I really messed this up by not understanding what was going on. I'm gonna throw it out and start again.

PR #2252, opened 2015-06-18
Edit group form

gdpelican:
Adds modals for Leave group, archive group, and edit group form

Edit group form should be ready for a first pass, there are some permissions / redirect issues with leave / archive group still.

screenshot from 2015-06-18 11 58 58


config/locales/client.en.yml
```diff
@@ -324,6 +324,36 @@ en:
       private: 'This group is only visible to members'
       parent_members: 'This group is visible to members of {{parentGroupName}} and members of {{groupName}}'
       public: 'This group is visible on the web'
+    edit_group:
```
robguthrie: can we call this key edit_group_form ?

lineman/app/components/group_form/edit_group_form.coffee
```diff
@@ -0,0 +1,52 @@
+angular.module('loomioApp').factory 'EditGroupForm', ->
+  templateUrl: 'generated/components/group_form/edit_group_form.html'
```
robguthrie: consistency says this should be edit_group_form/edit_group_form  would that work here?

lineman/app/components/group_form/leave_group_form.coffee
```diff
@@ -0,0 +1,13 @@
+angular.module('loomioApp').factory 'LeaveGroupForm', ->
+  templateUrl: 'generated/components/group_form/leave_group_form.html'
```
robguthrie: leave_group_form/leave_group_form ?

PR #2255, opened 2015-06-18
little style tweaks + perf improvement

rdbartlett:
@robguthrie here are your changes from this morning + couple wee css adjustments

### DIFF
--- a/lineman/app/components/thread_page/comment_form/comment_form.scss
+++ b/lineman/app/components/thread_page/comment_form/comment_form.scss
@@ -4,6 +4,8 @@
 
 .comment-form {
   @include card;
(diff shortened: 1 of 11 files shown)


lineman/app/components/thread_page/comment_form/comment_form.scss
```diff
@@ -4,6 +4,8 @@
 
 .comment-form {
   @include card;
+  /*make room for the "turn off angular" link */
+  margin-bottom: 5em;
```
gdpelican: Why em?

rdbartlett: no reason. temporary fix until we design a proper solution for interface
switching

PR #2259, opened 2015-06-21
allow people to set their own subject for invitations

fauno:
where should i put the spec? :)

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -39,7 +39,8 @@ class InvitationsController < ApplicationController
       InvitationService.invite_to_group(recipient_emails: @invite_people_form.emails_to_invite,
                                         message: @invite_people_form.message_body,
                                         group: @invitable,
-                                        inviter: current_user)
+                                        inviter: current_user,
+                                        subject: params[:invite_people_form][:subject])
     elsif @invitable.kind_of?(Discussion)
       MembershipService.add_users_to_discussion(users: @invite_people_form.members_to_add,
                                                 discussion: @discussion,
--- a/app/forms/invite_people_form.rb
+++ b/app/forms/invite_people_form.rb
@@ -3,7 +3,8 @@ class InvitePeopleForm
   include ActiveModel::Conversion
   extend ActiveModel::Naming
 
-  attr_accessor :recipients, :message_body, :parent_members_to_add
+  attr_accessor :recipients, :message_body, :parent_members_to_add,
+                :subject
 
   validates_presence_of :recipients
 
--- a/app/forms/invite_people_form.rb
+++ b/app/forms/invite_people_form.rb
@@ -3,8 +3,7 @@ class InvitePeopleForm
   include ActiveModel::Conversion
   extend ActiveModel::Naming
 
-  attr_accessor :recipients, :message_body, :parent_members_to_add,
-                :subject
+  attr_accessor :recipients, :message_body, :parent_members_to_add, :subject
 
   validates_presence_of :recipients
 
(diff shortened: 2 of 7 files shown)


app/controllers/invitations_controller.rb
```diff
@@ -39,7 +39,8 @@ def create
       InvitationService.invite_to_group(recipient_emails: @invite_people_form.emails_to_invite,
                                         message: @invite_people_form.message_body,
                                         group: @invitable,
-                                        inviter: current_user)
+                                        inviter: current_user,
+                                        subject: params[:invite_people_form][:subject])
```
gdpelican: This is unfortunately an email injection risk, I'm pretty sure. 
I'd much prefer if we put the subject into the @invite_people_form and sanitized it there.

app/forms/invite_people_form.rb
```diff
@@ -3,7 +3,8 @@ class InvitePeopleForm
   include ActiveModel::Conversion
   extend ActiveModel::Naming
 
-  attr_accessor :recipients, :message_body, :parent_members_to_add
+  attr_accessor :recipients, :message_body, :parent_members_to_add,
```
gdpelican: One line plz!

gdpelican:
Thanks @fauno! I've added a couple notes including a security concern.

Hrm... I see we don't have a test for either `inviations#create` or the `invite_people_form` ... sigh. I think writing a controller test for the create action for this would be the best not-that-much-work way to go. :balloon:!

app/controllers/invitations_controller.rb
fauno: ehr... yes, i'll strip spaces and new lines and put a limit, maybe 100 characters? (doesn't the mail gem take care of this?)

fauno: i just tested it and the mail gem removes the special chars on its own, so i can't make header injection by sending a subject like this "a subject\r\nBcc: included@header.com", it turns out "a subject Bcc: included@header.com"

robguthrie:
Hi @fauno Are you able to talk a bit about what value this feature brings? 
We're really pleased that you've worked to make this contribution but we're not sure that we want to add an extra field to the invitation form at this stage. 

We would love to hear your use case for this, and what your motivation was for doing it.

Also if you like, we could have a skype/hangout and chat about other development opportunities within Loomio if you're interested. Let me know!

fauno:
hey! we're starting to use loomio at [partido pirata en argentina](https://consenso.partidopirata.com.ar) and we needed to invite everyone in our main mailing list.  we've tried different message combinations but the default subject invitations seemed a bit stiff (and some say it looks like one of those bogus social network messages...) so we wanted to try different subjects too.  this change seemed simple enough to do it, instead of changing the locale and re-deploying loomio everytime, so i went with it.

we're planning more changes to loomio, some of which are pirate specific (for instance, groups are called ships :), but others may be useful so we'll proppose them as pull requests if you're ok with this.

i'm in irc.pirateirc.net or irc.freenode.net under this nick if you want to ping me :)

robguthrie:
Thanks for the explanation @fauno. If loomio had a link where "Anyone with the link can join", so you could send that link to your mailing list manually, would that have been a better experience than inviting everyone with their own link?

robguthrie:
@fauno I'm going to close this because it's for the older interface ( Do you know about the almost ready for release angularjs interface?) and it's not something we want to support. However I'm so pleased to see what you do with Loomio please feel free to keep making pull requests, I really appreciate it!

fauno:
@robguthrie i've seen it in loomio but couldn't make it work on our instance... is there any doc on it or extra step involved?

robguthrie:
Yep. You need to run lineman build and copy some files into public. We have
a deploy.sh script for loomio.org - you should check it out to understand
what is required - but we need to generalise it so others can use it.
@gdpelican. Instructions on lineman are in the Readme.md

PR #2262, opened 2015-06-22
Invitation form flash test

HSalmon:
Failures indicate that:
- inviting a Loomio user by email address doesn't currently work
- when an invitee is a member of the group into which they are being invited then an invitation is still sent...or it appears that way anyway...


robguthrie:
Wow. That's so great to know. Thank you @HSalmon!

config/locales/client.en.yml
```diff
@@ -380,7 +380,11 @@ en:
     send_invite: 'Send {{count}} invites'
     to: 'to: '
     default_message: "Hey there, come make decisions with us! We're using a collaborative decision-making tool called Loomio"
-
+    messages:
+      members_added_and_emails_sent: '{{members}} member(s) added and {{emails}} invitation(s) sent'
+      members_added: '{{members}} member(s) added'
+      emails_sent: '{{emails}} invitation(s) sent'
+      no_invitations: 'No invitation(s) sent!'
```
robguthrie: Will we need the "no invitations sent" case? I think our client will be smart enough to prevent this happening. IE: I imagine we'll have client side email validation and disable send button until the form is valid.

HSalmon:
@robguthrie, @gdpelican, updated branch with invitation tests.

config/locales/client.en.yml
gdpelican: Yeah, we can take this out I think.

HSalmon:
Anything more need to happen here before it can be merged @robguthrie, @gdpelican?

PR #2263, opened 2015-06-22
Memberships page

gdpelican:
Now we have a memberships page!
screenshot from 2015-06-22 17 44 08

And a way to get to it!
screenshot from 2015-06-22 17 43 27


gdpelican:
Some todos:
- Add flash messaging
- Add confirm modal for removing a user.

robguthrie:
Hey @rdbartlett and @gdpelican I have a suggestion for this page: We should always load all the memberships (ie: no pagination) in the group and have the ability to filter them client side by name or email address.

rdbartlett:
Yep agree. Can we make this iteration have no pagination and add a filter in the next iteration?

rdbartlett:
Hey @gdpelican if this page is being viewed by a coordinator, can we show email addresses rather than usernames please? It's not a great feature but I believe some coordinators rely on this page as a canonical list of addresses.

gdpelican:
Aw, I didn't tell anyone and probably should have, but I was thinking that
we load some decent chunk to start (25?), then allow for filtering
afterwards (ie, you put in a string and it searches and returns matches
from the server, same as invitations). I think allowing potentially
hundreds/thousands of members to show up in that ng-repeat is nebulous.

Filter is pretty simple to add and I intend to at least have a go at it in
this sprint.

I'm pretty apprehensive of giving coordinators their group members email
addresses (do group members know their email address is being given away?
We need to at least make a note of that when signing up somehow), and think
we should strongly move away from this. A medium term solution might be to
have a contact form modal next to each member which coordinators can
access, which could send them an email without them knowing the address.

rdbartlett:
ok why don't you leave the emails out and see if people complain, and we
can design a good solution at that point?

PR #2264, opened 2015-06-22
Hide thread actions on small screens

HSalmon:


robguthrie:
YEA! Isn't it nice when you know that nothing could be simpler?

robguthrie:
Thank you!!

PR #2274, opened 2015-06-24
Fix current proposal scrolling

gdpelican:


spec/controllers/api/memberships_controller_spec.rb
```diff
@@ -99,4 +99,28 @@
...
   end
 
+  describe 'group_preview' do
+    context 'success' do
+      before do
+        @preview_group = create :group
+      end
+
+      it 'sorts by coordinators first' do
+        get :group_preview, group_id: group.id, per: 1
```
robguthrie: great naming!

PR #2276, opened 2015-06-25
Fix failing invitations e2ez

HSalmon:


app/controllers/api/restful_controller.rb
```diff
@@ -92,7 +92,7 @@ def parse_date_parameters
   end
 
   def page_collection(collection)
-    collection.offset(params[:from]).limit(params[:per] || default_page_size)
+    collection.offset(params[:from]).limit(params[:per].try(:to_i) || default_page_size)
```
robguthrie: I think this should be `params.fetch(:per, default_page_size)` Fetch is a great way to provide a default when the key does not exist. Also do we want to apply a fetch on the params[:from] so it's params.fetch(:from, 1) ?

gdpelican: params,fetch(:per, default_page_size) doesn't solve this particular issue,
because our problem was that params[:per] was "5", which was getting
compared to an integer, which was blowing up. I'm not too confident in this
fix though; it's not a problem we're seeing in master at the moment, so I'm
tempted to just close this and apply this if we run into it again.

offset(nil) is the same as offset(0), so fetching params[:from] I don't
think has a benefit either, although I do agree that in general we want to
be fetching instead of calling params directly if we have default values.

robguthrie:
can we close this @gdpelican ?

gdpelican:
Yes I believe so.

PR #2277, opened 2015-06-25
Async load memberships

robguthrie:


lineman/app/models/user_model.coffee
```diff
@@ -12,12 +12,6 @@ angular.module('loomioApp').factory 'UserModel', (BaseModel) ->
       @membershipsView   = @recordStore.memberships.belongingTo(userId: @id)
       @notificationsView = @recordStore.notifications.belongingTo(userId: @id)
       @contactsView      = @recordStore.contacts.belongingTo(userId: @id)
-      @groupsView        = @recordStore.groups.belongingTo(id: { $in: @groupIds() })
-      @parentGroupsView  = @recordStore.groups.collection.addDynamicView()
-      @parentGroupsView.applyFind(parentId: {$eq: null}).applyFind(id: {$in: @groupIds()})
```
robguthrie: James I uncovered a bug with this.. it's taking a copy of group ids, so if memberships get added later they don't show up in here. Seems like views are only going to work for belongs_to.. which is fine

PR #2279, opened 2015-06-29
Expose public motion endpoint

gdpelican:
Prelim work for DemOS integration; don't merge yet.


robguthrie:
Hey James this looks good. I'm just holding a question about the 'public' bit. My assumption has always been that public and private discussions would be available via the same api, but private discussions would need authentication. (just like how beta works with /d and /g). I see this is a different format to our normal api stuff and I was wondering about naming it /api/popolo/motions rather than /api/v1/public/motions. So the format (v1, popolo) is the bit that changes.

gdpelican:
Yeah, my original thought was that exposing the API to non-logged-in users
at the moment would be too much work, but since this path has been a little
rocky as well I might give it a go.

I think the type of serialization wants to simply be something you can
control via params, such as
```
/api/v1/motions?spec=popolo
```
or maybe like
```
  /api/v1/motions/popolo
```

robguthrie:
Privacy is a preference on the resource and can change, we want to keep a constant url even when we change the privacy setting of a thread. (eg: oh hey I can't read that thread, can you make it public please?)

We'll bump the version (v1) number when we change our data model in an incompatible way, so it seems to me that this different format should have it's own identifier in the path.

If we keep v1 AND popolo in the same url then it does not make any sense to me in the way I designed it.

app/models/motion.rb
```diff
@@ -47,7 +47,7 @@ class Motion < ActiveRecord::Base
   scope :lapsed_but_not_closed,    -> { voting.lapsed }
   scope :closed,                   -> { where('closed_at IS NOT NULL').order(closed_at: :desc) }
   scope :order_by_latest_activity, -> { order('last_vote_at desc') }
-  #scope :visible_to_public,        -> { joins(:discussion).merge(Discussion.public) }
```
robguthrie: Is merge broken for you? I've always liked that we did not have to duplicate this scope

robguthrie:
I've had a wee break and am feeling much more prepared to discuss this like a nice person.

On the privacy thing, I'm feeling pretty strongly that a resource should have the same url no matter the privacy setting, just because privacy is an attribute of the resource which can change.

On the format issue, I'd be happy to discuss what the v1 part means and when we might change it.
I've always considered that v1 means the current Loomio data model, which is not formally defined yet, but guarantees attributes like closing_at, title, description. It also means that related records will be side-loaded and you can expect a response to include all the belongs_to records that relate to the record you asked for. 

I hope this makes sense to you and you understand why I thought popolo was a version code rather than something that could exist at the same time as the v1 code. 

Feel free to let me know what you think the v1 refers to if you like

gdpelican:
I'm hesitant to go that direction because I think it's a non-zero chance that the popolo spec itself will change, meaning it should be versioned and is not a version on its own.

In any case it's becoming clear that we're not quite ready to be doing this work, but I'm happy to put together a solution that works for tomorrow and then revisit it down the road.

robguthrie:
Oh I was thinking it could be popolo-v1 etc... like a namespaced version.

robguthrie:
or just popolo for now and popolo2 in the future

app/models/motion.rb
gdpelican: Ah, I had just assumed that was garbage since it was commented out and `Discussion.public` wasn't a thing. I'll swap it to merge and use `Discussion.visible_to_public` instead

app/controllers/api/base_controller.rb
```diff
@@ -1,7 +1,7 @@
 class API::BaseController < ActionController::Base
   include CurrentUserHelper
   include ::ProtectedFromForgery
-  before_filter :require_authenticated_user
+  before_filter :require_authenticated_user, except: :index
```
gdpelican: I'm not ready to fully commit to making the API public-accessible, but I think the index action should be a good place to start.

Next steps will be to go through groups, discussions, maybe comments controllers, and add the `visible_to_public` methods to them.

app/controllers/api/restful_controller.rb
```diff
@@ -95,10 +83,22 @@ def page_collection(collection)
...
+      public_records
+    end
+  end
+
   def visible_records
     raise NotImplementedError.new
   end
 
+  def public_records
+    resource_class.visible_to_public.order(created_at: :desc)
```
gdpelican: This is convenience more than anything else, but it works for groups, discussions, and motions (and could be easily adopted for comments too)

config/routes.rb
```diff
@@ -44,6 +44,10 @@
   ActiveAdmin.routes(self)
 
   namespace :api, path: '/api/v1', defaults: {format: :json} do
+    namespace :popolo do
```
gdpelican: My gripe with moving this out of the api namespace at the moment is that the RestfulController is in the API namespace, so otherwise we get something like

```
class Popolo::MotionsController < API::RestfulController # hrm..
```

And a folder structure like:
```
- popolo
  - motions_controller # inherits from a different folder?
- api
  - restful_controller
```

spec/controllers/api/popolo/motions_controller_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  let(:public_discussion) { create :discussion, group: public_group, private: false }
+  let(:private_discussion) { create :discussion, group: private_group }
+
+  let(:public_motion) { create :motion, discussion: public_discussion }
+  let(:private_motion) { create :motion, discussion: private_discussion }
+
+  before { public_motion; private_motion }
+
+  describe 'index' do
```
gdpelican: Yay it does all the filtering / paging nonsense already without me doing anything!

PR #2280, opened 2015-07-01
Show undecided members

HSalmon:


app/controllers/api/did_not_votes_controller.rb
```diff
@@ -0,0 +1,22 @@
+class API::DidNotVotesController < API::RestfulController
+
+  def create
+    raise NotImplementedError.new
```
gdpelican: I'd be happy to not have these NotImplementedError stubs; so long as we don't have a route in place for create/update/destroy, we won't reach them.

robguthrie: yeah.. ok cool

lineman/app/components/thread_page/undecided_panel/undecided_panel.coffee
```diff
@@ -0,0 +1,15 @@
+angular.module('loomioApp').directive 'undecidedPanel', ->
+  scope: {proposal: '='}
+  restrict: 'E'
+  templateUrl: 'generated/components/thread_page/undecided_panel/undecided_panel.html'
+  replace: true
+  controller: ($scope, Records, CurrentUser) ->
```
gdpelican: I think we can take out these dependencies, no?

spec/controllers/api/did_not_votes_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    MotionService.close(motion)
+  end
+
+  describe 'index' do
+
+    context 'success' do
+      it 'returns did not votes for a motion' do
+        get :index, motion_id: motion.id, format: :json
+        json = JSON.parse(response.body)
+        expect(json.keys).to include *(%w[did_not_votes])
```
gdpelican: I'd love to see a check for permissions in this test, the way we do with most of the other index actions

```
group_member = create :user
not_group_member = create :user
groups.members << group_member
...
expect(user_ids).to include group_member.id
expect(user_ids).to_not include not_group_member.id
```
^^ (The gist, anyway)

lineman/spec-e2e/proposal_spec.coffee
```diff
@@ -89,3 +89,30 @@ describe 'Proposals', ->
...
+        it 'hides all undecided members when undecided panel is open and hide link is clicked', ->
+          proposalsHelper.clickHideUndecidedLink()
+          expect(proposalsHelper.undecidedPanel().getText()).toContain('Show undecided members')
+
+      describe 'when proposal is closed', ->
+
+        beforeEach ->
+          threadHelper.loadWithClosedProposal()
+
+        xit 'shows all undecided members when the show link is clicked', ->
```
gdpelican: Hm, is this bit not working?

lineman/app/components/thread_page/proposal_positions_panel/proposal_positions_panel.coffee
```diff
@@ -5,6 +5,8 @@ angular.module('loomioApp').directive 'proposalPositionsPanel', ->
   replace: true
   controller: ($scope, Records, CurrentUser, ModalService, VoteForm) ->
 
+    $scope.undecidedPanelOpen = false
```
gdpelican: Don't need these changes I don't think?

app/serializers/did_not_vote_serializer.rb
```diff
@@ -0,0 +1,15 @@
+class DidNotVoteSerializer < ActiveModel::Serializer
+  embed :ids, include: true
+  attributes :id, :created_at
+
+  has_one :user, serializer: UserSerializer, root: 'users'
+  has_one :proposal, serializer: MotionSerializer, root: 'proposals'
```
gdpelican: man, this thing again, heh.

I'd be curious if `has_one :motion, key: :proposal, serializer: MotionSerializer, root: 'proposals'` would work.

app/controllers/api/did_not_votes_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def destroy
+    raise NotImplementedError.new
+  end
+
+  private
+
+  def visible_records
+    load_and_authorize :motion
+    @motion.did_not_votes.order(:id)
```
gdpelican: Why order by id?

robguthrie: not a real good reason at this stage. was imagining that when we paginate it we need an order

lineman/app/components/thread_page/undecided_panel/undecided_panel.coffee
```diff
@@ -0,0 +1,15 @@
...
+
+    $scope.undecidedPanelOpen = false
+
+    $scope.showUndecided = ->
+      $scope.proposal.fetchUndecidedMembers()
```
gdpelican: We don't generally have fetch methods on models, so this is a bit of a departure from the convention.
Why not 
```
Records.proposals.fetchUndecidedMembers($scope.proposal)
```
?

lineman/app/components/thread_page/proposal_positions_panel/proposal_positions_panel.coffee
robguthrie: Not necessary but makes it easier to say "is false or true" rather than "is undefined or false or true"

lineman/app/components/thread_page/undecided_panel/undecided_panel.coffee
robguthrie: true! thanks James

lineman/spec-e2e/proposal_spec.coffee
HSalmon: Hey James - it is working but it's a little complicated to test - you need a closed proposal, then a new member added, and I couldn't get the tests to work so I'm gonna have another stab again soon

spec/controllers/api/did_not_votes_controller_spec.rb
robguthrie: I don't get your gist sorry. We have checks that the user is able to load the motion...and we know the did not votes are from the motion. I would have thought the concern here is that you can get did not votes for a motion you're not supposed to look at.

lineman/app/components/thread_page/undecided_panel/undecided_panel.coffee
robguthrie: Yea.. fair enough. thanks

lineman/app/components/thread_page/proposal_positions_panel/proposal_positions_panel.coffee
gdpelican: Oh, I was just looking at the view for this (`proposal_positions_panel/proposal_positions_panel.haml`) and don't see any reference to showUndecided / undecidedPanelOpen, so I thought these might be vestigial.

spec/controllers/api/did_not_votes_controller_spec.rb
gdpelican: Yep, I think that's what I'm getting at. Our index test for discussions looks like:

```
      it 'returns discussions filtered by group' do
        get :index, group_id: group.id, format: :json
        json = JSON.parse(response.body)
        expect(json.keys).to include *(%w[discussions])
        discussions = json['discussions'].map { |v| v['id'] }
        expect(discussions).to include discussion.id # <-- this line
        expect(discussions).to_not include another_discussion.id # <-- and this one
      end
```

Which confirms that you're a) getting discussions from the specified group and b) not getting discussions from another group leaking in.

PR #2281, opened 2015-07-02
Membership requests

HSalmon:
- Need to use MembershipRequestsService in 'approve' controller action
- Need to extend MembershipRequestsService.approve to authorize with CanCan

### DIFF
--- a/app/controllers/api/membership_requests_controller.rb
+++ b/app/controllers/api/membership_requests_controller.rb
@@ -6,17 +6,21 @@ class API::MembershipRequestsController < API::RestfulController
     respond_with_collection
   end
 
-  def responded_to
+  def previous
     load_and_authorize :group
     @membership_requests = @group.membership_requests.responded_to
     respond_with_collection
   end
 
   def approve
-    load_and_authorize :group
-    authorize! :manage_membership_requests, @group
-    @membership_request = @group.membership_requests.where(id: params[:id]).first
-    @membership_request.approve!(current_user)
+    @membership_request = MembershipRequest.find(params[:id])
+    MembershipRequestService.approve(membership_request: @membership_request, actor: current_user)
+    respond_with_resource
+  end
+
+  def ignore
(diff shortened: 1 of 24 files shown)


app/controllers/api/membership_requests_controller.rb
```diff
@@ -0,0 +1,26 @@
+class API::MembershipRequestsController < API::RestfulController
+
+  def pending
+    load_and_authorize :group
```
robguthrie: @HSalmon - another note! - I think we should be checking if the user can manage_membership_requests for this group before revealing pending and previous requests.

HSalmon:
@robguthrie, @gdpelican, failure is unrelated. Still needs some styling but wondering whether that could be applied later in the week?

screen shot 2015-07-06 at 4 29 07 pm

lineman/app/components/membership_requests_page/membership_requests_page_controller.coffee
```diff
@@ -0,0 +1,30 @@
...
+  @approve = (membershipRequest) =>
+    Records.membershipRequests.approve(membershipRequest).then ->
+      FlashService.success "membership_requests_page.messages.request_approved_success"
+
+
+  @ignore = (membershipRequest) =>
+    Records.membershipRequests.ignore(membershipRequest).then ->
+      FlashService.success "membership_requests_page.messages.request_ignored_success"
+
+  @noPendingRequests = =>
```
HSalmon: This function doesn't seem to work - had to take it out of the view...

lineman/app/components/membership_requests_page/membership_requests_page_controller.coffee
```diff
@@ -0,0 +1,30 @@
...
+    @pendingRequests.length == 0
+
```
HSalmon: @noPendingRequests function doesn't seem to work - had to take it out of the view...

PR #2282, opened 2015-07-02
Change password

gdpelican:
This is ready for a peek I think, @robguthrie.
It's still just functionality (I haven't made it match the design yet)

I'm planning on driving out the change picture stuff today as well.


app/controllers/api/restful_controller.rb
```diff
@@ -116,31 +116,35 @@ def resource_symbol
...
   end
 
-  def respond_with_collection(scope: {})
-    render json: collection, root: serializer_root, scope: scope
+  def respond_with_collection(scope: {}, serializer: resource_serializer)
+    render json: collection, root: serializer_root, scope: scope, each_serializer: serializer
   end
 
-  def respond_with_resource
+  def respond_with_resource(scope: {}, serializer: resource_serializer)
```
gdpelican: I merged in the stuff from the popolo motion endpoint to get this ability to inject what serializer to use, which comes in handy for using CurrentUserSerializer over UserSerializer for the profile page.

app/services/user_service.rb
```diff
@@ -4,9 +4,27 @@ def self.delete_spam(user)
...
+    actor.ability.authorize! :deactivate, user
+    UserDeactivationResponse.create(user: user, body: params[:deactivation_response])
+    user.deactivate!
+    user
+  end
+
   def self.update(user:, actor:, params:)
     actor.ability.authorize! :update, user
-    user.update! params
+    user.update params
```
gdpelican: We want to nab the validation errors and pass em through, rather than error if a user's not valid

app/services/user_service.rb
```diff
@@ -4,9 +4,27 @@ def self.delete_spam(user)
...
+    user
+  end
+
+  def self.change_password(user:, actor:, params:)
+    actor.ability.authorize! :update, user
+    user.errors.add :current_password, I18n.t(:"error.current_password_did_not_match") unless user.valid_password? params.delete(:current_password)
```
gdpelican: I feel like this is some of the crappier code I've written recently.. I had another implementation where I had custom validators on the User model for this stuff, but that involved hanging an extra attr_accessor off of the user model for `current_password`, as well as two whole new custom validation classes (additional complexity in user.rb = :angry:) for what amounts to these two lines of validation code.

app/controllers/api/users_controller.rb
```diff
@@ -1,17 +1,31 @@
...
 
   def update_profile
-    service.update(user: current_user, actor: current_user, params: resource_params)
-    respond_with_current_user      
+    service.update user_params
+    respond_with_resource
+  end
+
+  def change_password
+    service.change_password(user_params) { sign_in resource, bypass: true }
```
gdpelican: We need to update our session after changing our password. I like passing blocks. Thus this :angel:

lineman/app/components/change_password_form/change_password_form.coffee
```diff
@@ -0,0 +1,14 @@
...
+
+    $scope.submit = ->
+      $scope.isDisabled = true
+      Records.users.changePassword($scope.user).then ->
+        $scope.isDisabled = false
+        FlashService.success('profile_page.messages.password_changed')
+        $scope.$close()
+      , ->
+        $scope.isDisabled = false
+        # $rootScope.$broadcast 'pageError', 'cantChangePassword', $scope.user
```
gdpelican: It's coming up on a time where I'd start to feel confident about writing a generalized version of how to handle validation errors, server errors, and successes on forms in a general way. Maybe not quite yet, but SOOON

lineman/app/components/deactivate_user_form/deactivate_user_form.coffee
```diff
@@ -0,0 +1,13 @@
+angular.module('loomioApp').factory 'DeactivateUserForm', ->
+  templateUrl: 'generated/components/deactivate_user_form/deactivate_user_form.html'
+  controller: ($scope, $rootScope, $window, CurrentUser, Records) ->
+    $scope.user = CurrentUser.clone()
+
+    $scope.submit = ->
+      $scope.isDisabled = true
+      Records.users.deactivate($scope.user).then ->
+        $scope.isDisabled = false
+        $window.location.reload()
```
gdpelican: Hard reload to get back to the user sign in screen (with appropriate flash message)

lineman/app/components/profile_page/profile_page.haml
```diff
@@ -1,21 +1,30 @@
...
+      %label{for: 'user-locale-field', translate: 'profile_page.locale_label'}
+      %select.form-control.profile-page__language-input{ng-model: 'profilePage.user.selectedLocale', ng-required: 'true', ng-options: 'locale.key as locale.name for locale in profilePage.availableLocales()', ng-change: 'setCurrentPrivacy()'}
 
-    %button.btn.btn-primary.profile-page__update-button{ng-click: 'profilePage.submit()', ng-disabled: 'isDisabled', translate: 'profile_page.update_profile'}
+      .profile-page__update-account
+        %button.btn.btn-primary.profile-page__update-button{ng-click: 'profilePage.submit()', ng-disabled: 'isDisabled', translate: 'profile_page.update_profile'}
+
+      .profile-page__change-password
+        %a{href: '', ng-click: 'profilePage.changePassword()', translate: 'profile_page.change_password_link'}
+    .profile-page-card{ng-if: 'profilePage.canDeactivateUser()'}
```
gdpelican: TODO: The designs have some other crazy modal pop up letting you know why you can't deactivate, but I'm just hiding the card for now.

app/controllers/api/restful_controller.rb
robguthrie: nice!

app/controllers/api/users_controller.rb
```diff
@@ -1,17 +1,31 @@
 class API::UsersController < API::RestfulController
```
robguthrie: Do you think this might actually be the "CurrentUser" controller? I see Users controller coming into play when we want to show user profiles of other users

app/controllers/popolo/motions_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Popolo::MotionsController < API::RestfulController
```
robguthrie: What about if this was PopoloAPI? (sorry, please ignore me if you're over this)

app/services/user_service.rb
robguthrie: Yea I understand..  are there some specs for this?

config/locales/client.en.yml
```diff
@@ -405,7 +405,23 @@ en:
 
   profile_page:
     messages:
-      updated: 'Profile updated'
+      updated: 'Your information has been updated'
```
robguthrie: Your profile has been updated? (more concise that information)

app/services/user_service.rb
gdpelican: Yeah, I have change password covered with

- It doesn't work when the current password isn't provided correctly
- It doesn't work when the password doesn't match the password confirmation
- It works when both of those things are correct

robguthrie:
Thanks James. Looks great. I like accessible_records

Are there index actions where we need to put require_authenticated_user back?
 
Are you planning on writing e2e tests for this?

Also I think we should consider removing/skipping the "why are you deactivating" stuff... as that is really best handled by intercom. @rdbartlett - any thoughts on that?

I like the simple pattern of submit(), isDisabled on form scopes.

app/services/user_service.rb
robguthrie: oh are they the controller tests rather than service specs?

gdpelican: Yeah; I've been having a good time recently with APIs using controller specs as defacto e2es... I test drive pretty much all of them out since it's usually pretty obvious what a new endpoint is supposed to do before it gets written.

app/controllers/api/users_controller.rb
gdpelican: Yeah I think that's a good call; do you mind if we make that change when we come to it though? It strikes me that other user profiles is a bit down the road from now (considering we don't really do it justice in beta)

gdpelican:
There's a looming thing here around displaying validation errors (specifically on the change password form, but certainly the group / discussion forms could use them as well.); I've confirmed that the API is returning errors correctly, but alongside the modal work will be a solution for displaying validation errors on the form.

gdpelican:
A quick survey of routes.rb makes me think that notifications, invitables, and memberships could possibly use a signed-in requirement for index, although it'd also be possible to just say that public_records for those things == []

gdpelican:
e2e tests = yes, one each for change picture, change password, and deactivate account.

app/controllers/api/users_controller.rb
robguthrie: I reckon it makes more sense to change it now, before it's merged into master.. it's only gonna be harder later on isn't it?

lineman/app/components/change_password_form/change_password_form.coffee
```diff
@@ -0,0 +1,17 @@
...
+
+    $scope.submit = ->
+      $scope.isDisabled = true
+      Records.users.changePassword($scope.user).then ->
+        $scope.isDisabled = false
+        FlashService.success('profile_page.messages.password_changed')
+        $scope.$close()
+      , (response) ->
+        $scope.isDisabled = false
+        if response.status == 422
```
gdpelican: I'm liking this pattern quite a bit - if the error status is unprocessable entity, add validation errors, otherwise broadcast it to the app.

This means we can universally handle 500 (something went wrong), 400 (bad request), or 404 (connection dropped?) in the same place.

lineman/app/components/change_picture_form/change_picture_form.coffee
```diff
@@ -0,0 +1,19 @@
...
+  templateUrl: 'generated/components/change_picture_form/change_picture_form.html'
+  controller: ($scope, $timeout, CurrentUser, Records, FlashService) ->
+    $scope.user = CurrentUser.clone()
+
+    $scope.submit = (kind) ->
+      $scope.user.avatarKind = kind
+      Records.users.updateProfile($scope.user).then changePictureSuccess
+
+    $scope.selectFile = ->
+      $timeout -> angular.element('.change-picture-form__file-input').click()
```
gdpelican: We define this behaviour in the link function of the attachment_form directive (the only other place we do file uploads at the moment), but I couldn't figure out how to add a link function to a modal. This is mildly hacky, but works fine.

lineman/app/components/profile_page/profile_page_controller.coffee
```diff
@@ -1,17 +1,35 @@
-angular.module('loomioApp').controller 'ProfilePageController', ($rootScope, CurrentUser, Records, FlashService, $location) ->
-  @user = CurrentUser.clone()
+angular.module('loomioApp').controller 'ProfilePageController', ($rootScope, Records, FlashService, $location, AbilityService, ModalService, ChangePictureForm, ChangePasswordForm, DeactivateUserForm) ->
+  @init = ->
+    @user = Records.users.find(window.Loomio.currentUserId)
```
gdpelican: I've had to refind the current user after profile page update to get the profile picture to rebind correctly. 

I suspect it's something with the new angular router, but then again I blame the hole in the ozone layer over New Zealand on the new angular router too.

lineman/app/components/validation_errors/validation_errors.haml
```diff
@@ -0,0 +1,3 @@
+.lmo-validation-error
```
gdpelican: This is a nifty little directive which takes a subject ($scope.user) and a field (name), and displays validation errors for that field once we get a 422 back. Wee!

lineman/app/models/base_model.coffee
```diff
@@ -27,7 +27,10 @@ angular.module('loomioApp').factory 'BaseModel', ->
       @updateFromJSON(data)
 
     clone: ->
-      attrs = @baseSerialize()[@constructor.singular] or {}
+      attrs = _.reduce @constructor.attributeNames, (clone, attr) =>
```
gdpelican: We already store the attribute names we import via the constructor, and using that as the base of things to clone works much better than attempting to serialize.

gdpelican: Also this use of reduce is fancy schmancy and I'm happy to consider rewriting it to be a little friendlier.

gdpelican:
Okay, this has turned into a bit of a beast, sorry about that.

There's lots of good stuff in here, though, happy to hear your thoughts @robguthrie @HSalmon

lineman/app/components/change_password_form/change_password_form.coffee
robguthrie: Yea I think it's pretty simple. I like the submit, isDisabled, close simplicity.

I think setErrors should be done within the records_interface function with chained promises. (like save() on Model)

With all the network calls moving to the record_interface I'm thinking we should consider moving base_model#save to the records_interface then we wouldn't need restful_client in the model at all. Then we separate the network stuff from the local stuff (find, where) completely?

lineman/app/models/base_model.coffee
robguthrie: seems good to me!

lineman/app/services/ability_service.coffee
```diff
@@ -66,5 +66,9 @@ angular.module('loomioApp').factory 'AbilityService', (CurrentUser) ->
       (!membership.admin or membership.group().adminIds().length > 1) and
       (membership.user() == CurrentUser or @canAdministerGroup(membership.group()))
 
+    canDeactivateUser: ->
+     _.all CurrentUser.memberships(), (membership) ->
+       !membership.admin or membership.group().adminIds().length > 1
```
robguthrie: aren't you going to need to serialize admin_ids or fetch admins for the group for this check to work? maybe we should serialize a property 'moreThanOneAdmin'

robguthrie:
nice work james!

lineman/app/components/change_password_form/change_password_form.coffee
gdpelican: Yeah I've been thinking there's a lot of duplication between #save and the
records interface stuff; I'd be cool with moving all the network calls to
the record interface.

lineman/app/services/ability_service.coffee
gdpelican: Yeah, that'd be much better!

gdpelican:
Yay this passed!

Thoughts on merging this in for demo today, @robguthrie? I guess the most outstanding thing would be renaming the UsersController to CurrentUserController (I had a real quick go at it and got frustrated because our restful_client assumes that the UserModel posts to the `/api/v1/users` route)

config/locales/client.en.yml
```diff
@@ -424,7 +424,31 @@ en:
 
   profile_page:
     messages:
-      updated: 'Profile updated'
+      updated: 'Your profile has been updated'
+      picture_changed: 'Your profile picture has been updated'
+      password_changed: 'Your password has been updated'
+      deactivated: 'Successfully deactivated your account'
+    change_picture:
```
robguthrie: James the rule I've been following and enjoying is that the translations follow the same naming as BEM. Which reduces nesting. So this would be change_picture_form.title etc. Following the same rule everywhere is going to make it really easy to break the translations up into individual files.. just like the css you'll find change_picutre_form keys in the change_picture_form.yml file.

Can you please change these to match the name of the block they belong to?

robguthrie:
Yea I'm ok with merging today but we need cards to make RestfulController assumptions configurable, and move UsersController to CurrentUserController. 

I really want to be following the namespacing rules in translations and css, so it would be great if you could change those. Sorry to police them so hard but I think it's important so everything is super predictable.

I've stated a real basic code review checklist here: https://github.com/loomio/loomio/wiki/Code-Review-Checklist

gdpelican:
Sorry, not sure what 'make RestfulController assumptions configurable' means?

lineman/app/components/profile_page/profile_page_controller.coffee
robguthrie: Random thought - Maybe it's a problem with correctly updating the indexes in lokijs (I'm pretty sure we're not calling update() on loki when we update a record, which tells it to reindex). Are there any loki views involved?

lineman/app/components/validation_errors/validation_errors.haml
robguthrie: Lovely!

lineman/app/models/user_records_interface.coffee
```diff
@@ -3,4 +3,13 @@ angular.module('loomioApp').factory 'UserRecordsInterface', (BaseRecordsInterfac
     model: UserModel
 
     updateProfile: (user) ->
-      @restfulClient.post 'update_profile', user.serialize()
\ No newline at end of file
+      @restfulClient.post 'update_profile', user.serialize()
```
robguthrie: I really like how simple this is

robguthrie: It's starting to look like it should subclass the restfulClient to me. What do you think?

spec/controllers/api/users_controller_spec.rb
```diff
@@ -30,4 +37,78 @@
...
+      it 'does not allow a change if passord confirmation doesnt match' do
+        old_password = user.encrypted_password
+        post :change_password, user: { password: 'new_password', password_confirmation: 'errwhoops'}, format: :json
+        expect(response).to_not be_success
+        expect(user.reload.encrypted_password).to eq old_password
+      end
+    end
+  end
+
+  describe 'deactivate' do
```
robguthrie: These API tests sure are great. It's nice that they're also a great place to confirm emails have been sent. (I realise we're not sending any in this case...)

I say that because in cucumber/rails land, we test for email sent in the cukes.. But I think it's going to be more fun for us to just test the UI with protractor, and test emails at this level.

robguthrie:
I'm fine for you to merge this whenever you like, @gdpelican it's great and we can improve any of the highlighted issues after merge, but I think feedback is most imporant thing right now.

However, are you in agreement about naming of translation keys follows BEM block name thing I'm asking you to follow?

robguthrie:
yea ignore that please. I was confused.

gdpelican:
Yep, I'm with you, but maybe I need a little guide for the guidelines you'd like me to follow, as I feel like that's been a stumbling block between us for a while.

I'm happy to follow a pattern, but I guess since it feels a little unintuitive / unclear what's being asked, I've just been following what's there already, and what's there already isn't always matching what you're looking for.

robguthrie:
Oh absolutely.. I think we should try to keep all that knowledge in the
Code Review Checklist for now.  Maybe it should move to CONTRIBUTING?

gdpelican:
Yeah, that sounds like a plan.

PR #2286, opened 2015-07-07
Finish thread lintel

HSalmon:


robguthrie:
Nice work.. however I think we should truncate with css rather than angular js.. see something like: http://mattsnider.com/css-string-truncation-with-ellipsis/#heading4

HSalmon:
screen shot 2015-07-07 at 4 57 04 pm

robguthrie:
this is merged into the membership requests branch now eh?

PR #2289, opened 2015-07-08
Truncate thread & proposal titles on thread preview

HSalmon:


lineman/app/components/thread_preview_collection/thread_preview_collection.scss
```diff
@@ -18,6 +18,7 @@
 .thread-preview__thread-header {
   @include fontCrapName;
   padding-bottom: 4px;
+  display: flex;
```
gdpelican: Could you remind me what this does @HSalmon? I know flexbox rocks but don't quite recall the specifics of how it works.

robguthrie: it's magic @gdpelican. see https://css-tricks.com/snippets/css/a-guide-to-flexbox/

the page is long, but just enabling flex on a parent means you can have children in a row that sit next to each other nicely

PR #2290, opened 2015-07-09
Add functionality for joining groups

HSalmon:


app/controllers/api/memberships_controller.rb
```diff
@@ -1,14 +1,12 @@
 class API::MembershipsController < API::RestfulController
 
-  def create
-    params[:group_id] = params[:membership][:group_id]
-    load_and_authorize :group
-    event = MembershipService.join_group group: @group, user: current_user
+  def join_group
+    @group = Group.find(params[:group_id])
```
gdpelican: We're running into that weird thing again where we load outside the service, then authorize inside it.

I think we talked about making the services smart enough to take an id, but I'm not sure if that feels too great either.

```
class MembershipService
  def join_group(group_id: nil, group: nil, actor:, params:)
    group ||= ModelLocator.find group_id
    actor.ability.authorize! :join_group, group
    # etc..
  end
end
```

I dunno if that feels any better, actually.

At the very least I think I'd prefer using ModelLocator here so you can pass an id or a key.
```
@group = ModelLocator.new(:group, params).locate
```

app/services/membership_service.rb
```diff
@@ -15,11 +15,11 @@ def self.remove_admin(membership:, actor:)
...
-  def self.join_group(user: nil, group: nil)
-    user.ability.authorize! :join, group
-    membership = group.add_member!(user)
-    Events::UserJoinedGroup.publish!(membership)
-  end
+  def self.join_group(actor: nil, group: nil)
+     actor.ability.authorize! :join, group
+     membership = group.add_member!(actor)
+     Events::UserJoinedGroup.publish!(membership)
+   end
```
gdpelican: you outta line! :racehorse: 

(whitespace issues)

lineman/app/components/group_page/group_page.haml
```diff
@@ -18,10 +18,7 @@
               %group_actions_dropdown{group: 'groupPage.group'}
             .lmo-btn-wrapper
               %notification_volume_dropdown{translate-root: 'group_page', group: 'groupPage.group'}
-          .group-page__nonmember-actions{ng-if: '!groupPage.isMember()'}
-            %a.btn.btn-success.btn-block.lmo-btn-icon{href: '', ng-click: 'groupPage.joinGroup()'}
-              %i.fa.fa-lg.fa-plus{aria-hidden: 'true'}>
-              %span{translate: 'group_page.join_group' }
+          %join_group_button{group: 'groupPage.group'}
```
gdpelican: I've been having a good time putting ng-ifs on the directives themselves, so we don't even need to load em up if they're not relevant

```
%join_group_button{group: 'groupPage.group', ng-if: 'canJoinGroup()'}
```

lineman/app/components/group_page/join_group_button/join_group_button.coffee
```diff
@@ -0,0 +1,23 @@
+angular.module('loomioApp').directive 'joinGroupButton', ->
+  scope: {group: '='}
+  restrict: 'E'
+  templateUrl: 'generated/components/group_page/join_group_button/join_group_button.html'
+  replace: true
+  controller: ($scope, AbilityService, ModalService, CurrentUser, Records, FlashService, MembershipRequestForm) ->
+    Records.membershipRequests.fetchMyPendingByGroup($scope.group.key)
+
+    $scope.isMember = ->
+      CurrentUser.membershipFor($scope.group)?
```
gdpelican: Do we need isMember() and canJoinGroup()? Seems like canJoinGroup could cover both.

lineman/app/components/group_page/group_page.haml
gdpelican: Ah scratch that, I see the join group button is a little more complicated than a simple on/off.

Did we consider doing two separate directives for join group and request membership?

lineman/app/components/group_page/membership_request_form/membership_request_form.haml
```diff
@@ -0,0 +1,24 @@
...
+    %button{type: 'button', class: 'close', ng-click: '$close()'}
+      %span{aria-hidden: true} &times;
+      %span.sr-only{translate: 'common.action.cancel'}
+    %h1.lmo-h1{translate: 'membership_request_form.heading'}
+
+  .modal-body
+    %fieldset
+      .form-group
+        %label{for: 'membership-request-name', translate: 'membership_request_form.name_label'}
+        %input.membership-request-form__name#membership-request-name.form-control{ng-model: 'membershipRequest.actor().name', ng-required: 'true', ng-disabled: 'true'}
```
gdpelican: Ooh, you could give a hoon to the new %validation_errors directive once that gets merged in here.

lineman/app/models/membership_request_model.coffee
```diff
@@ -6,11 +6,12 @@ angular.module('loomioApp').factory 'MembershipRequestModel', (BaseModel) ->
 
     initialize: (data) ->
       @baseInitialize(data)
-      @fakeUser =
-        name: @name
-        email: @email
-        avatarKind: 'initials'
-        avatarInitials: _.map(@name.split(' '), (t) -> t[0]).join('')
+      if !@byExistingUser()
```
gdpelican: Hm, I don't quite understand this. What is it used for again?

app/controllers/api/membership_requests_controller.rb
```diff
@@ -7,6 +7,12 @@ def pending
     respond_with_collection
   end
 
+  def my_pending
```
gdpelican: I'm not in love with these new pending methods; we should be using visible_records to define this.

```
def pending
  instantiate_collection { |collection| collection.pending }
  respond_with_resource
end

def previous
  instantiate_collection { |collection| collection.responded_to }
  respond_with_collection
end

private

def visible_records
  load_and_authorize :group
  @group.membership_requests
end
```

lineman/app/components/group_page/group_page.haml
robguthrie: I think I strongly prefer that the directive itself decides if it should display. It means you don't need "should this directive run" logic on the parent directive. Thus more self contained. It also means that as a user of the directive you don't need to know anything except what arguments it takes.

I think the reason ng-if ends up outside the directive is because it seems to break a directive if it is on the first element... thus the `.blank`.

I'm open to these being two buttons though.

lineman/app/components/group_page/join_group_button/join_group_button.coffee
robguthrie: I like how that reads too.. However canJoinGroup is something the join_group_button will need to know.. 

````
%join_group_button
%request_membership_button
````

Can feel a wee bit loopy but is also eaiser to use. The pattern of nf-if on the inside becomes more necessary when we look at the memebership_requests_card. It needs to fire a request to see if there are requests before you know if it should display or not. If you have ng-if on the outside, then you need to know about requesting membership_requests on the group_page_controller... yuck

File: lineman/app/models/membership_request_model.coffee (hunk 6 above)
Comment (robguthrie) [inline lineman/app/models/membership_request_model.coffee]: it's not super pretty.. but it's how the model on the server is.

MembershipRequests are either by a real user or by a logged out user, which we track as name and email.

We need a better LoggedOutUser abstractiction that behaves and is used the same as a User

File: lineman/app/services/ability_service.coffee [hunk 8]
```diff
@@ -68,3 +68,19 @@ angular.module('loomioApp').factory 'AbilityService', (CurrentUser) ->
...
+      group.visibleToPublic() or
+      CurrentUser.isMemberOf(group) or
+      (group.visibleToOrganisation() and CurrentUser.isMemberOf(group.parent()))
+
+    canJoinGroup: (group) ->
+      (group.membershipGrantedUpon == 'request') and
+      @canViewGroup(group) and
+      !CurrentUser.isMemberOf(group)
+
+    canRequestMembership: (group) ->
```
robguthrie: I think we need to catch up and write some specs for this file. Maybe even consider breaking it up into sections? @gdpelican

app/controllers/api/memberships_controller.rb
robguthrie: Oh yea.. I think I'm into this. Nice.

app/controllers/api/membership_requests_controller.rb
robguthrie: cool. I'm down!

lineman/app/services/ability_service.coffee
gdpelican: Yarr, abilities are a good thing to be testing.
I wonder how many times we'd need to be including multiple ability dependencies within a controller if we broke it up? I'm thinking something like

```
.controller($scope, GroupAbility, DiscussionAbility, MembershipAbility) ->
```

would make me sad.

robguthrie: agreed.

PR #2291, opened 2015-07-09
stop tracking users!

fauno:
this pr removes tracking users ip addresses and login times.  i realize you may not want to merge this, but i'm just leaving it here for other forks to use :)

also if you're using nginx+passenger, you can protect users with this [nginx.conf snippet](https://github.com/fauno/nginx-config/blob/develop/nginx.conf#L40)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,7 +12,7 @@ class User < ActiveRecord::Base
   SMALL_IMAGE = 25
   MAX_AVATAR_IMAGE_SIZE_CONST = 10.megabytes
 
-  devise :database_authenticatable, :recoverable, :registerable, :rememberable, :trackable, :omniauthable
+  devise :database_authenticatable, :recoverable, :registerable, :rememberable, :omniauthable
   attr_accessor :honeypot
 
   validates :email, presence: true, uniqueness: true, email: true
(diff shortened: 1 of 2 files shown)


gdpelican:
Hm... I pretty much agree that we don't really _need_ to be tracking ip address, but I think sign_in_count and last/current_sign_in_at are pretty low-damage tracking stats which are a key part of our analytics. Thoughts @robguthrie @rdbartlett?

rdbartlett:
We've used the ip tracking to block spammers. So yeah I think we need that
until there's an alternative solution for the spam problem.

gdpelican:
Yeah fair enough.

fauno:
"Richard D. Bartlett" <notifications@github.com> writes:

> We've used the ip tracking to block spammers. So yeah I think we need that
> until there's an alternative solution for the spam problem.

are those spammers automated bots or just trolls?

-- 
http://endefensadelsl.org

rdbartlett:
I haven't looked into it really closely, but I think they are bots. Rob
will have a better idea.

gdpelican:
Going to close this for now, but thanks for the heads-up around this, @fauno. Going forward we'll be sure to be very transparent about what information we're tracking and why, and to provide some instructions for other instance owners around how to stop tracking that information if they want to.

PR #2292, opened 2015-07-10
try moving to new travis infrastructure

robguthrie:


gdpelican:
This looks great. Can we merge? Be super nice to feel confident in our builds not erroring again,

robguthrie:
yes!

PR #2293, opened 2015-07-10
Unify custom routes service strategy

gdpelican:


robguthrie:
Two things come up for me looking at this.. I feel like methods should only be able to be called in one way. I'm wondering what alternatives there are here, like a common way of wrapping a service call so the model is located first.

Also changing services from returning the event to returning the model is going to cause a lot of stuff to break isn't it? My design with these services was you call them and pass the model, and they return an event. You don't need the model returned because you passed it in.

gdpelican:
Yeah, good points all.

I did this and then had a look at converting the existing restful routes
(create, update) to a
service.create(actor: actor, params: params) type syntax, and that didn't
feel too great, especially since it made me deal with permitted params
within the service (ick).

So.. I dunno. I feel like we're really close to a great solution here, but
I'm not certain how to take that final step.

robguthrie:
I think ModelLocator is probably doing more than just locating. It seems like its Loading and Initialising. Particularly in the cases of service.update.

If the controller had a common way to initialise the model before the service call that would be great. And it would leave the service behaviours untouched.

There is lots of great stuff in this pull request, I hope we can keep it.

gdpelican:
I think I'm going to chalk this up to an exploration; the minor inconvenience of loading and authorizing in different places isn't worth an extended effort to refactor this right now, I think. (What we have is pretty durn good already, I think)

There's a white :whale: out there of a more better restful strategy tho.

PR #2296, opened 2015-07-11
Versioning

gdpelican:
I've been thinking quite a bit about versioning recently, so I had a quick go at writing a script that'll 

Its based off of discourse's solution here:
https://github.com/discourse/discourse/blob/master/script/version_bump.rb
https://github.com/discourse/discourse/blob/master/lib/version.rb

And basically allows us to, with a single command, update the version in the app (and, soon, on github as well.)

The idea being that before (or even as) you deploy, you can tag what kind of deploy it is, update the repo and the github tag, and then push.

Usage is generally like this (current version assumed 0.10.1): 
```
ruby script/bump_version patch # => 0.10.2
ruby script/bump_version minor # => 0.11.0
ruby script/bump_version major # => 1.0.0
ruby script/bump_version rc1 # => 0.10.1.rc1
```

The script would perform one of three actions after writing the new version:

- no-commit: don't make a commit of the changes, just modify files
- commit: make a commit of the changes (and push to github and tag it?)
- deploy: make a commit of the changes and deploy them (run ./deploy.sh)

WIP, not for merging yet.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,17 +1,22 @@
 #!/bin/bash
 
-if [ -z ${1+x} ]; then SOURCE_BRANCH='master'; else SOURCE_BRANCH=$1; fi
+if [ -z ${1+x} ]; then DEPLOY_TYPE='patch'; else DEPLOY_TYPE=$1; fi
+# Do we need the ability to push other branches to production? Seems _really_ dangerous to me
+# if [ -z ${1+x} ]; then SOURCE_BRANCH='master'; else SOURCE_BRANCH=$1; fi
 
 echo "building angular assets and deploying branch:"
 echo $BRANCH
 
 BUILD_BRANCH=production-$(date +%Y%m%d%H%M%S)
-git checkout $SOURCE_BRANCH
+git checkout master
+ruby script/bump_version $DEPLOY_TYPE commit
+git push origin master
+
 git checkout -b $BUILD_BRANCH
 cd lineman && npm install && bower install && lineman build && cd ..
 cp -R lineman/dist/* public/
 cp lineman/vendor/bower_components/airbrake-js/airbrake-shim.js public/js/airbrake-shim.js
-git add public/img public/css public/js public/fonts && git commit -m "production build commit" && git checkout $SOURCE_BRANCH && git push loomio-production $BUILD_BRANCH:master -f
+git add public/img public/css public/js public/fonts && git commit -m "production build commit" && git checkout master && git push loomio-production $BUILD_BRANCH:master -f
 git branch -D $BUILD_BRANCH
 heroku run rake db:migrate -a loomio-production
 heroku restart -a loomio-production
--- a/lib/version.rb
+++ b/lib/version.rb
@@ -1,0 +1,7 @@
+module Loomio
+  module Version
+    def self.current
+      [MAJOR, MINOR, PATCH, PRE].reject(&:nil?).join('.')
+    end
+  end
+end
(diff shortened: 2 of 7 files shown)


gdpelican:
Kay, this is ready for a look, @robguthrie 

The kernel:

- The current version is accessible by calling `Loomio::Version.current`
- You can run `ruby script/bump_version` at any time to change the version.
- You can run `ruby script/bump_version patch|minor|major commit` to update the version and commit to git
- You can run `./deploy.sh` and it will automatically bump the patch number, commit it, and push the update to master
- You can run `./deploy.sh patch|minor|major` to change what kind of version commit it makes.

deploy.sh
```diff
@@ -1,17 +1,22 @@
 #!/bin/bash
 
-if [ -z ${1+x} ]; then SOURCE_BRANCH='master'; else SOURCE_BRANCH=$1; fi
+if [ -z ${1+x} ]; then DEPLOY_TYPE='patch'; else DEPLOY_TYPE=$1; fi
+# Do we need the ability to push other branches to production? Seems _really_ dangerous to me
```
robguthrie: I think how dangerous it is depends on the branch. But yea, I have found it useful in the past to be able to do that.

robguthrie:
Actually.. could we please keep deploying and version bumping separate for now?

I don't feel good about the way these two things are mixed together right now. I think there is room for improvement. I look forward to where you're trying to go but I think "cut new version from master" should not be the same as deploy code to a place.

What I'm saying is I'd rather if deploy was left alone and there was a new command to "do it all"

robguthrie:
Actually.. use case: I currently want to deploy a non master branch to loomio-clone. I'm thinking I'd like to modify the deploy script to support changing git remote and branch.

And this is a case where I don't want to be worried I'll be bumping version. I think...

gdpelican:
My two cents is that I don't want us pushing to loomio production without changing a version (this way we can display it in the app, include it with bug reports / UI feedback, errbit errors, etc.) Looking forward to when (more?) people have instances installed, we want to be able to respond to bug reports with something like "Your site is running 1.1.10, this is fixed in 1.1.12; please run bash ./update and your problem will go away <3". I feel pretty strongly that the version bump and the production push want to be linked, and that they should happen without friction at the same time, without us having to remember to do it, otherwise this will end up out of sync and be not useful.

I'm happy to put in a --no-version-bump option or detect whether we're pushing to loomio.org and only bump the version then, but I'm fairly certain the sensible default is to bump the version on a deploy to loomio-production.

I'm thinking something like
```
./deploy.sh # deploy to production, bump patch version
./deploy.sh -v minor # deploy to loomio.org, bump minor version
./deploy.sh -e loomio-clone # deploy to loomio clone, no version bump
```

robguthrie:
If we could also select the branch to deploy again then I'm happy for this to go ahead. Thanks for this work James.

robguthrie:
Oh I'm now completely with you on there being a really simple deploy to master and bump command. If there could also be a really simple deploy any remote and branch usage that would be perfect.

I'm keen to merge this as soon as you're happy (even if only to add the deploy alt remote/branch stuff myself aftwards)

robguthrie:
I've spent the morning rewriting the bash deploy script as a rake task. It's a lot easy to use ruby than bash. I think this version stuff and the new build/deploy thing will work nicely with each other.

gdpelican:
Yeah I spent a few minutes at some point trying to write `contains?` in
bash and almost tore my own head off.

robguthrie:
@gdpelican can I take this on and merge into my work?

gdpelican:
Sure

robguthrie:
included in #2306

PR #2297, opened 2015-07-12
change group_url, discussion_url etc to not use group subdomains

robguthrie:
Also update the readme. Not finished.. working on redirecting old urls and pending a bit more group discussion.


app/helpers/readable_unguessable_urls_helper.rb
```diff
@@ -3,99 +3,28 @@ module ReadableUnguessableUrlsHelper
...
+                       .merge(route_hash(instance, model) )
 
       url_for(options)
     })
 
     define_method("#{model}_path", ->(instance, options={}) {
       options = options.merge(only_path: true)
+                       .merge(route_hash(instance, model) )
 
-      self.send("#{model}_url", instance, options)
```
gdpelican: Whoa, crazy the amount of logic here around around subdomains.

app/helpers/locales_helper.rb
```diff
@@ -105,8 +100,8 @@ def first_selectable_locale(locales)
   end
 
   def filter_locales(input_locales, valid_locales)
-    input_locales = Array(input_locales).map &:to_s
-    valid_locales = Array(valid_locales).map &:to_s
+    input_locales = Array(input_locales).map &:to_sym
+    valid_locales = Array(valid_locales).map &:to_sym
```
gdpelican: I'm pretty sure this method is just 
```
(Array(input_locales) & Array(valid_locales)).map(&:to_sym)
```

app/controllers/redirect_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    redirect_to_resource_or_dashboard
+  end
+
+  def discussion_id
+    @resource = Discussion.where(id: params[:id]).where('id < 11500').first
+    redirect_to_resource_or_dashboard
+  end
+
+  def motion_id
+    @resource = Motion.where(id: params[:id]).where('id < 7300').first
```
gdpelican: I don't understand these magic numbers? Why are group and discussion the same thing?

robguthrie: They're approximate, carried from what they were when Mix wrote them. They're to support old urls but prevent new sequential urls.

PR #2298, opened 2015-07-12
Fix ordering on groups page and group page

gdpelican:


lineman/app/components/group_page/discussions_card/discussions_card.haml
```diff
@@ -12,7 +12,7 @@
       .group-thread-list-empty-access{ng-if: 'howToGainAccess()'}
         %span{translate: 'group_page.thread_list.empty.{{howToGainAccess()}}'}
 
-    .group-thread-list-item{ng-repeat: 'discussion in group.organisationDiscussions() track by discussion.id | orderBy: ["-lastActivityAt"] | limitTo:loaded'}
+    .group-thread-list-item{ng-repeat: 'discussion in group.organisationDiscussions() | orderBy: "-lastActivityAt" track by discussion.id | limitTo:loaded'}
```
robguthrie: the position of track by boggles me. Is it also ok at the end of the expression?

robguthrie:
thank you James.

gdpelican:
Yeah it's super weird. I don't 100% get what's going on under the hood, but
it seems like when you call 'track by', it takes an object: ({ id: 1, name:
'name' }), and transforms it to just the track by expression (1), so you
can't actually order things after that happens.

The thing that would make sense to me would be
```
discussion in discussions | orderBy:'name' | trackBy:'id' | limitTo:5
```

But of course that's not a thing. ¯\_(ツ)_/¯

PR #2301, opened 2015-07-13
Tweak/tidy nav bar css

AaronThornton00:
Here are some changes to pretty up the nav bar
(pretty subtle, but as is just about every item has a different height)

before:
screenshot 2015-07-13 18 19 25

after:
screenshot 2015-07-13 18 18 46

- Add left and right padding to nav-bar to be consistent with icon spacing
- Reduce size of group and bell icon so they don't spill out of their button and are visually consistent with other icons.
- Reduce padding on buttons to match left and right nav-bar padding
- Move search to it's own floating block as it breaks the right icon div on Firefox
- Remove margin form search input as it flows out of its container, reduce height and adjust icon
- All floating containers a standard 39px high
- All buttons a standard 33px high

### DIFF
--- a/lineman/app/components/navbar/navbar.scss
+++ b/lineman/app/components/navbar/navbar.scss
@@ -3,17 +3,29 @@
   background: white;
   height: $navbar-height;
   border-bottom: 1px solid $border-color;
-  padding: 2px 0px;
+  padding: 2px 5px;
   width: 100%;
   position: fixed;
 }
 
 .lmo-navbar__left{
   float: left;
+    padding: 3px 2px;
+  .fa-group {
+    font-size: 1.1em;
+  }
 }
 
 .lmo-navbar__right{
   float: right;
+  padding: 3px 2px;
+  .fa-bell-o {
+    font-size: 1.2em;
+  }
+}
+
+.lmo-navbar__right--search {
+    padding: 2px 2px 3px;
 }
 
 .lmo-navbar__center{
@@ -22,8 +34,13 @@
 
 .lmo-navbar__item{
   display: inline-block;
-  margin: 0 0 0 2px;
-  height: $navbar-height - 4px;
+  margin: 0;
+}
+
+.lmo-navbar__item {
+  .navbar-user-options .lmo-navbar__btn {
+    padding: 4px 5px;
+  }
 }
 
 .lmo-navbar__item--selected {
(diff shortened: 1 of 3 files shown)


AaronThornton00:
Anyone tell me why my .npmignore is'nt stoping the node modules?

gdpelican:
I think our gitignore is ignoring `loomio/lineman/node_modules`, but you appear to have a folder at `loomio/node_modules`

Try deleting that one, and running `npm install && bower install` in `loomio/lineman`, and see if that fixes it?

gdpelican:
Another possibility is that you've just done `npm install` with some of the basic packages, instead of installing them globally (`npm install -g protractor bower etc`)

AaronThornton00:
@gdpelican thanks man.  Good spotting.  A residual left over from my frantic attempts to get lineman working. Just needed to remove the dir.

lineman/app/components/navbar/navbar.scss
```diff
@@ -22,8 +33,10 @@
 
 .lmo-navbar__item{
   display: inline-block;
-  margin: 0 0 0 2px;
-  height: $navbar-height - 4px;
+  margin: 0;
+  .navbar-user-options .lmo-navbar__btn {
```
robguthrie: Please don't use unless you have to. I think it's ok for a .fa-bell type thing, but not for other BEM class names.

I'm trying to enforce that our CSS follows the BEM style. (Have a read of http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/ if you like)

In this case just put the declaration directly under this one, rather than nested within it.

robguthrie: Maybe also have a look at https://css-tricks.com/bem-101/ if you like. Let me know if either of those seem better than the other. I want to include a good link in the Code Review Checklist/Style guide

robguthrie:
THAT ASIDE.... It's so incredibly awesome to have you making a pull request again old mate! Thank you so so much!

AaronThornton00:
Nice Rob, BEM makes a lot of sense and that first doc is super simple to understand.
Its nice to be contributing too :gift_heart:

robguthrie:
Cool. @AaronThornton00 are you able to squash this into a single commit then I'll merge it?

AaronThornton00:
:boot :ant :spar

robguthrie:
Thank you!

PR #2306, opened 2015-07-14
google analtics for angular and beta

robguthrie:
not ready


app/views/layouts/angular.html.haml
```diff
@@ -14,7 +25,7 @@
     %start_menu
 
 
-    - if Rails.env.production?
+    - if Rails.env.production? and ENV['ERRBIT_HOST']
```
gdpelican: Not sure I understand why we're hiding the turn off angular option if Errbit isn't set up?

robguthrie: good point

lib/tasks/loomio.rake
```diff
@@ -1,4 +1,31 @@
 namespace :loomio do
+  task :deploy do
```
robguthrie: @gdpelican simple move of deploy to ruby code. working for deploy to loomio-clone

robguthrie: I'll probably move it to lib/tasks/deploy.rb

PR #2308, opened 2015-07-14
Email settings page

HSalmon:
Hey @robguthrie, still having trouble...maybe you could take a look if you've got a mo? Suspect nothing is working actually...


HSalmon:
@robguthrie, @gdpelican, I think this is ready for review?

config/routes.rb
```diff
@@ -384,6 +385,9 @@
   # this is a dumb thing
   get '/groups', to: 'dashboard#show'
 
+  # just to boot angular
```
gdpelican: Arg, this has been bugging me for a while. I wonder if there's a way for us to _not_ have to define meaningless routes in beta just to allow people to show up on those pages in angular.

(Also could we take this comment out before merge?)

lineman/app/app.coffee
```diff
@@ -47,6 +47,7 @@ angular.module('loomioApp').controller 'AppController', ($scope, $filter, $rootS
     {path: '/inbox', component: 'inboxPage' },
     {path: '/groups', component: 'groupsPage' },
     {path: '/profile', component: 'profilePage'},
+    {path: '/email_settings', component: 'emailSettingsPage' },
```
gdpelican: Hey it feels appropriate to nest this under `/profile` I think, could we do either `profile/settings` or `profile/email_settings`?

lineman/app/components/email_settings_page/email_settings_page_controller.coffee
```diff
@@ -0,0 +1,12 @@
+angular.module('loomioApp').controller 'EmailSettingsPageController', (Records, FlashService, CurrentUser, $location) ->
+  @init = ->
+    @user = CurrentUser
```
gdpelican: Hm, I'm thinking I may have set a poor precedent doing this init() thing on the profile page; I only did that in order to ensure that the profile picture updated correctly. I don't think you need it here, you can just do

```
@user = CurrentUser.clone()
```

(ps I believe we want to `clone()` the CurrentUser here, so the settings aren't updated before the form is submitted.)

lineman/app/app.coffee
robguthrie: I want to keep the routes consistent with existing ones.

lineman/app/components/email_settings_page/email_settings_page_controller.coffee
robguthrie: cool. I agree that init() feels pretty weird most times in a controller.

lineman/app/css/print_scss_includes.rb
```diff
@@ -3,5 +3,5 @@
   puts "@import '#{filename}';" unless filename =~ /settings/
 end
 Dir[ File.join('../components', '**', '*.scss') ].each do |filename|
-  puts "@import '#{filename}';" unless filename =~ /settings/
```
gdpelican: Hrm. I assume we need this line in order for us to pick up the `email_settings.scss`, but why was it there in the first place? Are there other files which match /settings/ we don't want?

robguthrie: No I've looked and both unless ~settings lines are rubbish. Vestigal css stuff. @HSalmon can you remove the other unless too?

lineman/app/app.coffee
robguthrie: Which, after looking at it.. tells me the route should be /email_preferences as in..
https://www.loomio.org/email_preferences

robguthrie: Hannah if you change your route to /email_preferences (you can just change the route name and leave the component as is) does that mean we don't need to add a route in here?

lineman/app/components/email_settings_page/email_settings_page.scss
```diff
@@ -0,0 +1,17 @@
+.email-settings-page {
+  max-width: $main-max-width;
+  margin: 0px auto;
+}
+
+.email-settings-page__email-settings {
```
robguthrie: I'm kind of surprised we don't have a .lmo-card general purpose card class. Maybe this could be the first one?

gdpelican: :+1:

lineman/app/models/user_records_interface.coffee
```diff
@@ -5,6 +5,8 @@ angular.module('loomioApp').factory 'UserRecordsInterface', (BaseRecordsInterfac
     constructor: (recordStore) ->
       @baseConstructor recordStore
       @profileClient = new RestfulClient 'profile'
```
robguthrie: @gdpelican I think this will fix a bug for you in profile picture update land. (this also helped convince me that the subclassing restfulclient and removing recordinterfaces approach feels right)

lineman/app/models/user_records_interface.coffee
```diff
@@ -5,6 +5,8 @@ angular.module('loomioApp').factory 'UserRecordsInterface', (BaseRecordsInterfac
...
+      @profileClient.onSuccess = (response) =>
```
gdpelican: This is a red flag for me; means we've set up our client / interface relationship a little odd.

lineman/app/models/user_records_interface.coffee
gdpelican: Simulataneous comment high five! :hand:

PR #2310, opened 2015-07-14
Fix inbox not loading as first screen

gdpelican:


lineman/app/components/inbox_page/inbox_page_controller.coffee
```diff
@@ -1,17 +1,23 @@
...
+  @groups = ->
+    CurrentUser.parentGroups()
 
-  @queryFor = (group) -> @["group#{group.id}"]
+  @init = =>
+    _.each @groups(), (group) =>
+      @["group#{group.id}"] = ThreadQueryService.groupQuery(group)
+    @baseQuery = ThreadQueryService.filterQuery('show_unread')
+  @init()
+  $scope.$on 'currentUserMembershipsLoaded', @init
```
gdpelican: Since we kicked off the memberships into a separate request, we have to wait for that to come back before we set up all the group queries.

robguthrie: whhooooop!

PR #2311, opened 2015-07-14
Swap in new icon and colors

AaronThornton00:
...and now for something a little more controversial.
 
Shake baby shake, shake, shake.

This is a pretty big change but I'll tell you my think'n.  I hope without offending anyone:
- the icons were pretty ruff and originality pretty slim
- buttons needed borders for accessibility
- really disliked the flag
- colours on buttons didn't match graph colours
- block and disagree were just shades of the same color
- pastel colour lack impact

Without saying too much more... We may have to have a Loomio to discuss?

Before:
screenshot 2015-07-14 19 52 53

After:
screenshot 2015-07-14 19 32 26

screenshot 2015-07-14 19 43 05

screenshot 2015-07-14 19 44 02


AaronThornton00:
@rdbartlett where else in the app uses icons?

rdbartlett:
I think we need icons for proposal, thread and group. Maybe also for
proposal outcome.

I love the new colours. The icons are an improvement I reckon but I am not
100% sure yet.

The proportions of the thumb to the fist seem a bit off. Have you played
with a longer thumb or a slimmer fist?

AaronThornton00:
Cool Rich, that's awesome.  To be honest the colours are the tricky part. I can make the thumb longer and in fact it will probably differentiate it from abstain some more.  The original icon actually came from the material docs examples.  Personally I think the block and the abstain are the most improved. But lets hear what the rest of the group has to say.

gdpelican:
I'm having a tough time with the purple; it doesn't associate with 'disagree' very well for me, and reserving red for blocks (which are pretty rare), and thus getting rid of the traffic lights we all know and love, seems a bit tough. I might be more inclined to a green / yellow / red / black-or-dark-grey scheme.

I like the new icons, although abstain isn't doing it for me at the moment.

Actually, would you mind posting those screenshots to a Loomio in the contributors group, @AaronThornton00? Like you say, I'd like to get more opinions on this before making any changes.

AaronThornton00:
@gdpelican yep, I wasn't expecting this to just slip in.  Good points about the traffic lights, but lets talk about it on Loomio

rdbartlett:
Yeah traffic lights plus black for block!

AaronThornton00:
Continue comments on this here https://www.loomio.org/d/Osukp0nP?proposal=rlUKRHAx

robguthrie:
Yay our thing.

PR #2312, opened 2015-07-14
Form Service 2: ReFormation.

gdpelican:
Here's a first pass at a general form service:

The basic gist is we have a factory which generates a submit function, which does the following:

- Accepts a scope and a model (required)
- Accepts a method to call as a submit function (optional; model.save is the default)
- Accepts a success flash message (optional)
- Accepts a success callback, for additional stuff to do on success (like redirect to a different page; optional)
- Sets an isDisabled switch on the form once it's submitted, and takes it off once a response comes back
- Closes the modal on success, if it's a modal form
- Sets validation errors on the model for a 422 response
- Broadcasts a page error event for other error responses (bad request, resource not found, internal server error, etc.)

Be interesting to see how this holds up to spreading it through all the forms in the app; I've had a good time putting it on the user profile forms and the group forms so far.

PS: Github, no emojis in PR titles? :-1:

### DIFF
--- a/lineman/app/services/form_service.coffee
+++ b/lineman/app/services/form_service.coffee
@@ -1,0 +1,27 @@
+angular.module('loomioApp').factory 'FormService', ($rootScope, FlashService) ->
+  new class FormService
+
+    errorTypes =
+      400: 'badRequest'
+      401: 'unauthorizedRequest'
+      403: 'forbiddenRequest'
+      404: 'resourceNotFound'
+      422: 'unprocessableEntity'
+      500: 'internalServerError'
+
+    submit: (scope, model, options = {}) ->
+      submitFn = options.submitFn or model.save
+      ->
+        scope.isDisabled = true
+        submitFn(model).then (response) ->
+          scope.isDisabled = false
+          model.setErrors()
+          scope.$close()                                                  if typeof scope.$close == 'function'
+          options.successCallback(response)                               if typeof options.successCallback == 'function'
+          FlashService.success options.flashSuccess, options.flashOptions if options.flashSuccess?
+        , (response) ->
+          scope.isDisabled = false
+          model.setErrors response.data.errors if response.status == 422
+          $rootScope.$broadcast errorTypes[response.status] or 'unknownError',
+            model: model
+            response: response
--- a/app/controllers/development_controller.rb
+++ b/app/controllers/development_controller.rb
@@ -283,7 +283,7 @@ class DevelopmentController < ApplicationController
   def test_subgroup
     unless @test_subgroup
       @test_subgroup = Group.create!(name: 'Johnny Utah',
-                                     parent_id: another_test_group.id, 
+                                     parent_id: another_test_group.id,
                                      visible_to: 'public')
       @test_subgroup.add_admin! patrick
     end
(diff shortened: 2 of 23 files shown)


lineman/app/services/form_service.coffee
```diff
@@ -0,0 +1,27 @@
...
+      403: 'forbiddenRequest'
+      404: 'resourceNotFound'
+      422: 'unprocessableEntity'
+      500: 'internalServerError'
+
+    submit: (scope, model, options = {}) ->
+      submitFn = options.submitFn or model.save
+      ->
+        scope.isDisabled = true
+        submitFn(model).then ->
```
gdpelican: This has a mildly awkward side effect of calling `model.save(model)` by default, which is weird but works fine because save isn't expecting arguments.

Going forward the better solution is to move the `save` function off the model and onto the records interface, which will allow us to just pass the name of a function.

```
FormService.submit $scope, model, 'deactivate'
```
```
submit: (scope, model, action = 'save', options = {}) ->
  ->
    Records[model.constructor.plural][action]().then ->
      # etc
```

This has the added benefit of allowing most forms to operate with a Records dependency, which gets moved to the FormService.

robguthrie: This is great @gdpelican. I have a few suggestions:

I'm still wondering if all the RecordsInterace methods like deactivate(user) that take the model should just be model instance methods that operate on self.

It would also mean that setErrors can always been handled within the model itself (and only once - chained on .then ) and FormService would not need to be doing that work. I really don't think FormService should be responsible for setting up the model errors.

Also I wonder if the scope.$close for modals should be there.. maybe if ModalService should be responsible for setting its close callback, then FormService does not need to know about modal behaviour.

So if the model took responsiblity for setting it's own errors and modals setup a $close() callback themselves.. then i think this would look really neat. How you do feel about those changes?

robguthrie:
Having said that this is a big improvement, those other changes I've suggested might be done in the future and I'm happy to merge it when you are.

gdpelican:
I think both of us have flopped back and forth as to whether the restful client interface class should be the model or the recordsInterface; I think we definitely agree it shouldn't be both... at this particular moment I like the idea of making the models dumber than they are now, rather than smarter, which leads me to think the recordsInterface is where we want to go; it's a bit like the question 'I have to implement a deactivate method in rails land. Do I put it on the User model, or in the UserService?'

But then again chaining the setErrors method so it doesn't have to be part of the public interface of the model would be great, and it's one less thing to need to pass around with the model in order to get it to do stuff.

I'm happy to wait a little while longer to make that call, as I'm still not at all really sure which is better.

Not quite sure I have a head around how the ModalService would set its success callback?

robguthrie:
I agree with all that. What if the RestfulClient thing was initialised with
the model, then you can call methods on top of that?
UserClient.for(user).save() type thing? So long as setErrors and other
clearly model stuff is not on the external part of it I'm happy too.

robguthrie:
Ok I'm feeling super convinced that the RecordsInterface class should die. 

Emerging from the ashes are things that inherit from RestfulClient like UserClient with methods like:
fetch, fetchByKey, destroy

build, import, find, findBy, where, remove, import, belongingTo all move to the Model protoype/class

And we'll expose the Model rather than the RecordsInterface at Records.collectionName. Ie Records.users will be the Model rather than the RI.

robguthrie:
Hopefully I find some time to jam on this in the next couple of days.

gdpelican:
@robguthrie I'd love to get this in first and iterate it with the records interface changes, rather than the other way round. Hoping to work on it a little tonight to get it to mergeable.

robguthrie:
super! yea, merge as soon as you like, it's a nice improvement.

PR #2313, opened 2015-07-14
Fix thread actions spacing

gdpelican:
This style is making our thread view look like crap and, after searching for a reason for it to be there for some amount of time, I couldn't find one. So let's ditch it. :tada:

### DIFF
--- a/app/controllers/development_controller.rb
+++ b/app/controllers/development_controller.rb
@@ -59,6 +59,14 @@ class DevelopmentController < ApplicationController
     redirect_to discussion_url(test_discussion)
   end
 
+  def setup_busy_discussion
+    cleanup_database
+    test_discussion
+    sign_in patrick
+    setup_all_notifications_work
+    redirect_to discussion_url(test_discussion)
+  end
+
   def setup_proposal
     cleanup_database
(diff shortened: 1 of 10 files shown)


rdbartlett:
Actually this is a namespacing error. There are two different uses for .thread-actions. It's on the bottom of comments (which this css breaks) and on the carets on the top right of cards (which this css fixes).

gdpelican:
Can the fix be removing the `thread-actions` class from the events in the thread then?

robguthrie:
I think thread-actions is a back block name for anything on the
thread-items. Can we remove all thread-actions mentions from there?

robguthrie:
back = bad

robguthrie:
I'm having a bit of a tidy up of this code... it's from our sorta BEM days and it's hurting now. Gimmie 20 mins

robguthrie:
Ok I've updated this a bit.. more correct BEM naming.. but not sure how accurate it is:
image

gdpelican:
Whoa looks pretty great to me, although I think the vote icon could still use a little bottom padding to match the spacing on the other event types

PR #2314, opened 2015-07-14
locales update - backslide on 1.0, some new up and comers

mixmix:


mixmix:
@robguthrie  change ready

@HSalmon there's one file in particular worth watching in these diffs. This time is a bit of a mess but generally it will give you a sweet overview of how the coverage is changing
 https://github.com/loomio/loomio/pull/2314/files#diff-ae0f14d6d3594db41c778b042c878a6c

Note we need to add to language_names and native_language_names strings for `:pl` an `:lt`
No real hurry at the moment. But they're worth watching in their rate of translation

robguthrie:
Wowowowowowowow!!

PR #2316, opened 2015-07-16
Snackify flash messages

HSalmon:
@gdpelican, @robguthrie, could probs do with some eyez :eyes:

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -56,6 +56,8 @@ en:
     outcome_label: Proposal outcome
     outcome_placeholder: Let everyone know the outcome of the proposal.
     publish_outcome: Publish outcome
+    messages:
+      created: Outcome published
 
   previous_proposals_card:
     heading: Previous Proposals
@@ -90,9 +92,9 @@ en:
       aria_label: 'Attach a file'
       remove_attachment: 'Remove attachment'
     messages:
-      created: 'Your comment was added to the thread'
-      updated: 'Comment successfully updated'
-      destroyed: 'Your comment has been deleted'
+      created: 'Comment added'
+      updated: 'Comment updated'
+      destroyed: 'Comment deleted'
     edit_comment: 'Edit comment'
     delete_comment: 'Delete comment'
     cancel_reply: 'Cancel Reply'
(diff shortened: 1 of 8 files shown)


config/locales/client.en.yml
```diff
@@ -173,6 +174,8 @@ en:
     select_position: 'Select position'
     statement_placeholder: 'Share your thoughts with the group.'
     aria_statement_label: 'Optionally enter a statement to be recorded with your position'
+    messages:
+      created: Vote added
```
gdpelican: For consistencies sake, can we make a habit of putting quotes around yaml messages?

lineman/app/components/flash/flash.scss
```diff
@@ -1,29 +1,21 @@
...
+  bottom: 20px;
+  left: 20px;
+  min-width: 220px;
+  position: fixed;
+  z-index: 200;
 }
 
 .flash-container {
-  padding: 5px 0;
+  background-color: $primary-text-color;
```
gdpelican: Hmm.. why is the background a text color?

Actually, a screenshot of this would be baller.

gdpelican:
A+++++++ Didn't make :eyes: melt WOULD REVIEW AGAIN A+

lineman/app/components/flash/flash.haml
```diff
@@ -2,5 +2,3 @@
   .flash-container{ng-show: 'display()', ng-animate: "'enter'", class: 'alert-{{flash.level}}'}
     .flash-message
       {{ flash.message | translate:flash.options }}
-    %button.close.dismiss-flash.pull-right{type: 'button', ng-click: 'dismiss()'}
```
gdpelican: Is there some code somewhere in here which makes the message actually go away? I guess I was expecting a javascript timeout or a css transition of some kind.

config/locales/client.en.yml
HSalmon: Do you want me to go through and change that now?

robguthrie: Hrmm.. I'm not super keen on this always quote rule. I like brevity and yml has already thought about this.

gdpelican:
I think it's a thing we need to decide on, because it's half on and half
off at the moment.

I prefer quotes because they always work, whereas no quotes breaks when you
need to interpolate and such.

PR #2318, opened 2015-07-17
Fix avatar

HSalmon:


lineman/app/components/user_avatar/user_avatar.haml
```diff
@@ -1,7 +1,5 @@
-.user-avatar{aria-hidden: 'true', class: '{{size}}'}
-  .avatar.avatar-initials{ng-if: 'user.avatarKind == "initials"', ng-class: '{ coordinator: coordinator }'}
+.user-avatar{aria-hidden: 'true', ng-class: '{"user-avatar--large": large, "user-avatar--small": small, "user-avatar--coordinator": coordinator}'}
```
gdpelican: ng-class is way overkill here imho, especially given that we're going to want to add more sizes and that the directive doesn't need to be dynamic based on size.

```
$scope.size = 'small' unless $scope.size?
```
```
.user-avatar{class: 'user-avatar--{{size}}'}
```

lineman/app/components/user_avatar/user_avatar.scss
```diff
@@ -1,31 +1,50 @@
-.avatar-initials{
-  @include smallIcon;
+// .avatar-initials{
```
gdpelican: Can we remove? Also I don't see a replacement for the user-avatar coordinator class?

lineman/app/components/user_avatar/user_avatar.haml
robguthrie: thank you.. we went through a bunch of css contortions and there are still knots to iron out. This is a good suggestion

HSalmon:
@gdpelican, @robguthrie, is this good to go?

gdpelican:
Hey Hannah, we'll need a smaller avatar ('tiny'?) for the profile picture in the top right of the navbar.
screen shot 2015-07-19 at 8 08 48 pm

gdpelican:
This looks good to me :ship:

PR #2319, opened 2015-07-18
Tweak/optimize navbar spacing for mobile

AaronThornton00:
#### Not ready for merge - broken lintel style ####

I decided to redo the navbar as I wasn't happy with the spacing on my iphone5

This PR optimises the navbar spacing on mobiles,  it now uses percentage widths instead of the floats.
- on mobile it gives more space for fat fingers to change nav filters
- on mobile it spreads search input to fill unused space
- keeps every nav item equally spaced across mobile devises
- fixes search dropdown positioning as it was half of the screen and at a different y-pos
- gives all nav dropdowns a consistent width and position.

before: iphone 6
screenshot 2015-07-18 15 41 04

after: S4
s4

after: iphone 5
iphone 5

after: web
web

### DIFF
--- a/lineman/app/components/thread_page/thread_page.haml
+++ b/lineman/app/components/thread_page/thread_page.haml
@@ -52,3 +52,5 @@
 
     %activity_card{discussion: 'threadPage.discussion'}
     %comment_form{comment: 'threadPage.comment'}
+
+    %thread_lintel
--- a/lineman/app/components/dropdown/dropdown.scss
+++ b/lineman/app/components/dropdown/dropdown.scss
@@ -103,8 +103,11 @@ li .dropdown-menu-item-label{
     position: static;
     .dropdown-menu{
       top: auto;
-      right: 0;
-      left: 0;
+      right: 0px;
+      left: 0px;
+      margin: 2px 8px;
+      min-width: 250px;
+      max-width: none;
     }
   }
 }
--- a/lineman/app/components/navbar/navbar.haml
+++ b/lineman/app/components/navbar/navbar.haml
@@ -1,5 +1,5 @@
 %header.lmo-navbar{role: 'navigation'}
-  .lmo-navbar__left
+  .lmo-navbar-item-container.lmo-navbar-item-container--left
     .lmo-navbar__item{ng-class: '{"lmo-navbar__item--selected": selected == "dashboardPage"}'}
       %a.btn.lmo-navbar__btn.lmo-navbar__btn-icon{href: '/dashboard', title: "{{ 'navbar.recent_title' | translate }}", tabindex: 1, ng-click: 'homePageClicked()'}
         %i.fa.fa-lg.fa-clock-o>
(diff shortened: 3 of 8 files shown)


AaronThornton00:
I'll try and get the psudo speech bubble elements in when I get a chance.
also some dev work as the search dropdown doesn't close when you open another dropdown.

gdpelican:
Sweet!

Hey @AaronThornton00, I'd love to get you plugged into the sprint board to capture the work that you're doing in these sprints, as I think it's super significant and awesome. (I was just using 1.0 earlier today and marvelling at how just a little design polish makes the app feel so much better.)

Maybe @rdbartlett and @robguthrie would be better suited to getting you plugged in than ol-far-away me, but the basic flow would be

- Decide to work on something
- Sign up for Trello and add the Scrum plugin for point estimating (https://chrome.google.com/webstore/detail/scrum-for-trello/jdbcdblgjdpmfninkoogcfpnkjmndgje?hl=en)
- Go to https://trello.com/b/skuMQvLJ/loomio-sprint-board and add a card under Next Up/Ready
- Estimate it (The estimation is a little touchy feely; I'd say it roughly translates to 3 points = half day, 5/8 points = whole day, 21 points ~= 3-4 days of work, etc.)
- Do the work
- Move it into this sprint: complete
- Now we've got some insight and visibility towards your contributions, yay!

AaronThornton00:
@gdpelican thanks.  I have spoke to Rob briefly concerning this.  I really don't want to commit to anything   timely. As I'm pretty busy with Storypark and building our house.  I just wanted to tidy up a few ruff edges while I had some time spare. I will just push up little PRs as I see them, if that suits everybody.  I think the colours and the icons should be the next big improvement for V1.0.  I feel we are getting there, but I feel more people need to connect with it.

gdpelican:
Yeah, I understand, although I think that the less work we have floating out there that like 'Oh yeah someone did that last week', and more 'This piece of work was done by this person', is hugely beneficial. 

I'm in a similar situation for the immediate future where I'm picking up things that I find interesting  / pressing as my time allows, but don't have solid time commitments, but when it comes time to sit down at the end of the week and go 'okay, what got done', I think your contributions should be on the list, even if there's not expectations of you time-wise going forward.

gdpelican:
I'm happy to take this on myself, actually. I could imagine an ideal future world where we have several contributors to this thing at a time and it was the job of the OSS coordinator to categorize and track the contributions over a sprint, and maybe measure OSS velocity. That would be a cool world.

lineman/app/components/thread_page/thread_page.haml
```diff
@@ -52,3 +52,5 @@
 
     %activity_card{discussion: 'threadPage.discussion'}
     %comment_form{comment: 'threadPage.comment'}
+
+  %thread_lintel
```
AaronThornton00: @gdpelican I have moved the lentil out of the navbar as it really belongs in the %main.thread-page div. Then i can inherit all of its styles.
I have it working where it is but it needs to be indented on more position. Unfortunately it stops working for reasons i can't tell.  Do you think you could have a look at this when you get some time?

gdpelican: It's because you're moving it inside the if() block for the thread page, which means that it isn't there when the 'currentComponent' event which tells it that we're viewing the discussion page comes through.

I'll have a further poke a little later on; should be fixable.

gdpelican: Alright, that should bump ya forward a little @AaronThornton00

AaronThornton00:
@gdpelican thanks a million.  This looks good now on all devises.  I'll just check the flash it still all good.

rdbartlett:
Love where this is going. Fyi @AaronThornton00 @HSalmon is working on moving the flash to the bottom of the screen ([Material style](http://www.google.com/design/spec/components/snackbars-toasts.html)) so you might not need to worry.

gdpelican:
Let me know if you're happy with this @AaronThornton00, looks pretty good from my end.

AaronThornton00:
Yip good to go bro. Just been away from my puta so couldn't edit the
description.

PR #2320, opened 2015-07-18
Polish is LIVE

mixmix:
@robguthrie  translation updates 

Note, **Polish has completed 1.0**, and is going live in this deploy.
their current stats are : 

```diff
+  frontpage: 61
+  help docs: 11
+  main: 54
+  main_1_0: 100
```
@HSalmon you might want to reach out to the polish translator(s) and tell them what the process is. Tell them how to see their translations live, and that because the app is always evolving new translations will be added and old ones will be removed (so translation is an ongoing thing)

### DIFF
--- a/.translation_coverage.yml
+++ b/.translation_coverage.yml
@@ -2,7 +2,7 @@ an:
   frontpage: 52
   help docs: 8
   main: 88
-  main_1_0: 30
+  main_1_0: 29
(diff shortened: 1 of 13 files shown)


robguthrie:
This is INCREDIBLE!!!! Wow. I'm super excited.

PR #2321, opened 2015-07-18
add loomio 1.0 progress bars to /translation

mixmix:
adds table headers, and allows translators to see progress of new 1.0 translations:

selection_313

### DIFF
--- a/app/assets/stylesheets/unstructured/_pages.scss
+++ b/app/assets/stylesheets/unstructured/_pages.scss
@@ -537,7 +537,7 @@ body.pages.translation {
     margin-bottom: 10px;
   }
   p.code { padding-left: 16px; }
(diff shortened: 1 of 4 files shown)


robguthrie:
@mixmix could you please call it jsclient or something rather than loomio_1.0 or main_1.0 .. it's not a very usable term.

mixmix:
where are you talking @robguthrie ? 

If you're talking about what we present to users, jsclient is not very user-friendly. If you're talking about the names in `.translation_coverage.yml` e.g. 

```yml
an:
  frontpage: 52
  help docs: 8
  main: 88
  main_1_0: 30
```

those are relatively easy to change. Just tell me exactly what you'd like to see for each resource.
We might need to get hannah to double check docs for links if you want to rename the original 'main' though

PR #2327, opened 2015-07-21
use current user locale in angular

robguthrie:


app/helpers/locales_helper.rb
```diff
@@ -70,6 +70,12 @@ def language_options
     @language_options ||= language_options_for selectable_locales, nil, incomplete_locales
   end
 
+  def angular_locales
+    selectable_locales.map do |locale|
```
gdpelican: Someday I think we may need to objectify locales.. I keep having this thought that all of this craziness around locales would be mitigated by just having a Locale object which would have key, name, fallback, etc.

lineman/app/components/profile_page/profile_page.haml
```diff
@@ -25,7 +25,7 @@
         %validation_errors{subject: 'profilePage.user', field: 'email'}
 
         %label{for: 'user-locale-field', translate: 'profile_page.locale_label'}
-        %select.form-control.profile-page__language-input{ng-model: 'profilePage.user.selectedLocale', ng-required: 'true', ng-options: 'locale.key as locale.name for locale in profilePage.availableLocales()', ng-change: 'setCurrentPrivacy()'}
```
gdpelican: Heh whoops. Pretty sure that was me.

gdpelican:
Looks great!

PR #2329, opened 2015-07-21
css tweak to mini pie chart

rdbartlett:
Pie chart is still a bit buggy, but this makes it less buggy.


robguthrie:
thanks rich!

PR #2333, opened 2015-07-22
Display version in-app

gdpelican:
I love that we're versioning so much, I wanna show it to the world!

Also, since we're getting a bit closer to showing this to a slightly wider circle, thought the ugly pink thing in the corner could use just a little attention.

screen shot 2015-07-21 at 10 45 42 pm


mixmix:
noice :cocktail:

gdpelican:
Merge :monkey:!

PR #2334, opened 2015-07-22
locales update

mixmix:
@robguthrie


gdpelican:
The merge :monkey_face: is on a bender today. Thanks @mixmix

PR #2335, opened 2015-07-22
add dropdown arrows via mixin

AaronThornton00:
Here is some more pretty :lips: 

https://trello.com/c/pAcsh2JU/567-add-arrows-to-dropdowns-via-mixin

Arrows added to the notification and user dropdowns.
Made a mixin so feel free to use it. 
I will add this to the search when someone gets around to fixing the it. It's a little more difficult at the mo.

Tested: Crome(Mac) Firefox(Mac), S4, iPhone5,6+ portrait and landscape. (although I need to tweak the nav in general in lanscape on small devices)

Before:
screenshot 2015-07-22 22 40 14

After:
screenshot 2015-07-22 22 40 44
screenshot 2015-07-22 22 39 48

### DIFF
--- a/lineman/app/components/mixins.scss
+++ b/lineman/app/components/mixins.scss
@@ -163,7 +163,6 @@ $cardPaddingSize: 15px;
 }
 
 @mixin box_shadow ($level) {
-
   @if $level == 1 {
     box-shadow: 0 1px 3px rgba(0,0,0,0.12), 0 1px 2px rgba(0,0,0,0.24);
   } @else if $level == 2 {
@@ -175,7 +174,6 @@ $cardPaddingSize: 15px;
   } @else if $level == 5 {
     box-shadow: 0 19px 38px rgba(0,0,0,0.30), 0 15px 12px rgba(0,0,0,0.22);
   }
-
 }
 
 @mixin displayFlex {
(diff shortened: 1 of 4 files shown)


gdpelican:
Hey @AaronThornton00 this looks like a nice improvement, but it's bugging me a bit that the arrows are a little off the center of the icons for notifications / profile. Easy to fix?

AaronThornton00:
This should be better across all devices

screenshot 2015-07-23 08 24 24

AaronThornton00:
@gdpelican this is kind of best fit as the icons positions change across devices and it is using % to position it absolutely.  Personally I think this is pretty accurate.

robguthrie:
@rdbartlett shall I merge this?

rdbartlett:
sure

AaronThornton00:
@robguthrie yes please.  Once this is in we can apply it to all the other dropdowns.  Currently they are just floating in space.

lineman/app/components/navbar/navbar.scss
```diff
@@ -46,6 +46,18 @@
   a{ color: $primary-text-color; }
 }
 
+.lmo-navbar__item--user, .lmo-navbar__item--notifications {
```
robguthrie: @AaronThornton00 - I think I've mentioned to you that the CSS guidelines we're really enjoying say: "Dont use nesting unless you really really have to" - I'm wondering if you need to be using nesting here?

The other thing I want to do is use mixin arguments to set the left value rather than adding more selectors.
I'll show you what I mean in another comment.

lineman/app/components/mixins.scss
```diff
@@ -135,9 +134,44 @@ $block-border-color: darken($block-color, 10%);
   } @else if $level == 5 {
     box-shadow: 0 19px 38px rgba(0,0,0,0.30), 0 15px 12px rgba(0,0,0,0.22);
   }
-
 }
 
+@mixin dropdown_arrow {
```
robguthrie: You could write
```
@mixin dropdown_arrow($left) {
...
  .arrow::before {
    left: $left;
}

AaronThornton00:
Yeah totally Rob, that makes a lot of sense

lineman/app/components/mixins.scss
robguthrie: Also the class should be dropdown-arrow rather than .arrow  the idea with BEM is that everything is namespaced and nesting is not used because it's too complex and dumb

robguthrie:
Cool. Do you want to do that yourself, in your own time (there is no hurry) or shall I fix that up for you?

AaronThornton00:
Up to you man I'm at work now and have a meetup and a movie tonight so if your keen to get it in soon you might have to do it yourself.  It's a fairly medial change, and I fully get your logic here. Saying that I am also happy to do it if you don't mind waiting a couple of days.

AaronThornton00:
@robguthrie @gdpelican I have tidied this up considerably as Rob outlined above. One small nesting, but am wondering when to draw the line as it seems the code is getting quite polluted with all these extra classes at the cost of being absolute with the css.  Anyway big improvement and I liked your idea about the mixin parameters Rob.

PR #2337, opened 2015-07-22
Add pending invitations card

HSalmon:


lineman/app/models/invitation_model.coffee
```diff
@@ -0,0 +1,17 @@
+angular.module('loomioApp').factory 'InvitationModel', (BaseModel) ->
+  class InvitationModel extends BaseModel
+    @singular: 'invitation'
+    @plural: 'invitations'
+    @indices: ['id', 'groupId']
+
+    initialize: (data) ->
```
robguthrie: you can drop this.. it's already there from BaseModel

app/models/invitation.rb
```diff
@@ -40,6 +40,10 @@ def group
     end
   end
 
+  def group_id
```
gdpelican: Can we move the into the serializer to avoid putting extra stuff on the model?

```
def group_id
  object.inviter_id
end
```

or, probably even better:

```
attribute :invitable_id, key: :group_id
```

lineman/app/components/memberships_page/pending_invitations_card/pending_invitations_card.haml
```diff
@@ -0,0 +1,11 @@
+%section.pending-invitations-card
+  %h3.lmo-card-heading{translate: 'pending_invitations_card.heading'}
+  {{group.invitations()}}
+  %ul.pending-invitations-card__pending-invitations
+    / %li.media.pending-invitations-card__pending-invitation{ng-repeat: 'invitation in group.pendingInvitations() track by invitation.id'}
```
gdpelican: Is this a work-in-progress?

HSalmon:
I think this is ready for review now @robguthrie, @gdpelican

app/controllers/api/invitations_controller.rb
```diff
@@ -10,6 +10,18 @@ def create
     respond_with_collection
   end
 
+  def pending
+    load_and_authorize :group
```
robguthrie: Need to check that the user is a group admin. IE can :view_pending_invitations, @group

app/controllers/api/invitations_controller.rb
```diff
@@ -10,6 +10,18 @@ def create
...
+    @invitations = page_collection(@group.invitations.pending)
+    respond_with_collection
+  end
+
+  def destroy
+    @invitation = Invitation.find(params[:id])
```
robguthrie: This should be done in the InvitationService, where it will check that you have permission to do this.

app/controllers/api/invitations_controller.rb
gdpelican: Note that you can pass a particular permission to load_and_authorize, i.e.

```
load_and_authorize :group, :handle_invitations # or whatever the ability is.
```

app/controllers/development_controller.rb
```diff
@@ -335,6 +342,14 @@ def membership_request_from_user
     @membership_request_from_user
   end
 
+  def pending_invitation
+    pending_invitation = InvitationService.invite_to_group(recipient_emails: ['judd@example.com'],
+                                                           message: 'Come and join the group!',
+                                                           group: test_group,
+                                                           inviter: patrick)
+    pending_invitation
```
gdpelican: Confused by this; we want to memoize it I think, so it doesn't get created more than once? Definitely don't need to store in a variable and then return that variable I don't think.

```
@pending_invitation ||= InvitationService.invite_to_group(...)
```

lineman/app/components/memberships_page/cancel_invitation_form/cancel_invitation_form.coffee
```diff
@@ -0,0 +1,9 @@
+angular.module('loomioApp').factory 'CancelInvitationForm', ->
+  templateUrl: 'generated/components/cancel_invitation_form/cancel_invitation_form.html'
+  controller: ($scope, invitation, FlashService, Records) ->
+    $scope.invitation = invitation
+
+    $scope.submit = ->
```
gdpelican: I believe the FormService is merged in, so let's use that instead. There's a bunch of examples in the other forms now.

lineman/app/components/memberships_page/pending_invitations_card/pending_invitations_card.haml
```diff
@@ -0,0 +1,18 @@
...
+        %tr
+          %th{translate: 'pending_invitations_card.email'}
+          %th{translate: 'pending_invitations_card.invitation_link'}
+          %th{translate: 'pending_invitations_card.cancel'}
+      %tbody
+        %tr{ng-repeat: 'invitation in group.pendingInvitations() track by invitation.id'}
+          %td.pending-invitations-card__recipient-email
+            {{invitation.recipientEmail}}
+          %td.pending-invitations-card__invitation-url
+            https://www.loomio.org/invitations/{{invitation.token}}
```
gdpelican: Whoa, what's this doing here? We'll have to pull out the host name for third party installs. Also can't quite wrap my head around why we need this here?

app/controllers/api/invitations_controller.rb
```diff
@@ -28,8 +40,4 @@ def invitation_parser
     @invitation_parser ||= InvitationParser.new(@invitations)
   end
 
-  def resource_serializer
```
gdpelican: Hey this is going to break the existing invitation system pretty badly, you may need to rethink how you're serializing these, or allow the new serializer to handle both cases

HSalmon: @robguthrie, do you have thoughts as to how to handle this?

lineman/app/components/memberships_page/pending_invitations_card/pending_invitations_card.haml
robguthrie: sokay.. it just needs to be a relative url.. speaking of URLS we should probably be using our url helper eh?

gdpelican: I thought the point was to copy-paste and send to people?

In any case I want a card in here for adding a 'resend invitation' link to this card, which will make it much better.

app/controllers/api/invitations_controller.rb
robguthrie: @gdpelican would be good if we can use normal invitation models in the existing invitation system. Is there anything stopping us from going that way?

lineman/app/components/memberships_page/cancel_invitation_form/cancel_invitation_form.haml
```diff
@@ -0,0 +1,12 @@
...
+  .modal-header
+    %button{type: 'button', class: 'close', ng-click: '$close()'}
+      %span{aria-hidden: true} &times;
+      %span.sr-only{translate: 'common.action.cancel'}
+    %h1.lmo-h1{translate: 'cancel_invitation_form.heading'}
+
+  .modal-body
+    %span{translate: 'cancel_invitation_form.question'}
+  .modal-footer
+    %button.btn.btn-danger{type: 'submit', translate: 'cancel_invitation_form.submit'}
```
HSalmon: This is what's causing the tests to fail. When I put a BEM class on this button, and then inspect it in the browser, it's not there, so when protractor tries to find the element by css it can't... @robguthrie, suggestions? Also the whole thing might need a quick going over...

robguthrie: if you put a class on it.. and the class is not there.. then I think you're editing the wrong file/another file is being used and we need to get to the bottom of that.

HSalmon:
@robguthrie, @gdpelican - reckon this could use a once over...

spec/controllers/api/invitations_controller_spec.rb
```diff
@@ -56,4 +58,12 @@
     # end
   end
 
+  describe 'pending' do
```
robguthrie: please add a test that hitting this url as non admin gives unauthorized

lineman/spec-e2e/helpers/invitations_helper.coffee
```diff
@@ -18,3 +21,19 @@ module.exports = new class InvitationsHelper
 
   submitInvitationsForm: ->
     element(By.css('.invitation-form__submit')).click()
+
+  visitMembershipsPage: ->
```
robguthrie: I would call this 'clickManageMembers' visit tells me you're going to enter a url

HSalmon:
@robguthrie, feedback implemented :)

robguthrie:
It's great!!!

robguthrie:
I'll merge when the tests pass

PR #2342, opened 2015-07-23
All activity in your group happens live

robguthrie:
I think this is ready for review. Works well for me.. New discussions and comments come up for any discussions in my dashboard. Also notificaitons come up instantly too.

### DIFF
--- a/lineman/app/models/discussion_model.coffee
+++ b/lineman/app/models/discussion_model.coffee
@@ -73,7 +73,10 @@ angular.module('loomioApp').factory 'DiscussionModel', (BaseModel) ->
       !@reader().discussionId?
 
     isUnread: ->
-      @unreadActivityCount() > 0
+      if @reader().lastReadAt?
+        @unreadActivityCount() > 0
+      else
+        true
 
     isMuted: ->
       @reader().volume == 'mute'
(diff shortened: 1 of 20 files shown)


lineman/app/models/discussion_model.coffee
```diff
@@ -73,7 +73,10 @@ angular.module('loomioApp').factory 'DiscussionModel', (BaseModel) ->
       !@reader().discussionId?
 
     isUnread: ->
-      @unreadActivityCount() > 0
+      if @reader().lastReadAt?
```
robguthrie: We had a bug in angular! New unread discussions would show up as read.. try it!

lineman/app/models/discussion_reader_model.coffee
```diff
@@ -12,11 +12,13 @@ angular.module('loomioApp').factory 'DiscussionReaderModel', (BaseModel) ->
       @volume = null
       @lastReadSequenceId = -1
 
-      if data.discussion_id?
-        discussion = @recordStore.discussions.find(data.discussion_id)
-        _.extend data,
-          id:       discussion.id
-          group_id: discussion.groupId
+      # commented so I can ask gdpelican about it
```
robguthrie: @gdpelican I've taken this out because it was giving me problems when inserting my own discussion reader.. I can probably put it back in though.. was just wondering why we do it.. it looks hacky but don't take that personally.

gdpelican: Won't this mean you can't call `discussion_readers.find(discussion_id)` and find the thing you're looking for?

Don't think I wrote this, but can't remember.

PR #2344, opened 2015-07-24
fix for landscape

AaronThornton00:
- Extend navbar auto centring icons for landscape phone and portrait tablet
- Expand button link to fill .lmo-navbar__item container so it looks prettier.

Before:
screenshot 2015-07-24 21 29 33

After:
screenshot 2015-07-24 21 29 01

### DIFF
--- a/lineman/app/components/navbar/navbar.scss
+++ b/lineman/app/components/navbar/navbar.scss
@@ -94,13 +94,22 @@
   text-align: center;
 }
 
-@media (max-width: 480px) {
+@media (max-width: 767px) {
+  .lmo-navbar__item {
+    display: inline-block;
+    text-align: center;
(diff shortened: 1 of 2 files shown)


lineman/app/components/navbar/navbar.scss
```diff
@@ -94,13 +94,22 @@
   text-align: center;
 }
 
-@media (max-width: 480px) {
+@media (max-width: 767px) {
+  .lmo-navbar__item {
+    display: inline-block;
+    text-align: center;
+    .lmo-navbar__btn {
```
robguthrie: Do you need to be nesting this rule? If it was a sibling to .lmo-navbar__item it would have the same effect right?

PR #2345, opened 2015-07-24
enlarge graph and layer position

AaronThornton00:
This little puppy is pretty much good to go.  (fixed ellipse issue) 
##### At the moment there is just a avatar placeholder hardcoded into the code, which someone can replace with logic. #####
I really hope you all like this!!!  It was just about a whole days work.

- hover buttons are now positioned absolutely
- this allows longer titles
- increased font-weight of thread title to allow for smaller line-heights
- more text in less space
- voting icons are now visually associated with the pie.  Which helps understand what they mean.
- mobile experience is clearer too

iphone5:
screenshot 2015-07-27 09 21 36
iphone6:
screenshot 2015-07-27 09 22 26

screenshot 2015-07-27 09 15 21
screenshot 2015-07-27 09 25 07

### DIFF
--- a/lineman/app/components/group_page/group_page.scss
+++ b/lineman/app/components/group_page/group_page.scss
@@ -17,6 +17,10 @@
   background-size: cover;
 }
 
+.group-page__description {
+  @include card;
+}
+
 .group-page__header{
   width: 100%;
   position: relative;
@@ -49,70 +53,8 @@
   margin-left: 108px + 15px;
   a{color: $primary-text-color;}
 }
+
 .group-page__privacy{
   float: left;
   margin-left: 15px;
-}
-.group-page__description,
-.group-page__threads,
-.group-page-members{
-  @include card;
-}
-
-.group-page__new-thread{
-  float: right;
-  margin-top: -5px;
-  margin-bottom: 10px;
-  .btn { @include fontSmall; }
-}
-.members-card__invite-members {
-  clear: both;
-  padding-top: 10px;
-}
-.group-member-list-avatar{
-  float: left;
-  margin: 0 2px 2px 0;
-}
-
-.group-page__threads-placeholder {
-  clear: both;
-}
-
-.group-thread-list-empty {
-  padding: 20px 0;
-  text-align: center;
-}
-.group-thread-list{
-  border-top: 1px solid $border-color;
-  clear: both;
-}
-.group-thread-list .thread-preview {
-  padding: 13px 0px;
-}
-
-.subgroups-card{
-  @include card;
-}
-.subgroups-card__list{
-  list-style: none;
-  padding: 0;
-  margin: 0;
-}
-.subgroups-card__list-item{
-  clear: both;
-  min-height: 50px;
-  margin-bottom: 12px;
-}
-.subgroups-card__list-item-logo{
-  width: 50px;
-  height: 50px;
-  background: url('http://placehold.it/50x50');
-  float: left;
-}
-.subgroups-card__list-item-name{
-  @include fontHelveticaBold;
-  margin-left: 60px;
-}
-.subgroups-card__list-item-description{
-  margin-left: 60px;
 }
--- a/lineman/app/components/mixins.scss
+++ b/lineman/app/components/mixins.scss
@@ -82,6 +82,12 @@ $block-border-color: darken($block-color, 10%);
   font-stretch:normal;
 }
 
+@mixin truncateText {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
 @mixin roundedCorners {
   -webkit-border-radius: 3px;
   -moz-border-radius: 3px;
@@ -99,26 +105,34 @@ $block-border-color: darken($block-color, 10%);
   height: 20px;
 }
 
+$boxSmallSize: 30px;
 @mixin boxSmall {
-  width: 30px;
-  height: 30px;
+  width: $boxSmallSize;
+  height: $boxSmallSize;
 }
 
+
+$boxMediumSize: 50px;
 @mixin boxMedium {
-  width: 50px;
-  height: 50px;
+  width: $boxMediumSize;
+  height: $boxMediumSize;
 }
 
+$boxLargeSize: 80px;
 @mixin boxLarge {
-  width: 80px;
-  height: 80px;
+  width: $boxLargeSize;
+  height: $boxLargeSize;
 }
 
 @mixin card{
+  @include cardNoPadding;
+  @include cardPadding;
+}
+
+@mixin cardNoPadding {
   @include roundedCorners;
   @include box_shadow(1);
   border: 1px solid $border-color;
-  padding: 17px;
   background: #fff;
   margin-bottom: 8px;
   i{ color: $grey-on-white; }
@@ -126,6 +140,12 @@ $block-border-color: darken($block-color, 10%);
     margin: 10px 0 0 0;
     clear: both;
   }
+}
+
+$thinPaddingSize: 10px;
+$cardPaddingSize: 15px;
+@mixin cardPadding {
+  padding: $cardPaddingSize;
 }
 
 @mixin thinRoundedBorder {
@@ -156,6 +176,14 @@ $block-border-color: darken($block-color, 10%);
     box-shadow: 0 19px 38px rgba(0,0,0,0.30), 0 15px 12px rgba(0,0,0,0.22);
   }
 
+}
+
+@mixin displayFlex {
+  display: flex;
+  /*display: -webkit-box;*/
+  /*display: -moz-box;*/
+  display: -ms-flexbox;
+  display: -webkit-flex;
 }
 
 
--- a/lineman/app/components/thread_preview/thread_preview.scss
+++ b/lineman/app/components/thread_preview/thread_preview.scss
@@ -1,5 +1,5 @@
 .thread-preview {
-  padding: 13px 13px;
+  min-height: $boxMediumSize + ($thinPaddingSize * 2);
   border-bottom: 1px solid $border-color;
   position: relative;
   &:hover {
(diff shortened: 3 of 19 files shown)


AaronThornton00:
I think a nice and easy progression from this would be for clicking on the position or the unvoted '?' to take you to the active proposal so you can participate or see more info.

robguthrie:
Wow. looks great @AaronThornton00. I think the position on bottom right looks great too.

However it seems odd to me that the title runs along the top. Would you be open to having the all the text to the right of the pie, and the pie full height?

When there is no pie, maybe author image? I'm pretty sure material has this kind of thing pretty well covered, and rich had something like it in his material visioning stuff.

robguthrie:
Also the number of lines of text bothers me. I think it's too many. Had any ideas about reducing total number of lines?

AaronThornton00:
@robguthrie totally agree.  I didn't want to go too far with this branch until I felt people liked it.  
I'll have a play with the text a bit more then.

robguthrie:
Thing is, for us to change the style it should go "all the way" to where it wants to go.. otherwise we're bringing in something that will change very soon again, which is even more work to support.

Do you have a copy of sketch?

AaronThornton00:
Yeah I understand, I am just trying not to stand on anyones toes. As I am not being paid for my contributions I have to check-in at various points so as not to waste my time. Currently I look at the current mobile experience, and think what is the point in showing the first six letters of the proposal name? It seems like that wasn't properly designed, or perhaps not properly tested.

I will have a play when I get some time, if I make any progress I will let you know. If I don't you haven't lost too much time. It doesn't take much time to move style text.

robguthrie:
I see what you mean about the closing text. That is an improvement!

AaronThornton00:
How does this grab ya @robguthrie 
screenshot 2015-07-26 13 20 27

- Hover action buttons are positioned absolutely over the top so you don't have to truncate the title early.
- Three lines of text.
- replace pie with avatar when no proposal

AaronThornton00:
@robguthrie @gdpelican @rdbartlett  I have updated this pull request now. Remove the out dated comments

AaronThornton00:
Perhaps when the proposal closes the position icon disappears, I think this would be a good visual clue to weather the proposal is active or not.

robguthrie:
Yea.. I like the idea that we keep the pie after closed. If we have a symbol for "outcome" that would help

robguthrie:
WOW though. This is really great @AaronThornton00!!!!!!!!

robguthrie:
I really like it

AaronThornton00:
Stoked.

rdbartlett:
This is beautiful. Can I see a version where the avatar is in its regular rounded square instead of a circle? The AT in a circle was a bit confusing on first glance because it looks similar to a pie. Might look ugly mixing squares and circles though.

AaronThornton00:
Here you go @rdbartlett.  I think the rounded square works fine.
Personally I'm not a fan of round avatars.  Reminds me of skype.

screenshot 2015-07-27 09 54 17

AaronThornton00:
Or... no border radius
screenshot 2015-07-27 10 36 38

rdbartlett:
Thanks speedy gonzales. Yep I think squares and circles look cool together
after all. I think we should stick with the standard rouded square avatar
for uniformity.

AaronThornton00:
Yep you should just about be able to copy the same code in

lineman/app/components/group_page/group_page.scss
```diff
@@ -59,12 +59,23 @@
   @include card;
 }
 
+.group-page__threads {
+  padding: 10px 0 17px;
+  h2 {
```
robguthrie: Again.. nesting. Can we avoid it?

lineman/app/components/mixins.scss
```diff
@@ -95,8 +95,8 @@ $block-border-color: darken($block-color, 10%);
 }
 
 @mixin boxTiny {
-  width: 20px;
-  height: 20px;
+  width: 40px;
```
robguthrie: Sorry but this is not allowed to be changed. Tiny is 20px that's the rules. We ahve 20, 30 and 50 px things.. Can you use one of those sizes?

lineman/app/components/group_page/group_page.scss
AaronThornton00: Yep of course we can.  Pretty sure the h2 will need a new class though

lineman/app/components/mixins.scss
AaronThornton00: No wazzas, i'll loose this class from the pie

lineman/app/components/group_page/group_page.scss
AaronThornton00: Actually I take that back.  The h2 styling needs to apply to the group description and the discussion.  So in this case it needs to be nested.

AaronThornton00:
@robguthrie I think this is good if you have no more comments

lineman/app/components/group_page/group_page.scss
robguthrie: Is it the same heading style as used elsewhere in the app? If so we have .lmo-card-heading

robguthrie: see lineman/app/components/utilities.scss

lineman/app/components/group_page/group_page.scss
```diff
@@ -59,12 +59,23 @@
...
+    padding-top: 7px;
+  }
+}
+
 .group-page__new-thread{
-  float: right;
-  margin-top: -5px;
-  margin-bottom: 10px;
-  .btn { @include fontSmall; }
+  .btn {
```
robguthrie: I think the correct BEM name here is .group-page__new-thread-button (rather than nesting .btn)

lineman/app/components/thread_preview/thread_preview.scss
```diff
@@ -8,72 +8,121 @@
   }
 }
 
+.thread-preview__avatar {
+  display: inline-block;
+  width: 10%;
+  .placeholder {
+    border-radius: 4px;
```
robguthrie: we have a roundedCorners mixin.. so please write
```
@include roundedCorners
```

robguthrie:
I don't think using percentages is the right way to position this. I'll have a look at it tomorrow, but I would have assumed this is something we want to use flexbox and min-width px for.

robguthrie:
Damn it looks good though.

AaronThornton00:
There is many ways to skin it. I find Percentages just give you
responsiveness to resizing across browsers. If someone else wants to code
it differently. I'm cool with it.

lineman/app/components/group_page/group_page.scss
AaronThornton00: As with the headers above I have tried this naming convention and it doesn't work.  Feel free to try and refactor it yourself.   I have tested this PR in 3 browsers and just about every mobile device, if you make changes please retest.

lineman/app/components/group_page/group_page.scss
AaronThornton00: Rob that is not going to work. The lmo-card-heading does not have a margin-left for a reason. It relies on the padding of the container. If you look at the diff you will see I have removed a whole lot of negative margins, which I consider to be 'incorrect' css, 'incorrect' not because the script is ugly , but because it behaves differently in different browsers. It's un predictable. As the margins in the container need to be 0 to allow the separating lines between the threads to span 100%, I need to pad the headers for this case.  I personally don't think nesting is dumb it can be very powerful.  If you look at the BEM docs it emphasises that it is not absolute and does not work in all cases.  To me it is more important how it looks to the user and how it scales across browsers and different devices.  The css layer should be flexible and changing as the product changes.  Nesting a header is not going to stop that.  Feel free to refactor once it is in place.

lineman/app/components/thread_preview/thread_preview.scss
AaronThornton00: This I can do, although it is really just a temp class for demonstration purposes.

AaronThornton00:
@robguthrie I appreciate the tidy codebase and you all are doing an awesome job, so I hope my comments don't grate at anyone.  I like BEM and I tried to use it where I could.  What I have written and the changes I have made have replaced some borderline css. Including float within floats, floats without clearfixs, and negative margins. My contributions is changing a lot in a short amount of unpaid time so I can't always get things perfect.  Feel free to refactor at a later date. This PR will also need testing in IE too as I don't have it on my dev env.  I have tested this PR in 3 browsers and just about every mobile device, if you make changes please retest.

AaronThornton00:
Let me know if you want me to squash it

robguthrie:
Thanks @AaronThornton00 I really appreciate your work. Even being paid I've not had time to actually check out and run the branch yet, but I'll do that this afternoon.

I agree that negative margins are bad and am keen to remove them. I'll have a go at removing percentages if I don't like how the page behaves and let you know how I go.

robguthrie:
Regarding padding etc.. I think we need to resolve the problem you're running up against. Should cards not have padding in them by default? Maybe we should have a card and card-inner so that padding can be optional?

robguthrie:
P.s. I'm not suggesting you do this work.. Just want your opinion so I can fix this for the future.

AaronThornton00:
All good man. Have a play

AaronThornton00:
It plays very nicely!! :space_invader: :space_invader:

lineman/app/components/utilities.scss
```diff
@@ -154,6 +154,18 @@ label.label-with-details{
   margin-top: 22px;
 }
 
+.lmo-card {
```
robguthrie: @AaronThornton00 I've removed all the padding declarations in favor of using lmo-card and lmo-card--no-padding.

robguthrie:
@AaronThornton00 I've cleaned up a whole bunch of css - moving css that was in the wrong places into the right places (ie: All css in a file called discussions-list has a block name of discussions-list, yay)

And I've tried to fix the problems with padding. Now you can use cardNoPadding when you don't want the padding and then chuck in a card-padding when you need it.

I've still got to change the thread-preview but it does need to change. having tried it I see that the text cant use all the space due to the current % design. I'm sure we can improve that by using flexbox.. which I'll have a crack at in a few hours.

AaronThornton00:
What text?  The discussion name has to stop short for the star and the proposal title has to stop short due to the closing soon message. You can easily extend it to 100%.... But id you want to add your flexbox...

AaronThornton00:
Nice work on the css rob

robguthrie:
Ok.. I've rebuilt it. Percentages are just not going to give a good enough result.

Showing closing soon on a small screen:
image

With lots of room:
image

With action buttons showing (might lift them a few px so they don't stop the proposal title):
image

robguthrie:
@AaronThornton00 what do you reckon of hand on the left? ( I like it because it leaves the area next to the start of the text cleaner)

AaronThornton00:
@robguthrie tight!.  Nice work. :facepunch:   I think I actually prefer it on the left.  It's less cluttered away from the text

AaronThornton00:
You have broken that dodgy mute button.  I did the same.  There is a little x that is absolutely positioned

robguthrie:
yea dodgy as mute button

AaronThornton00:
Also just check a long thread header with an unread count on the end

AaronThornton00:
There must be an icon for that mute

AaronThornton00:
If not I will make one

robguthrie:
Long header with unread count is fine.

AaronThornton00:
+ :metal:

rdbartlett:
omg this rules

robguthrie:
FYI - I've tested on chrome, ff, safari and made the necessary changes. I'll fix the tests today and I think this is good to go. 

I'm starting to feel like I can do CSS properly.

Also I think we need to consider changing our padding sizes in the future (just 2 vars used everywhere) as it's kind of intense on mobile.

Reckon we might get this merged today.

AaronThornton00:
Yus!

rdbartlett:
KAPOWWWW

PR #2347, opened 2015-07-25
get more info from rack attack

robguthrie:
@gdpelican is this ok to merge? Just wanna see more data.


gdpelican:
I think so, yeah.

PR #2350, opened 2015-07-27
locales update - it mostly done in 1.0

mixmix:
where it = italian

@robguthrie


mixmix:
someone is making a tentative start on Maori : 



you know who it is at all @HSalmon   ?

PR #2351, opened 2015-07-29
Add separate directives for memberships page

HSalmon:


HSalmon:
@robguthrie, I think this is ready to go...

lineman/app/components/memberships_page/memberships_panel/memberships_panel.haml
```diff
@@ -0,0 +1,16 @@
+.blank
```
robguthrie: No need to do this now.. but FYI .blank is a hack for a bug we're discovered when you try to put ng-if on the first element in a template. When there is no .ng-if  then you don't need .blank.. however this can't hurt.

robguthrie:
Thanks @HSalmon !

PR #2353, opened 2015-07-30
Fix invitation message bug

HSalmon:
Tested this in the browser and it works, but it's hard to write tests because we still don't have a good method for checking the content of emails...maybe this is something we can look into?


robguthrie:
Thanks @HSalmon !

PR #2355, opened 2015-07-30
add miniprofiler

robguthrie:
need to test on clone before deploy


gdpelican:
What's the status on this one @robguthrie ?

robguthrie:
close for now, we can revisit later on. Thanks

PR #2356, opened 2015-07-30
Add validation to forms

robguthrie:
sits on top of maxlength PR.. but it's only really 2 lines.

@gdpelican I've not tested this yet.. but is adding %validation_errors like so the only thing we need do to use that directive?


gdpelican:
Should be just fine.

To test, might need to put 'novalidate' on the html form to stop the Browsers That Dont Suck :tm: from intercepting the form submit (or use a Browser That Sucks :tm:)

PR #2358, opened 2015-07-31
Fix broken add subgroup link

HSalmon:


HSalmon:
Hey @robguthrie, I think this is ready to go if tests pass...

PR #2360, opened 2015-08-01
Fix toggle coordinator

gdpelican:
These are careless mistakes, guys. I thought we had a test around adding / removing a coordinator?


robguthrie:
We do! spec-e2e/manage_memberships_spec.coffee covers adding but not removing coordinator. What is broken without this patch?

gdpelican:
Hm, both adding and removing would be broken, with a 'Records is not
defined' error every time you try to hit the toggleAdmin button

gdpelican:
Ah, I guess the e2e tests pass ok because we successfully add a 'coordinator' class to the user avatar, even though the post to the server fails.

Negative side effect of 'fire and forget', I suppose, although adding a line to test the flash message showing up might help cover this a bit better.

robguthrie:
Thanks for the info. Seems like checking for flash messages is a simple way
to get complete testing. I'll add it to the dev guide.

robguthrie:
Also I think calling this careless might be a bit careless. I think we just
need to do formal smoke testing (ie: another person verifies it works) if
we want to catch these things.

gdpelican:
Careless might not have been the right word, but I do feel like this is the
type of regression that'll be unacceptable once our userbase is on this
thing, especially when any of a) a simple code review, b) a smoke test, or
c) a complete regression test would have caught it, meaning we need to be a
little more vigilant about these things going forward.

robguthrie:
I agree a simple process improvement will help. Do you want to try to name the process improvement you would like to see?

PR #2362, opened 2015-08-01
Disable forms while processing

gdpelican:
Adds a nice white overlay to forms while they're processing.

### DIFF
--- a/lineman/app/components/thread_page/comment_form/comment_form.coffee
+++ b/lineman/app/components/thread_page/comment_form/comment_form.coffee
@@ -7,13 +7,12 @@ angular.module('loomioApp').directive 'commentForm', ->
     group = $scope.comment.discussion().group()
     discussion = $scope.comment.discussion()
 
-    applyFormService = (=>
-      $scope.submit = FormService.submit $scope, $scope.comment,
-        flashSuccess: 'comment_form.messages.created'
-        successCallback: =>
-          $scope.comment.body = ''
-          $scope.comment.newAttachmentIds = []
-    )()
+    $scope.submit = FormService.submit $scope, $scope.comment,
+      flashSuccess: 'comment_form.messages.created'
+      successCallback: =>
+        $scope.comment.body = ''
+        $scope.comment.newAttachmentIds = []
+        $scope.comment.parentId = null
 
     $scope.$on 'replyToCommentClicked', (event, parentComment) ->
       $scope.comment.parentId = parentComment.id
--- a/lineman/app/components/thread_page/comment_form/comment_form.coffee
+++ b/lineman/app/components/thread_page/comment_form/comment_form.coffee
@@ -3,15 +3,17 @@ angular.module('loomioApp').directive 'commentForm', ->
   restrict: 'E'
   templateUrl: 'generated/components/thread_page/comment_form/comment_form.html'
   replace: true
-  controller: ($scope, FlashService, Records, CurrentUser) ->
+  controller: ($scope, FormService, Records, CurrentUser) ->
     group = $scope.comment.discussion().group()
     discussion = $scope.comment.discussion()
 
-    $scope.submit = ->
-      $scope.comment.save().then ->
-        $scope.comment = Records.comments.build(discussion_id: discussion.id)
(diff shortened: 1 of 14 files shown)


lineman/app/components/thread_page/comment_form/comment_form.coffee
```diff
@@ -3,15 +3,17 @@ angular.module('loomioApp').directive 'commentForm', ->
...
+  controller: ($scope, FormService, Records, CurrentUser) ->
     group = $scope.comment.discussion().group()
     discussion = $scope.comment.discussion()
 
-    $scope.submit = ->
-      $scope.comment.save().then ->
-        $scope.comment = Records.comments.build(discussion_id: discussion.id)
-        $scope.$emit('commentSaveSuccess')
-        FlashService.success('comment_form.messages.created')
+    applyFormService = (=>
```
robguthrie: in coffee you can write `do` to invoke a function

gdpelican: Ah, good know, although looking at this again, I don't think I need to name this function, I can just say `$scope.submit = FormService.submit`

Before I was trying to do something where I called applyFormService in the success callback to reset the comment, but didn't work.

lineman/app/components/thread_page/comment_form/comment_form.coffee
```diff
@@ -3,15 +3,16 @@ angular.module('loomioApp').directive 'commentForm', ->
...
 
-    $scope.submit = ->
-      $scope.comment.save().then ->
-        $scope.comment = Records.comments.build(discussion_id: discussion.id)
-        $scope.$emit('commentSaveSuccess')
-        FlashService.success('comment_form.messages.created')
+    $scope.submit = FormService.submit $scope, $scope.comment,
+      flashSuccess: 'comment_form.messages.created'
+      successCallback: =>
+        $scope.comment.body = ''
```
robguthrie: I think it's a bit weird/wrong choosing to reset fields rather than building a new comment. It could lead to problems in the future.

It's only a side effect of the current create process that the actual comment object from the form is not also the final comment object when saved. When you update a comment record we clone it for this very reason. Are you ok with returning to the `Records.comments.build` method of reset?

robguthrie:
Looks great except for the comment reset thing I've commented on. However I feel a bit blind looking at this because I'm not sure what a disabled form will look like, are you able to include a screenshot of how this affects things?

lineman/app/components/thread_page/comment_form/comment_form.coffee
gdpelican: Hm, I did try to stick with that method and wasn't able to get it to go, the intuitive '$scope.comment = Records.comments.build' implementation has the awkward behaviour of clearing the comment form, but then re-posting your first comment when you submit again. 

I'm pretty sure it's because FormService retains a reference to the $scope.comment which was passed in, so changing the comment on the scope doesn't affect the comment the FormService is looking at.

This isn't a problem in other places because our forms are mostly in modals, so resetting the model afterwards isn't a thing.

Maybe we need to expose a method on the form service to reset the model it's looking at somehow?

robguthrie: That's weird that the formservice holds state from the previous use. It seems to me that the $scope.comment reference is old because you need to wrap the FormService.submit in a function..

gdpelican: FormService.submit does return a function, but I think it might need to
take the model as an argument, which I'll test out at some point

code is clearer; instead of

```
submit: (scope, model, options) ->
  ->
```

I think it needs to be
```
submit: (scope, options) ->
  (model) ->
```

Which involves changing all of the `ng-click: 'submit()' calls in our views
to take a model.

gdpelican: Anyway I think I'm going to back this change out of this PR and make it in a different one.

gdpelican:
Yep, a 'disabled' form just has a absolute div over it which fades it out, so it looks like it's not interactive anymore.
screen shot 2015-08-02 at 3 02 23 am

PR #2363, opened 2015-08-01
Fix deactivate group

gdpelican:
:see_no_evil:

### DIFF
--- a/app/serializers/group_serializer.rb
+++ b/app/serializers/group_serializer.rb
@@ -22,7 +22,8 @@ class GroupSerializer < ActiveModel::Serializer
              :logo_url_medium,
              :cover_url_desktop,
              :has_discussions,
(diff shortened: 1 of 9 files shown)


app/serializers/group_serializer.rb
```diff
@@ -22,7 +22,8 @@ class GroupSerializer < ActiveModel::Serializer
              :logo_url_medium,
              :cover_url_desktop,
              :has_discussions,
-             :has_multiple_admins
+             :has_multiple_admins,
```
robguthrie: I think we only need to send has_multiple_admins over if the user is an admin. Is it less computation to check if the current user is an admin before sending this attribute? (Sorry for the unrelated question)

robguthrie:
Thank yoooouuu

PR #2365, opened 2015-08-02
Fix broken underline on active page

gdpelican:
This broke at some point!

screen shot 2015-08-02 at 11 13 13 am


robguthrie:
Thanks James.

PR #2367, opened 2015-08-02
Add subgroup logos to group page

gdpelican:


robguthrie:
Nice one. Thanks James. It's great to get a group_avatar directive.. however I predict that group logos are all going to be the same on the subgroups_card and that we need to change the design. Never mind about that.. we'll see what happens. Thanks for doing this.

PR #2369, opened 2015-08-03
Dashboard shuffle (not for merging yet)

gdpelican:
Alright, this is ready for a look I think.

Notable things that are still bad at the moment:

- Visit inbox, then visit dashboard. No starred / proposal threads. Hit the 'recent' button again. Starred / proposal threads show up. Weep softly. (EDIT: fixed this, which takes me from `weep` to `wee!`)
- The muted e2e is failing for me because it's saying no muted threads are found. Works perfectly for the participating e2e (which is activated the same way), and works fine manually in the browser. Sigh.
- Need some optimization work on the dashboard query, I think because the discussion_serializer is loading up an author, group, and motion for each of 50 discussions.

### DIFF
--- a/app/controllers/api/discussions_controller.rb
+++ b/app/controllers/api/discussions_controller.rb
@@ -3,60 +3,44 @@ class API::DiscussionsController < API::RestfulController
   load_resource only: [:create, :update]
 
   def dashboard
-    instantiate_collection { |collection| filter_collection collection }
-    respond_with_discussions
+    instantiate_collection { |collection| collection_for_dashboard collection }
+    respond_with_collection serializer: DiscussionWrapperSerializer, root: 'discussion_wrappers'
+  end
+
+  def inbox
+    instantiate_collection { |collection| collection_for_inbox collection }
+    respond_with_collection serializer: DiscussionWrapperSerializer, root: 'discussion_wrappers'
   end
 
   def index
     load_and_authorize :group if params[:group_id] || params[:group_key]
     instantiate_collection
-    respond_with_discussions
-  end
-
-  def show
-    respond_with_discussion
+    respond_with_collection serializer: DiscussionWrapperSerializer, root: 'discussion_wrappers'
   end
 
   private
-
-  def respond_with_discussion
-    if resource.errors.empty?
-      render json: DiscussionWrapper.new(discussion: resource, discussion_reader: discussion_reader),
-             serializer: DiscussionWrapperSerializer,
-             root: 'discussion_wrappers'
-    else
-      respond_with_errors
-    end
-  end
-
-  def respond_with_discussions
-    render json: DiscussionWrapper.new_collection(user: current_user, discussions: @discussions),
-           each_serializer: DiscussionWrapperSerializer,
-           root: 'discussion_wrappers'
-  end
 
   def visible_records
-    Queries::VisibleDiscussions.new(user: current_user, groups: visible_groups).sorted_by_latest_activity
+    Queries::VisibleDiscussions.new(user: current_user, groups: visible_groups)
   end
-
-  private
 
   def visible_groups
     Array(@group).presence || current_user.groups
   end
 
-  def filter_collection(collection, filter = params[:filter])
+  def collection_for_dashboard(collection, filter: params[:filter])
     case filter
-    when 'show_proposals'     then collection.not_muted.with_active_motions
-    when 'show_participating' then collection.not_muted.participating
-    when 'show_starred'       then collection.not_muted.starred
-    when 'show_muted'         then collection.muted
-    when 'show_unread'        then collection.not_muted.unread
-    else                           collection.not_muted
+    when 'show_participating' then collection.not_muted.participating.sorted_by_importance
+    when 'show_muted'         then collection.muted.sorted_by_latest_activity
+    else                           collection.not_muted.sorted_by_importance
     end
   end
 
-  def discussion_reader
-    @dr ||= DiscussionReader.for(user: current_user, discussion: @discussion)
+  def collection_for_inbox(collection)
+    collection.not_muted.unread.sorted_by_latest_activity
+  end
+
+  def collection=(value)
+    @discussions = DiscussionWrapper.new_collection user: current_user, discussions: value
   end
 end
--- a/app/extras/queries/visible_discussions.rb
+++ b/app/extras/queries/visible_discussions.rb
@@ -43,6 +43,14 @@ class Queries::VisibleDiscussions < Delegator
     end
   end
 
+  def join_to_starred_motions
+    unless @joined_to_starred_motions
+      join_to_discussion_readers
+      @relation = @relation.joins("LEFT OUTER JOIN motions smo ON smo.discussion_id = discussions.id AND smo.closed_at IS NULL AND dv.starred = true")
+      @joined_to_starred_motions = true
+    end
+  end
+
   def with_active_motions
     join_to_motions
     @relation = @relation.where('mo.id IS NOT NULL')
@@ -69,7 +77,7 @@ class Queries::VisibleDiscussions < Delegator
 
   def muted
     join_to_discussion_readers && join_to_memberships
-    @relation = @relation.where('(dv.volume = :mute) OR (dv.volume IS NULL AND m.volume = :mute) ', 
+    @relation = @relation.where('(dv.volume = :mute) OR (dv.volume IS NULL AND m.volume = :mute) ',
                                 {mute: DiscussionReader.volumes[:mute]})
     self
   end
@@ -86,10 +94,9 @@ class Queries::VisibleDiscussions < Delegator
     self
   end
 
-  def sorted_by_latest_motions
-    @relation = @relation.joined_to_current_motion
-                         .preload(:current_motion, {group: :parent})
-                         .order('motions.closing_at ASC, last_activity_at DESC')
+  def sorted_by_importance
+    join_to_starred_motions && join_to_motions
+    @relation = @relation.order('smo.closing_at ASC, mo.closing_at ASC, dv.starred DESC NULLS LAST, last_activity_at DESC')
     self
   end
 
--- a/app/helpers/development/dashboard_helper.rb
+++ b/app/helpers/development/dashboard_helper.rb
@@ -1,0 +1,59 @@
+module Development::DashboardHelper
+  def starred_proposal_discussion
+    create_discussion!(:starred_proposal_discussion) { |discussion| star!(discussion); add_proposal!(discussion) }
+  end
+
+  def proposal_discussion
+    create_discussion!(:proposal_discussion) { |discussion| add_proposal!(discussion) }
+  end
+
+  def starred_discussion
+    create_discussion!(:starred_discussion) { |discussion| star!(discussion) }
+  end
+
+  def participating_discussion
+    create_discussion!(:participating_discussion) { |discussion| participate!(discussion) }
+  end
+
+  def recent_discussion(group: test_group)
+    create_discussion!(:recent_discussion, group: group)
+  end
+
+  def old_discussion
+    create_discussion!(:old_discussion) { |discussion| discussion.update last_activity_at: 2.years.ago }
+  end
+
+  def muted_discussion
+    create_discussion!(:muted_discussion) { |discussion| mute!(discussion) }
+  end
+
+  private
+
+  def create_discussion!(name, group: test_group, author: patrick)
+    var_name = :"@#{name}"
+    if existing = instance_variable_get(var_name)
+      existing
+    else
+      instance_variable_set(var_name, Discussion.create!(title: name.to_s.humanize, group: group, author: author, private: false).tap do |discussion|
+        yield discussion if block_given?
+      end)
+    end
+  end
+
+  def star!(discussion, user: patrick)
+    DiscussionReader.for(discussion: discussion, user: user).update starred: true
+  end
+
+  def mute!(discussion, user: patrick)
+    DiscussionReader.for(discussion: discussion, user: user).update volume: DiscussionReader.volumes[:mute]
+  end
+
+  def participate!(discussion, user: patrick)
+    DiscussionReader.for(discussion: discussion, user: user).participate!
+  end
+
+  def add_proposal!(discussion, name: 'Test proposal', actor: jennifer)
+    MotionService.create(motion: Motion.new(name: name, closing_at: 3.days.from_now, discussion: discussion), actor: actor)
+  end
+
+end
--- a/lineman/app/components/dashboard_page/dashboard_page_controller.coffee
+++ b/lineman/app/components/dashboard_page/dashboard_page_controller.coffee
@@ -5,60 +5,66 @@ angular.module('loomioApp').controller 'DashboardPageController', ($rootScope, $
 
   Records.votes.fetchMyRecentVotes()
 
-  @perPage = 25
+  @perPage = 50
   @loaded =
     show_all:           0
     show_muted:         0
-    show_proposals:     0
     show_participating: 0
 
+  @views =
+    recent: {}
+    groups: {}
+
   @timeframes =
-    today:     { from: '1 second ago', to: '-10 year ago' }
+    today:     { from: '1 second ago', to: '-10 year ago' } # into the future!
     yesterday: { from: '1 day ago',    to: '1 second ago' }
     thisweek:  { from: '1 week ago',   to: '1 day ago' }
     thismonth: { from: '1 month ago',  to: '1 week ago' }
     older:     { from: '3 month ago',  to: '1 month ago' }
   @timeframeNames = _.map @timeframes, (timeframe, name) -> name
-  @timeframeQueryFor = (name) -> @timeframes[name].view
 
-  @groups = {}
+  @recentViewNames = ['proposals', 'starred', 'today', 'yesterday', 'thisweek', 'thismonth', 'older']
+
   @groupThreadLimit = 5
   @groups = -> CurrentUser.parentGroups()
-  @groupName = (group) -> group.name
-  @groupQueryFor = (group) -> @groups[group.key]
-  @moreForThisGroup = (group) -> @groupQueryFor(group, { filter: @filter }).length() > @groupThreadLimit
+  @moreForThisGroup = (group) -> @views.groups[group.key].length() > @groupThreadLimit
 
   @displayByGroup = ->
-    _.contains ['show_starred', 'show_muted'], @filter
+    _.contains ['show_muted'], @filter
 
-  @updateQueries = ->
-    _.each @groups(), (group) =>
-      @groups[group.key] = ThreadQueryService.groupQuery(group, { filter: @filter })
-    _.each @timeframeNames, (name) =>
-      @timeframes[name].view = ThreadQueryService.timeframeQuery
-        name: name
-        filter: @filter
-        timeframe: @timeframes[name]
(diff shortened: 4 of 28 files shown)


app/controllers/api/discussions_controller.rb
```diff
@@ -3,60 +3,52 @@ class API::DiscussionsController < API::RestfulController
   load_resource only: [:create, :update]
 
   def dashboard
-    instantiate_collection { |collection| filter_collection collection }
-    respond_with_discussions
+    instantiate_collection { |collection| collection_for_dashboard collection }
+    respond_with_collection scope: { visible_in_dashboard: true }, serializer: DiscussionWrapperSerializer, root: 'discussion_wrappers'
```
gdpelican: This is kinda the best part of this update, I think.

Instead of worrying about what's loaded from dashboard and what's not, we attach a 'visible_in_dashboard' flag to the serialized discussion. If it's loaded in the dashboard call, it has 'visible_in_dashboard' set on it. Then, we only display threads in the dashboard which have that flag on it; happy happy.

app/controllers/api/restful_controller.rb
```diff
@@ -3,6 +3,10 @@ class API::RestfulController < API::BaseController
   rescue_from CanCan::AccessDenied                    do |e| respond_with_standard_error e, 403 end
   rescue_from ActionController::UnpermittedParameters do |e| respond_with_standard_error e, 400 end
 
+  def show
```
gdpelican: Finally moved this to the restful_controller; not sure why we didn't do that before. :)

app/helpers/development/dashboard_helper.rb
```diff
@@ -0,0 +1,59 @@
+module Development::DashboardHelper
+  def starred_proposal_discussion
```
gdpelican: Too much DSL? I dunno, I found this to be a nice easy way to make a bunch of discussions I needed.

lineman/app/components/dashboard_page/dashboard_page_controller.coffee
```diff
@@ -5,49 +5,54 @@ angular.module('loomioApp').controller 'DashboardPageController', ($rootScope, $
...
 
-  @updateQueries = ->
-    _.each @groups(), (group) =>
-      @groups[group.key] = ThreadQueryService.groupQuery(group, { filter: @filter })
-    _.each @timeframeNames, (name) =>
-      @timeframes[name].view = ThreadQueryService.timeframeQuery
-        name: name
-        filter: @filter
-        timeframe: @timeframes[name]
+  @updateQueries = =>
```
gdpelican: I've moved these into a simpler 'this.views' structure, instead of a convoluted threadFor method.

This is still a little complex with the ThreadQueryService, but it's definitely better, and you can much better understand the 'what' of what it's doing, if not the 'how' at first glance.

lineman/app/components/dashboard_page/dashboard_page_controller.coffee
```diff
@@ -5,49 +5,54 @@ angular.module('loomioApp').controller 'DashboardPageController', ($rootScope, $
...
   @loadMore = =>
     from = @loaded[@filter]
     @loaded[@filter] = @loaded[@filter] + @perPage
 
-    Records.discussions.fetchDashboard
+    Records.discussions.fetchDashboard(
       filter: @filter
       from:   from
-      per:    @perPage
+      per:    @perPage).then @updateQueries
```
gdpelican: This fixed a dumb bug we were having!

lineman/app/components/inbox_page/inbox_page_controller.coffee
```diff
@@ -4,21 +4,26 @@ angular.module('loomioApp').controller 'InboxPageController', ($scope, $rootScop
   $rootScope.$broadcast('analyticsClearGroup')
 
   @threadLimit = 5
+  @views =
+    groups: {}
+
+  @loading = -> !(CurrentUser.inboxLoaded and CurrentUser.membershipsLoaded)
```
gdpelican: Wait to initialize inbox until we have all of the current user's memberships (ie groups) and the inbox query is complete.

lineman/app/components/thread_preview_collection/thread_preview_collection.haml
```diff
@@ -1,3 +1,3 @@
 .thread-previews
-  .blank{ng-repeat: 'thread in query.threads() track by thread.id | limitTo: limit'}
+  .blank{ng-repeat: 'thread in query.threads() | orderBy:importance track by thread.key | limitTo: limit'}
```
gdpelican: This also fixed a dumb bug we were having (apparantly the ids of discussions weren't being read as unique for some reason... still not sure exactly why.)

lineman/app/services/current_user.coffee
```diff
@@ -4,8 +4,15 @@ angular.module('loomioApp').factory 'CurrentUser', ($rootScope, Records) ->
...
     Records.import(window.Loomio.seedRecords)
+
+    currentUser =  Records.users.find(window.Loomio.currentUserId)
+
     Records.memberships.fetchMyMemberships().then ->
+      currentUser.updateFromJSON membershipsLoaded: true
       $rootScope.$broadcast 'currentUserMembershipsLoaded'
-    currentUser =  Records.users.find(window.Loomio.currentUserId)
+
+    Records.discussions.fetchInbox().then ->
```
gdpelican: Here's a thing I might have thought was crazy before I did it...

Today, the unread count query happens in the navbar directive. That's weird. Next best place? Well, I picked the CurrentUser model, alongside the memberships query. (after all, we're _always_ performing this request when you're logged in, because we always want that unread count to show up immediately) 

I can't help but think there may be a separate place to put this user / app initialization stuff though; it's a bit split between app.coffee and CurrentUser now, and current user feels like it's doing maybe 3 things at the moment, which it should really only be doing `Records.users.find(Loomio.currentUserId)`

lineman/app/services/thread_query_service.coffee
```diff
@@ -52,15 +48,21 @@ angular.module('loomioApp').factory 'ThreadQueryService', (Records) ->
...
-          return false if thread.isMuted() and filter != 'show_muted'
-          return false if thread.readerNotLoaded()
-          switch filter
-            when 'show_all'           then true
-            when 'show_muted'         then thread.isMuted()
-            when 'show_unread'        then thread.isUnread()
-            when 'show_participating' then thread.isParticipating()
-            when 'show_starred'       then thread.isStarred()
-            when 'show_proposals'     then thread.hasActiveProposal()
+
```
gdpelican: Now threads get to run the gauntlet to see if they qualify for each thread collection.

app/controllers/api/discussions_controller.rb
robguthrie: Cool. This is helping me understand how to use scope...

app/controllers/api/discussions_controller.rb
```diff
@@ -3,60 +3,52 @@ class API::DiscussionsController < API::RestfulController
...
+  end
+
+  def inbox
+    instantiate_collection { |collection| collection_for_inbox collection }
+    respond_with_collection scope: { visible_in_inbox: true }, serializer: DiscussionWrapperSerializer, root: 'discussion_wrappers'
```
robguthrie: Hrrmm.. what about if you just passed has_discussion_reader rather than visible_on_inbox ? Are you thinking that we move the "what to display" logic to the server side or is this just a hint to the client about what can be displayed?

app/controllers/api/discussions_controller.rb
```diff
@@ -3,60 +3,52 @@ class API::DiscussionsController < API::RestfulController
...
+    if params[:filter] == 'show_unread'
+      { visible_in_inbox: true }
+    else
+      { visible_in_dashboard: true }
     end
   end
 
-  def discussion_reader
-    @dr ||= DiscussionReader.for(user: current_user, discussion: @discussion)
+  def collection=(value)
```
robguthrie: Hey I ged a bad feeling about using the syntatic sugar like this. It makes it really hard to know that `collection =` is something other than just setting a local variable.

app/helpers/development/dashboard_helper.rb
robguthrie: na it's cool

lineman/app/components/dashboard_page/dashboard_page_controller.coffee
robguthrie: Which one?

gdpelican: Ah, just one I was struggling with where you'd visit inbox, then hit
dashboard, and your threads wouldn't reload correctly because it wasn't
updating the queries after the initial dashboard query came back. (Bug
wasn't in prod, just in this branch.)

lineman/app/services/current_user.coffee
```diff
@@ -4,8 +4,15 @@ angular.module('loomioApp').factory 'CurrentUser', ($rootScope, Records) ->
...
+      currentUser.updateFromJSON inboxLoaded: true
```
robguthrie: Yea, I think we should have an ApplicationState service or something rather than attaching this to the current user

app/controllers/api/discussions_controller.rb
gdpelican: I was more thinking to make the discussion know which query (or queries) it
originated from, so we don't get leakage issues. So, in the record store,
we can differentiate between discussions which were loaded via the thread
page, via notifications, via dashboard, etc.

lineman/app/services/thread_query_service.coffee
```diff
@@ -1,40 +1,37 @@
...
-      threads: ->
-        @filter @view.data()
-      length: ->
-        @threads().length
-      any: ->
-        @length() > 0
+    threadQueryFor = (view, filters, queryType) ->
+      view:       view
+      filter:     createFilter([].concat(filters), queryType)
+      threads: -> @filter @view.data()
```
robguthrie: how does this work? -> will give you the wrong this/@. I would have expected you needed to use => here

lineman/app/services/thread_query_service.coffee
robguthrie: I wonder if this approach is going to be CPU intensive. Happy to proceed with this approach for now though.

app/controllers/api/discussions_controller.rb
robguthrie: I still don't understand why you would get leakage issues... if a discussion matches the filter it should be ok to display, regardless of what query it came from.. and if that isn't the case we're going to have problems when discussions come from other places ie: the message channel...

So I can understand marking discussions which don't have readers, as they can be problematic.. but I think we should design the filters to be able to use discussions from any query if they match.

app/controllers/api/discussions_controller.rb
```diff
@@ -3,60 +3,44 @@ class API::DiscussionsController < API::RestfulController
...
     end
   end
 
-  def discussion_reader
-    @dr ||= DiscussionReader.for(user: current_user, discussion: @discussion)
+  def collection_for_inbox(collection)
+    collection.not_muted.unread.sorted_by_latest_activity
+  end
+
+  def collection=(value)
```
robguthrie: I would like to suggest that you change the name of this method.. I think the syntatic sugar his hiding that there is work going on.

app/extras/queries/visible_discussions.rb
```diff
@@ -86,10 +94,9 @@ def sorted_by_latest_activity
     self
   end
 
-  def sorted_by_latest_motions
-    @relation = @relation.joined_to_current_motion
-                         .preload(:current_motion, {group: :parent})
-                         .order('motions.closing_at ASC, last_activity_at DESC')
+  def sorted_by_importance
+    join_to_starred_motions && join_to_motions
+    @relation = @relation.order('smo.closing_at ASC, mo.closing_at ASC, dv.starred DESC NULLS LAST, last_activity_at DESC')
```
robguthrie: fyi - starred cannot be null

app/extras/queries/visible_discussions.rb
```diff
@@ -43,6 +43,14 @@ def join_to_motions
     end
   end
 
+  def join_to_starred_motions
```
robguthrie: What is a starred motion? ... seems a little peculiar

lineman/app/services/thread_query_service.coffee
robguthrie: I do really wish we could push this stuff into LokiJS it has indexes and stuff just for this kind of thing.

app/controllers/api/discussions_controller.rb
gdpelican: This is overriding the collection= method on RestfulController

app/extras/queries/visible_discussions.rb
gdpelican: It's null when you're left joining on a table and it can't find a left join
though.

app/extras/queries/visible_discussions.rb
gdpelican: Ah, it's a thread with an active proposal and a star, which we want to
prioritize over starred threads and threads with active proposals.

app/controllers/api/discussions_controller.rb
gdpelican: Nope, I think this is begging for trouble. This leakage thing is the reason
we get weird behaviour like when you first load recent threads, it gives
you some threads that are recent, and some threads from a while ago before
it updates, with chunks missing in between. It means that when you 'load
more' recent threads, you can't be sure that you're getting the next 50
records (because you may have some older threads which already exist),
which means after they're loaded, the scroll position leaves you
disoriented (ie, after load, you're not looking at the same threads on the
screen as you were before the load; the user gets totally lost because some
threads are pushed down and others aren't.. and it's like 'okay, why when I
load more did it give me a bunch of threads from July 10th when I was
looking at a bunch of threads from July 4th before?')

Basically I've tried it out the way you're suggesting and found it to be a
pretty bad time.

robguthrie: So, if you view dashboard, then view inbox, you need to wait for all the discussions to be redownloaded with a different variable added to them before they'll show?

gdpelican: Yep, otherwise you get an awkward half-loading screen where it looks like
you only have unread threads in your recent threads list until the rest of
the request comes down.

robguthrie: But avoiding redownloading everything is the reason we moved to a rich client. To me, we may as well just be writing a server side app if we do this.

app/extras/queries/visible_discussions.rb
robguthrie: ohh! right. Cool

app/controllers/api/discussions_controller.rb
robguthrie: I don't think that this leakage problem is really a problem. I think it's fine, and we've not had any feedback from users to say it's a problem.

I would really like to move the client to a place where we just download new stuff, and it's instant to change between pages. I feel like this approach prevents both of those things. I guess we should have a chat?

lineman/app/services/thread_query_service.coffee
```diff
@@ -1,49 +1,40 @@
...
-                     options['filter'] or 'show_unread',
-                     options['limit'] or 5
+    filterQuery: (filter, options = {}) ->
+      threadQueryFor createBaseView(filter, options['queryType'] or 'all')
 
     timeframeQuery: (options = {}) ->
-      threadQueryFor createTimeframeView(options['name'], options['timeframe']),
-                     options['filter'] or 'show_all',
-                     options['limit'] or 100
+      threadQueryFor createTimeframeView(options['name'], options['filter'] or 'show_all', 'timeframe', options['timeframe']['from'], options['timeframe']['to'])
```
gdpelican: I've simplified this class a lot. Now it is much easier to understand.

Now we're finally applying conditions to a loki view directly based on the query type and filters you pass in.

robguthrie: wha hoo! :necktie:

lineman/app/services/thread_query_service.coffee
```diff
@@ -52,15 +43,19 @@ angular.module('loomioApp').factory 'ThreadQueryService', (Records) ->
...
-          return false if thread.isMuted() and filter != 'show_muted'
-          return false if thread.readerNotLoaded()
-          switch filter
-            when 'show_all'           then true
-            when 'show_muted'         then thread.isMuted()
-            when 'show_unread'        then thread.isUnread()
-            when 'show_participating' then thread.isParticipating()
-            when 'show_starred'       then thread.isStarred()
-            when 'show_proposals'     then thread.hasActiveProposal()
+    applyFilters = (view, filters, queryType) ->
```
robguthrie: This looks really great. 

Have you considered using applyFind on the simple properties? I think using Find that means Loki can use indices rather than apply a function to every record.

gdpelican: None of those (oh, except hasActiveProposal, now that I look at it, I can
do that one) are simple, I don't think, they're all reader methods.

It'd be interesting to see an alternate implementation of this where we're
displaying a list readers instead of a list of discussions, since most of
the functionality on dashboard / inbox is around 'what have YOU done with
this thread', ie, the reader.

lineman/app/components/dashboard_page/dashboard_page_controller.coffee
```diff
@@ -5,52 +5,57 @@ angular.module('loomioApp').controller 'DashboardPageController', ($rootScope, $
...
   LoadingService.applyLoadingFunction @, 'loadMore'
 
-  @refresh = ->
+  @refresh = =>
     @updateQueries()
```
robguthrie: Hey James, just looking back over this. It's a really awesome piece of work. 

So I keep wondering when a bug is going to come up with how we update records. 
LokiJS has this `update(obj)` function [docs](http://lokijs.org/#/docs) which it looks like you're supposed to call to tell it to rebuild indexes and views.

So I wonder if you have to call updateQueries() here because we're failing to tell Loki that we've updated a record earlier on. 

Do you feel like that is the case, or is something else going on here?

lineman/app/components/thread_preview_collection/thread_preview_collection.haml
robguthrie: that is really really weird.

lineman/app/components/dashboard_page/dashboard_page_controller.coffee
gdpelican: Perhaps. I might not have my head entirely around when loki should and
shouldn't be updating.

An interesting example of that would be that when I changed it from the
previous implementation to purely views, it wouldn't live update the views
when they changed (ie, I click 'unstar' on a starred thread, and it doesn't
drop out of the starred box on dashboard until I revisit the page)

I manually fixed that one with a broadcast, but you're right that I can't
help but think there's something else that should be going on.

lineman/app/services/thread_query_service.coffee
robguthrie: I agree. Maybe we should be merging readers into discussion records and just having one object on the client?

gdpelican: I think that's right, and I think that might end up resolving our pesky
'tangential' issues as well. Actually I think it just solves everything,
heh.

robguthrie:
@gdpelican Rich and I have smoke tested it.. Its fine to merge and deploy but can you squash it up first?

Thanks for this!

gdpelican:
Also squashed! :hamburger:

PR #2372, opened 2015-08-04
Display username on hover

HSalmon:


HSalmon:
I think this is ready too @robguthrie

PR #2373, opened 2015-08-04
Attempt at fixing load earlier activity bug

HSalmon:
@gdpelican, @robguthrie, not sure if this is a good way to fix this but...

### DIFF
--- a/lineman/app/components/thread_page/activity_card/activity_card.coffee
+++ b/lineman/app/components/thread_page/activity_card/activity_card.coffee
@@ -67,7 +67,8 @@ angular.module('loomioApp').directive 'activityCard', ->
     LoadingService.applyLoadingFunction $scope, 'loadEventsBackwards'
 
     $scope.canLoadBackwards = ->
-      $scope.firstLoadedSequenceId > $scope.discussion.firstSequenceId and
+      $scope.discussion.hasActivity() and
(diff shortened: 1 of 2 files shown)


lineman/app/components/thread_page/activity_card/activity_card.coffee
```diff
@@ -67,7 +67,8 @@ angular.module('loomioApp').directive 'activityCard', ->
     LoadingService.applyLoadingFunction $scope, 'loadEventsBackwards'
 
     $scope.canLoadBackwards = ->
-      $scope.firstLoadedSequenceId > $scope.discussion.firstSequenceId and
+      $scope.discussion.hasActivity() and
+      ($scope.firstLoadedSequenceId > $scope.discussion.firstSequenceId) and
```
gdpelican: @HSalmon the bug here is related to `events_records_interface.coffee#minLoadedSequenceIdByDiscussion` returning Number.MAX_VALUE as a default when it should be 0.

Also, I can't for the life of me remember why we didn't just write that method on the model
```
  discussion.minSequenceId()
```

in discussion_model.coffee:
```
  minEventId: ->
    _.min @events(), (event) -> event.sequenceId
```

gdpelican:
A nice following change to this fix would be adding a message which said 'no activity yet!' to the activity card of threads which did not have any events. :)

PR #2374, opened 2015-08-04
proposal on right for larger screens

robguthrie:
Quick crack at it. Not bad.

image

### DIFF
--- a/lineman/app/components/base.scss
+++ b/lineman/app/components/base.scss
@@ -25,18 +25,3 @@ input[type="file"]{
 strong{
   @include fontHelveticaBold;
 }
-
-.main-container{
-  padding-top: 48px;
-  z-index: 0;
-  min-width: 320px;
-}
-
-.btn {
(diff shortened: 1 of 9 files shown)


AaronThornton00:
Likey, likey (maybe more margin at top of both activity and proposal)

AaronThornton00:
...or less in between proposal and activity

rdbartlett:
Sometimes you just gotta do it.

robguthrie:
One thing I really like about the current design is that the context panel is the same width as the activity panel. Not sure what I'll do about that if anything.

Anther thing I've noticed about this, is how much less vertical space we use when we fold the proposal up like this. Suddenly I feel like my fingers have a lot less work to do just to navigate.

robguthrie:
The more things change, the more they stay the same huh?

AaronThornton00:
I think they call it 'going full circle'

robguthrie:
need to solve lintel

PR #2377, opened 2015-08-05
locales update

mixmix:
@robguthrie


robguthrie:
Thank you @mixmix.

Are you able to keep doing this while you're away?

PR #2378, opened 2015-08-06
Fix load earlier activity bug

HSalmon:


app/models/events/new_discussion.rb
```diff
@@ -1,7 +1,8 @@
 class Events::NewDiscussion < Event
   def self.publish!(discussion)
     event = create!(kind: 'new_discussion',
-                    eventable: discussion)
+                    eventable: discussion,
+                    sequence_id: 0)
```
gdpelican: @robguthrie Can you think of anything dangerous about doing this?

The current bug is as follows:

We have an auto-sequencer set up on event models which are in a 'discussion_id' channel.
We don't add a sequence id if event.discussion is nil
The 'discussion_id' of a 'new_discussion' event is nil, because the discussion id is stored in `eventable_id`

So, the 'new_discussion' event is sequenced, and it's sequenced in the 'discussion_id == nil' channel. Meaning it's equal to the number of discussions created since the sequencer was put in (~40k at this point)

The logic on the client side says something like 'if the minimum loaded event for this discussion has sequence id > discussion.firstSequenceId, show loadPreviousActivity'. This returns true if there is one event, and that event has sequence id ~40,000.

robguthrie: Thanks for this.. However I'm still confused. My reasoning behind not giving new_discussion events a discussion_id was that they don't show up in the discussion thread.

IE: The rule has been that only event which are supposed to show up in the discussion thread are sequenced.

My question is: how is the new_discussion event related to the bug you're fixing?

robguthrie: My concern is that the event sequence for a discussion only makes sense for those records with a discussion_id. So if you need to create a sequence_id here.. then also add discussion_id.

gdpelican: Hm, that feels scary to me but I'm not sure why. I guess we could put it in and smoke test it, but the idea of a whole mess of new events going from `event.discussion.present? == false` to `event.discussion.present? == true` made me afraid.

The new_discussion event is related to the bug I'm fixing because it shows up in the eventsView. When I call $scope.discussion.events(), I get a new_discussion event back. Which would be ok if it had a nil sequenceId (or a sequenceId of 0, which this fix does), but is not okay if it has a sequenceId of 40,000 (and it's the only event), because it then goes "Wow, the first sequenceId is over 9,000! We must have previous content to load, show the load previous content button'

Oh, I should mention somewhere that this fix isn't backwards compatible, we'll need to run an update on existing 'new_discussion' events to get it to fix old threads.

robguthrie: Can we stop new_discussion events from showing up in the eventsView? it's not used by anything is it?

robguthrie: How does it get a sequenceId of 40,000? I'm still confused and happy to chat in person about it.. but I don't understand .. The general idea is that only events with sequence id count towards the discussion thread... if this does not have a sequence id, then how does it sometimes get a very high one?

gdpelican: So the logic for sequencing looks like this:
```
  acts_as_sequenced scope: :discussion_id, column: :sequence_id, skip:
lambda {|e| e.discussion.nil? }
```

but, if I make a 'new_discussion' event:
```
e.discussion_id # => nil
e.discussion # => the new discussion object (because we alias 'discussion'
to 'eventable' in new_discussion.rb)
```

This means it gets sequenced, with a scope of 'nil'
This means it gets put in the same pile with all the other new_discussion
events (maybe other events too, I dunno)
This means it has a really high sequence id, because it's outside the
appropriate channel

So that's the bug, but I'm hesitant to mess with anything too deep in the
event pile (such as making it so that 'new_discussion' events have a
discussion_id, all of a sudden), because I think it'll break stuff.

robguthrie: OH! Thank you. Now I understand. Can we please change the lambda to check for discussion_id.nil? and set all the new_discussion event sequences to 0?

gdpelican:
Sounds goood!

gdpelican:
a) need to not sequence 'new_discussion' events
b) need to not include 'new_discussion' events when calling $scope.discussion.events()

gdpelican:
I reckon this might be good to go.

gdpelican:
Any objections to merging this, @robguthrie? This bug bugs me. :)

robguthrie:
noooo! Thanks team

PR #2379, opened 2015-08-06
remove country select, where is it used? it takes 30megs of memory!

robguthrie:
just want to see if the tests pass.. cant remember where we use this.


robguthrie:
@gdpelican, @HSalmon sorry I've merged this before getting CR. I've found a reasonably good guide on memory profiling and will be making a few patches (hopefully) like this. Let me know if you have concerns.

PR #2380, opened 2015-08-07
Add load more to group page

HSalmon:


robguthrie:
Looks great. @gdpelican any comments?

app/controllers/development_controller.rb
```diff
@@ -274,6 +281,15 @@ def test_discussion
     @test_discussion
   end
 
+  def test_discussions
+    100.times do
+      Discussion.create!(title: Faker::Company.bs,
```
gdpelican: If we end up using this in more than one test, we should look at the active-record-import gem (https://github.com/zdennis/activerecord-import), because it'll start slowing down our e2ez really quickly.

lineman/app/components/group_page/discussions_card/discussions_card.coffee
```diff
@@ -13,7 +14,13 @@ angular.module('loomioApp').directive 'discussionsCard', ->
         from:     $scope.loaded
         per:      $scope.perPage
       $scope.loaded += $scope.perPage
-      Records.discussions.fetchByGroup options
+      Records.discussions.fetchByGroup(options).then (data) ->
+        if data.discussions?
```
gdpelican: This logic confuses me a little, I think because there's a lot of negatives.

What if it was allDiscussionsLoaded, which defaults to false, and is set to true if `(data.discussions or []) < $scope.perPage`?

gdpelican:
:+1:

app/controllers/development_controller.rb
gdpelican: As a side note, I've noticed we're basically duplicating FactoryGirl stuff in the development controller, do you think there might be a way to include those methods here, so this could just be

```
100.times { create(:discussion, group: group }
```
?

robguthrie: Happy to consider improvements to development controller in another branch.
I said to @Hsalmon that I'm happy for it to get shitty then get cleaned up
later

HSalmon:
@robguthrie, @gdpelican, I think this is ready to go now

HSalmon:
@gdpelican, @robguthrie - is this good to go? If so can we merge it?

PR #2384, opened 2015-08-08
Position label counts in the navbar correctly

AaronThornton00:
- This PR repositions badges in navbar absolute relative to the %i instead of the button.
- Also added drop shadow to badge class app wide

https://trello.com/c/fAIz9MAm/566-label-counts-in-the-navbar-aren-t-nested-correctly

before:
screenshot 2015-08-04 21 38 57

after:
screenshot 2015-08-08 12 26 24

### DIFF
--- a/lineman/app/components/navbar/navbar.haml
+++ b/lineman/app/components/navbar/navbar.haml
@@ -7,9 +7,9 @@
     .lmo-navbar__item{ng-class: '{"lmo-navbar__item--selected": selected == "inboxPage"}'}
       %a.btn.lmo-navbar__btn.lmo-navbar__btn-icon.has-badge{href: '/inbox', title: "{{ 'navbar.unread_title' | translate }}", tabindex: 2}
         %i.fa.fa-lg.fa-inbox>
+          %span.badge{ng-show: 'unreadThreadCount() > 0'}
+            {{unreadThreadCount()}}
         %span.lmo-navbar__btn-label{translate: 'navbar.unread'}
-        %span.badge{ng-show: 'unreadThreadCount() > 0'}
-          {{unreadThreadCount()}}
     .lmo-navbar__item{ng-class: '{"lmo-navbar__item--selected": selected == "groupsPage"}'}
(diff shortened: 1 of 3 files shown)


lineman/app/components/navbar/navbar.haml
```diff
@@ -7,9 +7,9 @@
     .lmo-navbar__item{ng-class: '{"lmo-navbar__item--selected": selected == "inboxPage"}'}
       %a.btn.lmo-navbar__btn.lmo-navbar__btn-icon.has-badge{href: '/inbox', title: "{{ 'navbar.unread_title' | translate }}", tabindex: 2}
         %i.fa.fa-lg.fa-inbox>
+          %span.badge{ng-show: 'unreadThreadCount() > 0'}
```
gdpelican: Hm, why this move? 

Not sure it feels right to nest a span inside an <i> (isn't it funny how fontawesome has basically gotten 100% squatters right on the italic html tag?), and it feels like this would behave the same without it, since it'll be positioned against the relative <a> above it?

AaronThornton00: I don't know if this has anything to do with font-awesome per se, seems pretty standard in iconography markup.  The reason this is nested inside the <i> is that we want the badge to be positioned absolutely relative to the icon(which is fixed width).  The button on the other hand is responsive and changes size based on the screen size.  This is pretty common behaviour.  If you were a markup purist you wouldn't be using BEM right?

gdpelican: I don't think markup purity has anything to do with BEM, but yeah that
reasoning makes sense.

AaronThornton00:
Sorry just a frustrated dig at all the class names on one element after refactoring my dropdown arrow PR.

lineman/app/components/navbar/navbar.haml
AaronThornton00: Sorry just a frustrated dig at all the class names on one element after refactoring my dropdown arrow PR.

robguthrie: If you want to sort that out you need to remove the .btn class. It's not
really BEM to mix both utility classes and BEM classes. Making a button
mixin would be the way to go.

PR #2387, opened 2015-08-09
Style feedback card

AaronThornton00:
Hey peeps just a trial with this one.

I know @rdbartlett has already voice some concerns. Which is fair enough, just trying to clear my list. Have a look and close it if you don't think it is an improvement.
(If you give it enough lift it does not look like it is a component of the app)

- More depth to separate it from app
- Rounded corners to fit with app styling
- More emphasys on Feedback
- Simplify text on switch link
- Color a correctly
- Trim unused width so it doesn't get in the way

Before:
screenshot 2015-08-09 15 11 34

After:
screenshot 2015-08-09 15 02 22

### DIFF
--- a/lineman/app/components/mixins.scss
+++ b/lineman/app/components/mixins.scss
@@ -190,16 +190,22 @@ $cardPaddingSize: 15px;
 #turn-off-angular {
   z-index: 100;
   position: fixed;
-  left: 11px;
+  left: 0px;
   bottom: 10px;
-  width: 120px;
+  width: auto;
   text-align: center;
-  padding: 3px;
+  padding: 10px;
   background-color: white;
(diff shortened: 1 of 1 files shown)


rdbartlett:
:+1:

rdbartlett:
Hey @AaronThornton00 I like this - one question: did you forget to commit the changes to client.yml?

I was going to suggest if we're changing the label to "Back to beta", we should keep that language consistent with the "Switch back to old Loomio" on the angular switch-off screen and the "Switch back to the old Loomio" on the angular switch-on screen.

AaronThornton00:
@rdbartlett hey bro feel free to change any of the copy, 'back to beta' was just a concise idea.  You can skin it how you wish.

rdbartlett:
I'm just confused because the screenshot differs from the PR

AaronThornton00:
Yeah sorry, I must have snap shotted the before after making the change to the yml file.
You can just look at master for a true comparison.

PR #2391, opened 2015-08-09
add indexes based on herokus slowest queries report

robguthrie:


gdpelican:
I'd be super interested to see a dump of this slowest queries report, where did you find it?

robguthrie:
https://dashboard.heroku.com/apps/loomio-production/resources

from there you can see heroku postgres queries report. Grab those an throw explain in front of the query and psql will tell you what approaches it takes for each part of the query.

PR #2392, opened 2015-08-10
Ensure subgroups display on groups page

HSalmon:
...even when user is not member of parent group


HSalmon:
@robguthrie, @gdpelican - anything more need to be done for this to get merged?

PR #2394, opened 2015-08-11
Ctrl click to open thread in new window

gdpelican:
I added this to the dashboard; @robguthrie I'm thinking this maybe wants to go into a helper method so we can call it whenever there's a link somewhere though? I was kinda thinking of writing an `lmo-href` directive.


rdbartlett:
Neat! Q's: 

1) does it work the same for Cmd and Ctrl?
2) does setting the target to "_tab" mean each link you ctrl+click will open in its own new tab?

gdpelican:
Once I get my grubby fingers back onto a windows keyboard, I'll test. Off the top of my head I think metaKey just works for the windows key, and will have to add ctrl support. _tab opens a new tab for me, while I was having trouble with _blank. The docs say different, but wanted to go with the thing that works for now.

Obviously this'll require a little X browser / OS testing before it's ready to go.

robguthrie:
Cool.. Thanks @gdpelican. Yes I think we should hold off on this until the technique is confirmed to work on Windows, OSX and Linux.  Also I think it's worth doing lmo-href immediately as we'll need to use this all over the place.

gdpelican:
I don't _really_ like the `threadUrl` method, and would be quite partial to calling something like

```
a{lmo-href:'{{discussion.path()}}'}
```

This is bad because it requires DiscussionModel to know about LmoUrlService, but good because 

- It keeps LmoUrlService out of LmoHref
- It doesn't require us to switch based on record.singular anywhere (ie determine whether the thing I get passed is a group, or a discussion, or a comment... possible but bad OO.)
- It means there's not heaps of duplication of this threadUrl method in a bunch of controllers; put it on the model once and be done with it; models know the path to show themselves.

In rails land, this would be a decorator-level method.. too view-minded for the model, but too object-specific and commonly used for a helper or a view. Thoughts @robguthrie @HSalmon ?

robguthrie:
Thanks James.

I really like routes all being defined in one place. I know it's bad OO to switch on object type (and I'd normally try to avoid it) but it's great for routing - it gives you flexibility to redirect stuff, map one thing to another (eg discussion to /d) as well as fallbacks/catch alls. Being able to see all the routes of a rails app/engine from routes.rb is great.

So I would be keen for lmo-href which in turn uses lmoRoute(object) to return a route for a given string or object (strings beginning with '/' would just pass through unmodified/return themselves)

```
%a{lmo-href: '/pricing'} 
%a{lmo-href: 'discussion'}
```

Also, I hope we can avoid using interpolation when calling the lmo-href directive. It seems unnecessary.

robguthrie:
Also I think `path` is not a good term to use.. as it does not capture that we're also possibly going to return query and fragment parts of a url when routing. but I'm not against it if it makes things easy

gdpelican:
Kay, I'll need to do some fiddly tweaking on the cross browser logic, but the guts of this are looking really good now.

lineman/app/components/thread_preview/thread_preview.haml
```diff
@@ -11,9 +11,8 @@
       %i.fa.fa-times>
       %span{translate: 'volume_levels.mute'}
 
-  %a.thread-preview__link{href: '/d/{{ thread.key }}'}
+  %a.thread-preview__link{lmo-href: '', lmo-href-model: 'thread'}
```
gdpelican: Passing a string sometimes and a model other times is a pretty bad time I reckon, but this is a nice and clear compromise.

To be clear:
This:
```
a{lmo-href: 'dashboard'} # or '/dashboard'
```
will go to /dashboard, whereas this:
```
a{lmo-href: '', lmo-href-model: 'thread'}
```
Will figure out the path of the thread and go there.

If we find we need to pass query parameters at some point (we don't at the moment), we can add a `lmo-href-query` bit to handle that.

gdpelican: I also added support in here so we can go to a specific route within a model path, ie
```
a{lmo-href: 'memberships', lmo-href-model: 'group'}
```

will go to the memberships page of the group.

robguthrie: that's really cool. I agree that overloading lmo-href with string or model is not as nice as lm-href-model.. I wonder if you're overloading it again with this last (cool) feature. May I suggest you consider:

```
%a{lmo-href-model: "[group, 'memberships']"}
```

like how rails does it, for this last case?

robguthrie:
This is a massively good improvement, nice one James.

robguthrie:
Yea.. Looking over the use case examples.. I think having 2 attribute directives is a bit confusing and I really hope you prefer the array notation :)

gdpelican:
How would you feel about

```
a{lmo-href: '', lmo-href-model: 'group', lmo-href-path: 'memberships'}
```
?

robguthrie:
weirder still. Multiple directives feels strange when you're really just trying to specify one url. Are you unhappy with array grouping?

robguthrie:
So I'm strongly of the opinion that you should only have to specify one attribute.

Examples I like:
```
%a{lmo-href: '/someplace'}  -> /someplace
%a{lmo-href-for: 'group'} -> /g/:key
```

I think maybe the array notation is not that versatile.. it would be great to be able to specify query and fragment.. Lets try some more. How to link to add vote on proposal? (real example from email, might one day be used from within app..)


```
%a{lmo-href-for: "{model: proposal, path: 'votes/new', query: 'position=yes'"} -> /m/31f5Hluc/votes/new?position=yes
```

gdpelican:
It's not multiple directives though, it's multiple attributes. We already use a similar pattern with the translation package, and I'm pretty certain we're going to have a bad time if we try to cram passing an eval expression and a string into a singular attribute.

rdbartlett:
wow you people are good at programming

robguthrie:
@gdpelican that last commit looks awesome. Are you happy with it?

gdpelican:
Not yet, super broken still, but yes feels good so far.

gdpelican:
Sigh. This approach isn't working for me.

```
%a{lmo-href-for: "{ model: group, path: 'memberships'}"}
```

Attempting to do that gets me a digest iteration infinite loop, I think because passing the model through in an object like that is triggering a digest for some reason, which triggers the haml view, which triggers the lmo-href-for controller, etc.

Anyway, I'm just about out of my timebox on this and ready to just go with a solution that works. We can circle back around to this later.

robguthrie:
Oh no! Good luck. do what you gotta do!

gdpelican:
I think this is good to go @robguthrie

robguthrie:
Hey I would be ok with this.. but I have another naming suggestion if it's not too late.

Rather that lmo-href-path about about lmo-href-action then it would not seem like it needs a leading '/' (which is weird to me) and would sound more routery.

But I think you've done are really great job and happy either way.

gdpelican:
action's good, and the leading slash is superfluous at this point I think.

robguthrie:
Need anything from me on this @gdpelican ? Happy to smoke test

gdpelican:
Actually, the e2ez passing is good enough to me. Let's merge it!

PR #2398, opened 2015-08-12
Add ability to sign out of angular interface

HSalmon:


HSalmon:
@robguthrie, @gdpelican - is this ready to go?

lineman/app/components/navbar/navbar_user_options.coffee
```diff
@@ -3,5 +3,9 @@ angular.module('loomioApp').directive 'navbarUserOptions', ->
   restrict: 'E'
   templateUrl: 'generated/components/navbar/navbar_user_options.html'
   replace: true
-  controller: ($scope, CurrentUser) ->
+  controller: ($scope, CurrentUser, Records, $http) ->
```
gdpelican: No need for Records here, I don't think.

HMM the $http service.. is there a way we can move this call into the RestfulClient, so we don't have the '/api/v1' endpoint floating around in more than one place? We've done a really great job of keeping our api calls threaded through a single class so far.

gdpelican: A SessionRecordsInterface _might_ not be necessary (although it might if we decide to add log in to this thing)

But for now, I reckon something like

```
@sessionClient = new RestfulClient('sessions')
@sessionClient.destroy().then -> ...
```

would do us for now.

gdpelican: Also, we in other places inject the $window object instead of using window directly, for testability reasons. 

Injecting the window object means that in our unit tests we can say 'inject a $window instance which has a Loomio.currentUserId variable on it', and not pollute the global namespace for the other tests. (it may not apply here, but I'd rather not give a bad example for future developers)

gdpelican:
Left a comment about the signout line, but I smoke tested it and it works great! Thanks @HSalmon

HSalmon:
@gdpelican, @robguthrie, I think I've implemented the feedback - anything else required?

gdpelican:
Looks great!

PR #2400, opened 2015-08-12
Header styles first pass

HSalmon:
First attempt at Markdown header styles...@rdbartlett, @robguthrie

### DIFF
--- a/lineman/app/components/utilities.scss
+++ b/lineman/app/components/utilities.scss
@@ -229,8 +229,22 @@ label.label-with-details{
   width: 75%
 }
 
-.lmo-markdown-wrapper img {
-  max-width: 100%;
+.lmo-markdown-wrapper {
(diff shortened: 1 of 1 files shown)


HSalmon:
screen shot 2015-08-13 at 11 12 29 am

rdbartlett:
FYI Rob, Hannah and I had a chat about this today. I reckon this PR looks good, and it opened up a few questions for me for a further job:

* how it looks - the H1 in the user content is currently the same font as the page H1, H2 is a bit smaller and then H3 and H4 are the same size as the body text. Do we like that?
* namespacing: is this the right way to structure the CSS?
* markup for accessibility: the page title is H1, the card heading is H2, so does that mean the highest heading level available for user content should be H3?

robguthrie:
That looks great! Happy to merge.

PR #2402, opened 2015-08-13
increase group page cover photo size to 320px

rdbartlett:
Merging this step now will make the following design + dev steps a lot easier.

image

Still looks bung on mobile but I'm most of the way to figuring out a nice solution for that.


robguthrie:
super! merge when you like

PR #2406, opened 2015-08-13
Serialize parent author name with comment

gdpelican:
Remove empty comment authors when parent is not loaded


gdpelican:
Hey this is much easier than duelling with AMS.

robguthrie:
this is simple genius

PR #2407, opened 2015-08-13
CMD + Enter submits discussion, comment, vote, proposal forms

gdpelican:


lineman/app/services/key_event_service.coffee
```diff
@@ -12,15 +12,21 @@ angular.module('loomioApp').factory 'KeyEventService', ($rootScope) ->
...
+        if shouldExecute(active, originalEvent)
           angularEvent.preventDefault() and originalEvent.preventDefault()
           execute(active)
 
-    defaultShouldExecute: (active = {}) ->
-      !_.contains ['INPUT', 'TEXTAREA', 'SELECT'], active.nodeName
+    defaultShouldExecute: (active = {}, event = {}) ->
+      !event.ctrlKey and !event.altKey and !_.contains(['INPUT', 'TEXTAREA', 'SELECT'], active.nodeName)
+
+    submitOnEnter: (scope) ->
```
gdpelican: Hey, so the way this works:

- For a form, mark some fields as lmo-primary-form-input
- In the form controller, mark the form as accepting acceptOnEnter events. (A better iteration of this might apply it to all forms, possibly through the FormService)
- User goes to the form, fills it out, hits cmd/ctrl+enter
- If the element in focus is a) a lmo-primary-form-input field, and b) is within the scope of the form, call form.submit (this is what's great about generalizing the name of the submit action)

I applied to vote, proposal, discussion, and comment forms, but it'll be easy to expand to more if we need it.

robguthrie: wow. that's amazing. you are the best!!

rdbartlett:
Hey this sounds awesome! but this doesn't work for me (Chrome + OSX). Has it been verified in production?

gdpelican:
Yeah, although we only rolled it out for certain forms. Creating / editing
discussion and creating a comment, I think; maybe one or two others.

PR #2408, opened 2015-08-13
Don't overflow lintel header

gdpelican:
Before:
screen shot 2015-08-13 at 10 42 05 am

After:
screen shot 2015-08-13 at 10 41 56 am


robguthrie:
√

PR #2410, opened 2015-08-13
Add alert informing user of invalid filename

HSalmon:


HSalmon:
@robguthrie, @gdpelican and I paired - this seemed like the only thing to do at this stage

robguthrie:
Thing is.. I wonder if the medicine is worse than the illness. I think that the majority of files that have already been uploaded with chars that would not be allowed by this validation don't have any problems. We might be bothering users unnecessarily?

gdpelican:
I had that thought too. 
It was pretty amazing how bad the jquery-fileupload-rails gem was, documentation / sourcecode diving wise.

robguthrie:
I'm really sorry guys!

HSalmon:
We could just leave it - it's fixed in 1.0 and that's where people will be soon, right? I haven't had any enquiries about file uploads recently and maybe if I do get one I could just ask them to rename their file..? Not sure...I want us to help users but if this is more of a hindrance for them then maybe we should leave it

gdpelican:
I think leaving this as is is likely the best pick for now; we're close enough to 1.0 that if we get more complaints about it, we can say 'try the new interface, it works there!'

PR #2412, opened 2015-08-14
Slack webhook

gdpelican:
Necromancy!!! :skull:

Anyway, yeah I played with this a little over the weekend, and am pretty happy with this implementation. It was bugging me a bit that we keep talking about this thing, and that it was only a couple hours of work away from a useable solution.

Love to hear thoughts, and ideally get it merged in and give it a hoon internally at some point; it's got *super* low interface with the existing codebase (it's 100% additions, and the only real changes to existing codebits is the after_create hook in event.rb -- probably a future gem, actually.)

How it works:
We'll manually create a webhook object (this would be a premium feature, after all, amiright?), which has:
  - a **webhook url** (we get this from slack, it's where to send the payload.)
  - a **hookable** (a polymorphic discussion / group), which is the 'wrapper' for which events to send out (if your hookable is a discussion, it's all events in that discussion, if
  - some **event_kinds**, which specifies which event kinds to trigger the webhook on (currently available are most of the motion ones: new_motion, motion_closing_soon, motion_outcome_created/updated. It's pretty easy to add more.)
  - A **kind**... this only takes 'slack' at the moment, but we can send payloads to other things too.

This webhook object gets published on the after_create hook of an event, through a WebhookSerializer, which delivers a payload to the given uri. You can see some of the stuff I was sending through in the #loomio-webhook-test channel.

(PS the vote / view it on loomio buttons work with 1.0... so cool!)

Sorry for the long commit history. Have to rebase or cherrypick that away somehow.

@robguthrie @rdbartlett @HSalmon

screen shot 2015-08-17 at 4 41 07 pm


app/models/concerns/routing.rb
```diff
@@ -0,0 +1,10 @@
+module Routing
```
gdpelican: To allow us to use `discussion_url` without manually passing a host in the serializer

app/models/event.rb
```diff
@@ -36,6 +38,13 @@ def publish_message
     MessageChannelService.publish_event(self)
   end
 
+  def publish_to_webhooks
+    group = self.discussion.group
```
gdpelican: Publish to the webhooks associated with
- The eventable's discussion
- The eventable's discussion's group
- The eventable's discussion's group's parent group

app/serializers/webhook_serializer.rb
```diff
@@ -0,0 +1,3 @@
+class WebhookSerializer < ActiveModel::Serializer
+  attributes :text, :username, :attachments, :icon_url
```
gdpelican: We might need to have different serializers for different targets (hipchat, trello, whatever), but since it's just slack at the moment the one will do I think

app/services/webhook_service.rb
```diff
@@ -0,0 +1,22 @@
+require 'rest-client'
+
+class WebhookService
+
+  def self.publish!(webhook:, event:)
+    return false unless webhook.event_types.include? event.kind
```
gdpelican: Only publish if this event's kind is in the list of event kinds for this webhook.

gdpelican:
Hey @robguthrie I'd love to either have a second set of eyes on this, or just merge it and set up a webhook for the loomio channel to give it a go.

I think the risk of it breaking anything else in the app is pretty low.

robguthrie:
Only two comments..

It seems like rest-client is bringing in yet another Gem. I wonder if httpparty could be used, seeing as we already have it. But I don't mind too much if you're feeling attached to rest-client.

Can we move the external network requests to a delayed job? Having a network request in the same process as saving records etc is really fragile. If you can find a nice place to put a delayed job ActiveJob call in here then I'd be happy.

gdpelican:
Ah, yep, happy to use httparty.

I was under the impression when I first wrote this, for some reason, that we were kicking event creation off to a side job already, but I see that's not the case now. I'm happy to boot webhookin' off to a separate process, since it's non-essential.

robguthrie:
That's great.

PR #2417, opened 2015-08-18
Fuse discussions and readers

gdpelican:
If these tests pass I'll eat my :tophat:


app/serializers/discussion_serializer.rb
```diff
@@ -36,4 +58,16 @@ def filter(keys)
...
+  def scope
+    super || {}
+  end
+
+  def reader
+    @reader ||= DiscussionReader.for(user: scope[:user], discussion: object)
+  end
+
+  def has_valid_reader?
+    @has_valid_reader ||= reader.present? && reader.user.present?
```
robguthrie: why the check for user?

app/controllers/api/discussions_controller.rb
```diff
@@ -40,7 +46,8 @@ def collection_for_inbox(collection)
     collection.not_muted.unread.sorted_by_latest_activity
   end
 
-  def collection=(value)
-    @discussions = DiscussionWrapper.new_collection user: current_user, discussions: value
+  def discussion_event
```
robguthrie: Whats this about?

gdpelican: It's copied over from the old DiscussionReaderController, it's saying 'give me the event for the discussion with the given sequence id', if no sequence id is given, we get the discussion instead. It's used in the mark_as_read action

app/serializers/discussion_serializer.rb
gdpelican: We check for the user because we're getting the user from scope[:user], which is nil unless we specify it explicitly.

This probably won't stay exactly like this; I'd rather have it default to 'include reader fields' rather than exclude them by default, but we can't use 'current_user' in the serializer, so we have to pass it in via the scope.

gdpelican:
Gonna reopen this against master

PR #2419, opened 2015-08-19
Slack webhook upgrades

gdpelican:
- Add links to proposal and discussion titles
- Don't send payload for subgroups which aren't visible to parent group


robguthrie:
Lovely. Squishy squishy?

gdpelican:
Squashed!

PR #2420, opened 2015-08-20
Tighten up mentioning matching

robguthrie:
I've been frustrated that current angular mentioning matches strings at any point. A quick recap of how we do this in beta and how it works on other sites tells me that we only want to match against the beginnings of words so I've gone and done a wee update so that is the case for us. 

EG when searching for 'rob'
* 'robguthrie' should be true
* 'emrob guthrie' should be false (but it is currently a match without this PR)
* 'james robinson' should be true

Sorry I couldn't help myself.


robguthrie:
I almost tried to replace mentio.. then I realised I was being really crazy

PR #2421, opened 2015-08-20
Public api endpoints

gdpelican:
Expose public content on our index actions for groups, discussions, and proposals


robguthrie:
Kind of confused as to why you did this. I think it looks nice but we need to have a conversation about if we want to include this. 

My main concern is that we want to update the group settings page to be about sharing a link. Enabling sharing a link means technically your group is public.. but really it's only visible to anyone with link. I don't want anyone to be able to automatically pull "public" groups until we work out the UX around that.

robguthrie:
Does that makes sense? I'm hesitant about adding stuff that will make it harder if we want to make this slight but crucial change to how we represent privacy settings.

We'll be doing the design (group settings, invitations) to give the understanding next week so I'm happy to look at it then.

gdpelican:
Closing this for now; will circle back after we have a more full conversation about it.

PR #2422, opened 2015-08-20
Fuse discussions and readers

gdpelican:


lineman/app/models/discussion_model.coffee
```diff
@@ -118,3 +100,19 @@ angular.module('loomioApp').factory 'DiscussionModel', (BaseModel) ->
     maxLoadedSequenceId: ->
       item = _.max @events(), (event) -> event.sequenceId or 0
       item.sequenceId
+
+    changeVolume: (volume) ->
```
robguthrie: The permissions for update discussion are really different to those for set volume. For instance I can't update all the discussions I can read. I reckon we'll need a special endpoint for setting volume.

robguthrie: And then DiscussionService.update does not get more complex

spec/services/discussion_service_spec.rb
```diff
@@ -112,6 +110,13 @@
                                  params: discussion_params,
                                  actor: user
       end
+
+      it 'can save reader attributes' do
```
robguthrie: Yea.. I really think this feels weird and we should not be saving reader attributes inside the DiscussionService.update method.  Let's make them their own Api calls

lineman/app/models/discussion_model.coffee
gdpelican: Cool, I've made this change.

app/services/discussion_service.rb
```diff
@@ -55,7 +55,25 @@ def self.update(discussion:, params:, actor:)
...
 
     ThreadSearchService.index! discussion.id if update_search_vector
-    DiscussionReader.for(discussion: discussion, user: actor).set_volume_as_required!
+    DiscussionReader.for(user: actor, discussion: discussion).set_volume_as_required!
     event
   end
+
+  def self.update_reader(discussion:, params:, actor:)
+    reader = DiscussionReader.for(discussion: discussion, user: actor)
+    actor.ability.authorize! :show, discussion
```
gdpelican: I thought this wanted to be `authorize :update, reader`, but that doesn't really work because it only checks that the reader's user is equal to the permission user.. which is inherently true here.

I could throw that check in here as well, but what I really want to be sure of is that the user can see the discussion s/he's trying to read.

robguthrie: I feel like REST gives us a poor API when it looks like this. Exposing the discussion_reader like this feels like we're not abstracting the implementation from the behaviour.

Rather than
/api/v1/discussions/update_reader?{discussion_reader: {starred: val, volume: val}

I'd really like to see the discussion's api documentation look more like this:
/api/v1/discussions/:id/mark_as_read
/api/v1/discussions/:id/set_volume?volume=mute /api/v1/discussions/:id/star
/api/v1/discussions/:id/unstar

Then when we change the way we implement stars our API does not need to change.

robguthrie:
So can I merge this now? It looks great.

I've tried to check that authorization is still done properly for star etc, so people can't star a record to get given a copy of it in response.. and it looks like you're doing that right.

gdpelican:
Yeah let's merge it! There's specs around the star permissions, so that
should be good.

PR #2423, opened 2015-08-20
Add activity items missing in angular

HSalmon:


HSalmon:
@robguthrie, @gdpelican, could definitely do with a review I'd say!

app/services/motion_service.rb
```diff
@@ -23,19 +23,25 @@ def self.update(motion:, params:, actor:)
...
       end
 
       if motion.closing_at_changed?
-        Events::MotionCloseDateEdited.publish!(motion, actor)
+        event = Events::MotionCloseDateEdited.publish!(motion, actor)
       end
 
       motion.save
       ThreadSearchService.index! motion.discussion_id if sync_search_vector
+      if event
```
gdpelican: This is super confusing. I think a better way to handle it is returning an event if it exists, and nil otherwise (then the restful controller will default to responding with the collection)

```
return event.reload if event
```

Also why do we need to reload?

lineman/app/components/thread_page/activity_card/discussion_title_edited.haml
```diff
@@ -0,0 +1,10 @@
+%article.thread-item{ng-controller: 'DiscussionTitleEditedItemController', aria-labelledby: 'discussion-{{discussion.id}}'}
```
gdpelican: At some point we want to go in and refactor this so there's not so much duplication of markup... also I'd be amazed if we actually needed to use ng-controller here

A thing for another time, though.

app/serializers/event_serializer.rb
```diff
@@ -11,7 +11,7 @@ class EventSerializer < ActiveModel::Serializer
   has_one :vote, serializer: VoteSerializer
 
   def actor
-    object.eventable.try(:user)
+    object.user || object.eventable.try(:user)
```
gdpelican: Erg, the wonkiness of the events system is really killing us here; why do we need to set the discussion_id on motion events? Why is the actor sometimes on the event directly and sometimes on the eventable? Why is event.discussion ok sometimes, and Event.where(discussion_id: eventable.discussion_id) necessary in others?

I think streamlining this is probably we need to take a look at sometime in the medium-term future, since it's such a central part of the app (and will become more central as we move towards plugin architecture / app-to-app comms)

HSalmon: @gdpelican, @robguthrie and I are gonna have a stab at refactoring this whole system tomorrow morning, so stay tuned!

gdpelican:
Hey I just looked through my comments on this and realized I was likely being a bit harsh due to a long and arduous travel day, so sorry for bad tonevibes; this is still great work.

HSalmon:
@gdpelican, are you up for a review of this branch? It's a big 'un

app/models/discussion.rb
```diff
@@ -1,6 +1,7 @@
 class Discussion < ActiveRecord::Base
   PER_PAGE = 50
   SALIENT_ITEM_KINDS = %w[new_comment new_motion new_vote motion_outcome_created]
+  SAFE_ITEM_KINDS = %w[new_comment new_motion new_vote motion_outcome_created motion_outcome_updated]
```
gdpelican: What's 'safe' about these? Are they items which appear in the activity feed?

app/models/record_edit.rb
```diff
@@ -0,0 +1,27 @@
+class RecordEdit < ActiveRecord::Base
+
+  belongs_to :record, polymorphic: true
+  belongs_to :author, class_name: 'User'
+  belongs_to :user, foreign_key: 'author_id' #duplicate author relationship for eager loading via discussion.items
+  has_one    :event, as: :eventable
+
+  validates_presence_of :author, :record
+
+  def self.capture(record, author)
```
gdpelican: Could you explain what this is for and why PaperTrail won't do it for us?

HSalmon: Because the client needs to know about edits

app/models/discussion.rb
HSalmon: These are the activity items that will appear in beta

app/models/record_edit.rb
gdpelican: I dunno, I'm skeptical of putting this into the DB twice. I'd much rather have a VersionSerializer and pull out paper trail with our own implementation when it comes to that, rather than duplicate our edits and have a whole nother table to deal with; we already have all this info saved.

app/services/discussion_service.rb
```diff
@@ -40,22 +40,14 @@ def self.update(discussion:, params:, actor:)
...
-    event = true
-    if discussion.title_changed?
-      event = Events::DiscussionTitleEdited.publish!(discussion, actor)
-    end
-
-    if discussion.description_changed?
-      event = Events::DiscussionDescriptionEdited.publish!(discussion, actor)
-    end
-
+    return discussion if discussion.changed == ['uses_markdown']
```
gdpelican: Hey @robguthrie, I want to name the rules around what we return from services for update (and in general, really)

I think I prefer this:

- true (or nil) when it's successful and there is no event
- the event, if one has been generated
- raise an error if it fails (we want to know about it, no silent failures)
OR
- return false if it fails (this way we don't have to rescue in the existing discussions controller)

Does that sound reasonable?
Then this turns into
```
return true if discussion.changed.reject { |c| c == 'uses_markdown' }.empty?
```

app/services/discussion_service.rb
```diff
@@ -40,22 +40,14 @@ def self.update(discussion:, params:, actor:)
...
+    return discussion unless discussion.changed?
+    
+    event = Events::DiscussionEdited.publish!(discussion, actor)
```
gdpelican: This is nitpicky, but I'd rather save the discussion before we publish the event, just in case something goes wrong (event though we've already verified the discussion is valid)

app/services/motion_service.rb
```diff
@@ -19,26 +19,12 @@ def self.update(motion:, params:, actor:)
...
-        Events::MotionCloseDateEdited.publish!(motion, actor)
-      end
+    return false unless motion.valid?
 
-      motion.save
-      ThreadSearchService.index! motion.discussion_id if sync_search_vector
-    else
-      false
-    end
+    event = Events::MotionEdited.publish!(motion, actor)
```
gdpelican: Same nitpick as with discussions above

db/migrate/20150824025314_add_author_id_to_record_edits.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthorIdToRecordEdits < ActiveRecord::Migration
```
gdpelican: Unless both of you have run these migrations and this'll be a pain in the butt to reconcile, I'd rather just rollback and add this column to the create_record_edits migration, rather than adding a new one in a separate file.

spec/factories.rb
```diff
@@ -1,4 +1,7 @@
-FactoryGirl.define do
+FactoryGirl.define do  factory :record_edit do
```
gdpelican: Whitespace issues here?

app/models/record_edit.rb
gdpelican: PaperTrail supports a 'changeset' hash property with a minor migration out of the box, and I don't think it's a great pick at this point to rip it out. Was there an attempt at using the original versioning stuff there that didn't work for some reason?

app/models/record_edit.rb
```diff
@@ -0,0 +1,27 @@
...
+    record_edit.previous_values = record.changed_attributes
+    record_edit.new_values = {}
+    record_edit.previous_values.each_pair do |k, v|
+      record_edit.new_values[k] = record.send(k)
+    end
+    record_edit
+  end
+
+  def self.capture!(record, author)
+    record_edit = self.capture(record, author)
```
gdpelican: Here's a cool trick I learned recently:

```
self.capture(record, author).tap(&:save!)
```

app/services/discussion_service.rb
robguthrie: Yea I've been holding rules in my head but it's great to discuss them for real.

Service methods return:
true, record or event if all is well.
false if record validation fails.
and they raise an exception if you do no have permission to perform the action.

PR #2425, opened 2015-08-20
Add confirm discard changes/Upgrade record store

robguthrie:
e2es are almost passing. Should get this done before monday

Still to do:
e2es and smoke testing for confirm cancellation on group, discussion, proposal, vote, comment forms
enable for comment form
bug: proposal form is always modified


robguthrie:
Not ready for CR but if you want to have a look that's fine.

PR #2428, opened 2015-08-21
remove border and add padding to user profile btn in nav

AaronThornton00:
I hope this PR doesn't get in anybodies way but I would love it if someone could merge this as it bugs me daily.

Finally realised what the issue was that has been bugging me.  The navbar has a border bottom that it doesn't need. Removing this adds the crispness back to the shadow too.

This PR:
- removes border bottom from navbar 
- pads out user profile button to full height of navbar

Before:
screenshot 2015-08-22 10 24 02

After:
screenshot 2015-08-22 10 23 28

### DIFF
--- a/lineman/app/components/navbar/navbar.scss
+++ b/lineman/app/components/navbar/navbar.scss
@@ -2,7 +2,6 @@
   @include box_shadow(2);
   z-index: 1000;
   background: white;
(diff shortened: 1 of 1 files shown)


rdbartlett:
holla!

AaronThornton00:
Yusss!!

PR #2429, opened 2015-08-24
Inline translations WIP

gdpelican:
This is a result of airplane faffing about; the best kind of faffing about.

This totally works for comments at the moment, and is fabulously great.

I'm going to stop working on it for now tho, unless some group gives us some feedback that they want it in 1.0 (pricing and bugfixing is more important, but this will be a nice thing to knock off at some point.)


gdpelican:
Closing, but this'll be a good one when it gets yanked up again.

PR #2434, opened 2015-08-26
fix proposal closing at, also add motion_edited, discussion_edited activity items

robguthrie:
its a big merge of all the work. it's been reviewed by everyone mostly


robguthrie:
@gdpelican the papertrail version stuff come out nicely. thanks for your help!

lineman/app/components/thread_page/activity_card/motion_edited_item_controller.coffee
```diff
@@ -0,0 +1,11 @@
+angular.module('loomioApp').controller 'MotionEditedItemController', ($scope) ->
+  $scope.proposal = $scope.event.version().proposal()
+  version = $scope.event.version()
+  $scope.proposalName =
+    if version.attributeEdited('name')
+      version.changes.name[1]
```
gdpelican: Hmm.. I assume we do this because the object changes become serialized out as `["old_value", "new_value"]`?

I wonder if we actually need the old value in there, and if it might be better to just get the new value in the serializer, like

```
object.object_changes.inject({}) do |result, (key, value) |
  result[key] = value[1]
  result
end
```

(erg, maybe not... why are simple hash transformations so funky?)

lineman/app/models/proposal_model.coffee
```diff
@@ -88,3 +88,19 @@ angular.module('loomioApp').factory 'ProposalModel', (BaseModel, AppConfig) ->
         _.difference(@group().members(), @voters())
       else
         @recordStore.users.find(_.pluck(@didNotVotes(), 'userId'))
+
+    createOutcome: =>
+      @remote.postMember @id, "create_outcome",
```
gdpelican: Hey I feel like I'd like this to read

```
@remote.postMember @id, 'create_outcome', @serialize('outcome') # or @serialize(['outcome'])
```

and let serialize take a list of fields to serialize (defaulting to everything). Thoughts?

gdpelican:
Looks pretty great to me! Just need to get those undecided member specs passing.

Great work!

lineman/app/components/thread_page/activity_card/motion_edited_item_controller.coffee
robguthrie: I've been asking myself that ('Why are hash transforms so hard?') a lot recently.

Yea.. that's a good idea.. might do that.

lineman/app/models/proposal_model.coffee
robguthrie: Cool. I'll make a note of that and let you know when I add it to record store

PR #2437, opened 2015-08-27
sneaky tweaky

rdbartlett:


robguthrie:
Thanks Richard!

app/assets/stylesheets/pricing_table.scss
```diff
@@ -9,7 +9,7 @@
 }
 
 .pricing-table__option {
-  background-color: $pages-background;
+  background-color: #e2e1de;
```
gdpelican: Er, are we sure we're not going to use these colors again? Otherwise we should name em in `unstructured/_settings`

PR #2438, opened 2015-08-27
tweak AAARRR

rdbartlett:
Hey @robguthrie I have just made a little tweak to make the AAARRR table more valuable. I know it is hacky, but it is cheap and useful. Soon we will be pulling this straight out of Chargify but this makes me happy for now :)

image


robguthrie:
magnifique!

PR #2439, opened 2015-08-27
A11y fixes

HSalmon:
https://trello.com/c/2Z2W2suw/621-a11y-remove-extra-main-region-on-group-page

https://trello.com/c/y29Rcm8i/648-3-a11y-pass-on-stars

### DIFF
--- a/lineman/app/models/discussion_model.coffee
+++ b/lineman/app/models/discussion_model.coffee
@@ -109,7 +109,7 @@ angular.module('loomioApp').factory 'DiscussionModel', (BaseModel, AppConfig) ->
       @remote.patchMember @keyOrId(), 'set_volume', { volume: volume }
 
     toggleStar: ->
-      @remote.patchMember @keyOrId(), if @starred then 'unstar' else 'star'
+      @remote.patchMember @keyOrId(), if @starred then 'star' else 'unstar'
 
     markAsRead: (sequenceId) ->
       if isNaN(sequenceId)
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -570,6 +570,9 @@ en:
       thismonth: 'This month'
       older: 'Older than a month'
 
+  star_toggle:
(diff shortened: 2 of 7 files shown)


lineman/app/models/discussion_model.coffee
```diff
@@ -109,7 +109,7 @@ angular.module('loomioApp').factory 'DiscussionModel', (BaseModel, AppConfig) ->
       @remote.patchMember @keyOrId(), 'set_volume', { volume: volume }
 
     toggleStar: ->
-      @remote.patchMember @keyOrId(), if @starred then 'unstar' else 'star'
```
gdpelican: Oof, I don't like doing it this way; it makes the toggleStar method counterintuitive and means we can't manually call it without first setting the opposite star value. I assume it's doing this because ng-change is firing the model change before this method, could we try something in the ng-model-options perhaps to change this?

```
input{ng-model: 'thread.starred', ng-model-options: '{updateOn: "blur"}'}
```

If blur doesn't work, I wonder if click, change, or mouseup might.

rdbartlett:
Wonderful! Hannah are you happy to start an "updates" thread in the
Accessibility group to notify people about these changes once they are
deployed?

lineman/app/models/discussion_model.coffee
gdpelican: Oh, would you mind tearing out the `toggle` method in the star_toggle_controller as well? I don't think we need it anymore (or the $scope.$on 'starToggled' call in dashboard page controller.. I believe the new angular record store updates have fixed the bug that those were covering)

gdpelican: It'd also be interesting to see what 'debounce' did here.. in a perfect world it would provide a little buffer time so that the web request can fire off before we update the model locally.

```
input{ng-model: 'thread.starred', ng-model-options: '{debounce: 100}'}
```

PR #2440, opened 2015-08-27
Marking recently created items as read

gdpelican:


gdpelican:
This is ready for a look @robguthrie @HSalmon; I feel pretty good about it, but still need to do a smoke test (/ write an e2e maybe)

robguthrie:
Wow. This looks really nice. The only quibble I have is the + 1.second .. Can't we use the exact created at?

gdpelican:
Hm, I couldn't get that to work, but looking at it again I'm going to try going in and tweaking the `viewed!

gdpelican:
Okay, I did a little refactor of the discussion reader stuff within events, which were a bit all over the place.

I ended up writing a `author_thread_item!` method on discussion reader, which encapsulates participating in a thread, viewing the item you just creating, and setting the volume to loud if necessary (all things we've been doing in chunks, sometimes in the event#publish method, sometimes in the service.)

robguthrie:
This is wonderful work. Please squash and we can deploy!

PR #2444, opened 2015-08-28
I can't believe it's not button (from an accessibility standpoint)

robguthrie:


gdpelican:
Whoa...

I would have really appreciated the chance to have a look at this before it got merged in. That toggle star method needs looking at, at the very least.

robguthrie:
Arrr sorry James I had the impression you'd already been over it. I took your input about toggleStar and changed it to saveStar.


Happy to take any suggestions now though.

lineman/app/filters/smart_time.coffee
```diff
@@ -0,0 +1,25 @@
+angular.module('loomioApp').filter 'smartTime', ->
```
robguthrie: this is not used... im a fool for comitting it

PR #2445, opened 2015-08-28
simplify started by string on proposal card

robguthrie:
might wanna do this in more places.


robguthrie:
this is not a real good idea

PR #2448, opened 2015-08-29
locales update

mixmix:
@robguthrie notice there are a couple of deletion type changes in this one. 

can you just confirm this is correct before merging please :


config/locales/client.de.yml
```diff
@@ -234,11 +234,6 @@ de:
   proposal_position_block: Veto
   group_welcome_modal:
     heading: Willkommen in Deiner neuen Gruppe
```
mixmix: e.g. here

robguthrie:
Mix I'm not totally sure what you're asking.. but we are making a lot of translation changes right. 
Maybe there is a better strategy for managing them?

Could you reframe the question for me please?

mixmix:
Basically wanted you to check that that diff looked ok.

Merge it, it's probably fine and i was just being cautious

robguthrie:
thank you mr mix. I super big time appreciate you doing this. xxxxxxx

PR #2452, opened 2015-08-31
Group images

gdpelican:
This feature rocks!

Would love to hear some feedback and get this bit ready to be merged so we can move on to default cover images

### DIFF
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -1,5 +1,6 @@
 class API::GroupsController < API::RestfulController
   load_and_authorize_resource only: :show, find_by: :key
+  load_resource only: :upload_photo, find_by: :key
 
   def archive
     load_resource
@@ -11,6 +12,19 @@ class API::GroupsController < API::RestfulController
     load_and_authorize :group
     @groups = @group.subgroups.select{|g| can? :show, g }
     respond_with_collection
+  end
+
+  def upload_photo
+    ensure_photo_params
+    service.update group: resource, actor: current_user, params: { params[:kind] => params[:file] }
+    respond_with_resource
+  end
+
+  private
+
+  def ensure_photo_params
+    params.require(:file)
+    raise ActionController::UnpermittedParameters.new([:kind]) unless ['logo', 'cover_photo'].include? params.require(:kind)
   end
 
 end
--- a/lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
+++ b/lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
@@ -1,0 +1,21 @@
+.group-cover-photo-modal
+  .lmo-disabled-form{ng-show: 'isDisabled'}
+  .modal-header
+    %button{type: 'button', class: 'close', ng-click: '$close()'}
+      %span{aria-hidden: 'true'} &times;
+      %span.sr-only{translate: 'common.action.cancel'}
+    %h1.lmo-h1.modal-title{translate: 'group_cover_modal.heading'}
+
+  .modal-body
+    %button.lmo-btn-link{href: '', ng-click: 'selectFile()'}
+      .lmo-box--small.lmo-float--left
+        .user-avatar__initials--small
+          %i.fa.fa-lg.fa-camera
+      %span.lmo-option-text{translate: 'group_cover_modal.upload_link'}
+    .lmo-clearfix
+    %input.hidden.group-cover-photo-modal__file-input{type: 'file', ng-model: 'files', ng-file-select: 'upload(files)'}
+    %p{translate: 'group_cover_modal.helptext'}
+    %p{translate: 'group_cover_modal.image_size_helptext'}
+
+  .modal-footer.lmo-clearfix
+    %button.btn.btn-warning{ng-click: '$close()', type: 'button', translate: 'common.action.cancel'}
--- a/lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
+++ b/lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
@@ -1,4 +1,4 @@
-.group-cover-photo-modal
+.cover-photo-form
   .lmo-disabled-form{ng-show: 'isDisabled'}
   .modal-header
     %button{type: 'button', class: 'close', ng-click: '$close()'}
@@ -13,7 +13,7 @@
           %i.fa.fa-lg.fa-camera
       %span.lmo-option-text{translate: 'group_cover_modal.upload_link'}
     .lmo-clearfix
-    %input.hidden.group-cover-photo-modal__file-input{type: 'file', ng-model: 'files', ng-file-select: 'upload(files)'}
+    %input.hidden.cover-photo-form__file-input{type: 'file', ng-model: 'files', ng-file-select: 'upload(files)'}
     %p{translate: 'group_cover_modal.helptext'}
     %p{translate: 'group_cover_modal.image_size_helptext'}
 
--- a/lineman/app/components/group_page/group_page.haml
+++ b/lineman/app/components/group_page/group_page.haml
@@ -1,32 +1,37 @@
 .loading-wrapper.container.main-container
   %loading{ng-if: '!groupPage.group'}
   %main.group-page{ng-if: 'groupPage.group'}
-    .group-page__cover{style:'background-image:url({{groupPage.group.coverUrl()}})'}
-    %section.group-page__header{aria-label: "{{ 'group_page.header.aria_label' | translate }}"}
-      .group-page__logo{style:'background-image:url({{groupPage.group.logoUrl()}})', alt: "{{ 'group_page.group_logo' | translate }}"}
-      .group-page__name-and-actions
-        .group-page__name
-          %h1.lmo-h1
-            %a{ng_show: 'groupPage.group.parentId', href: '/g/{{groupPage.group.parent().key}}'}
-              {{groupPage.group.parentName()}}
-            %span{ng_show: 'groupPage.group.parentId'}> \-
-            {{groupPage.group.name}}
-        %group_privacy_dropdown{group: 'groupPage.group'}
-        .group-page__member-actions.lmo-btn-group.lmo-btn-group-right{ng-if: 'groupPage.isMember()'}
-          .lmo-btn-wrapper
-            %group_actions_dropdown{group: 'groupPage.group'}
-          .lmo-btn-wrapper
-            %notification_volume_dropdown{translate-root: 'group_page', group: 'groupPage.group'}
-        %join_group_button{group: 'groupPage.group'}
-        .clearfix
-    .row
-      .col-sm-8
-        %section.group-page__description{aria-labelledby: 'description-card-title'}
-          %h2#description-card-title.lmo-card-heading{ translate: 'group_page.description'}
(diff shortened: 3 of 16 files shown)


lineman/app/components/group_page/logo_photo_form/logo_photo_form.coffee
```diff
@@ -0,0 +1,14 @@
+angular.module('loomioApp').factory 'LogoPhotoForm', ->
+  scope: {group: '='}
+  templateUrl: 'generated/components/group_page/logo_photo_form/logo_photo_form.html'
+  controller: ($scope, $rootScope, $timeout, group, Records, FlashService) ->
+
+    $scope.selectFile = ->
+      $timeout -> document.querySelector('.group-logo-modal__file-input').click()
+
+    $scope.upload = (files) ->
```
gdpelican: Lots of duplication here with the cover_photo_form; a single controller with two templates? They're really similar forms.

app/controllers/api/groups_controller.rb
```diff
@@ -13,4 +14,17 @@ def subgroups
     respond_with_collection
   end
 
+  def upload_photo
```
gdpelican: The code for updating cover / logo is identical, so I really don't want to have two actions for it. Since it's a multipart POST request, we can't include params and a file at the same time without hoop jumping, thus this `group/:key/upload_photo/:kind` route

lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
```diff
@@ -0,0 +1,19 @@
+.group-cover-photo-modal
+  .modal-header
+    %button{type: 'button', class: 'close', ng-click: '$close()'}
+      %span{aria-hidden: true} &times;
```
rdbartlett: Gotcha: need quotes around `aria-hidden: 'true'`

lineman/app/components/group_page/group_page_controller.coffee
```diff
@@ -8,9 +8,15 @@ angular.module('loomioApp').controller 'GroupPageController', ($rootScope, $rout
     $rootScope.$broadcast 'setTitle', @group.fullName()
     $rootScope.$broadcast 'analyticsSetGroup', @group
     MessageChannelService.subscribeTo("/group-#{@group.key}")
+    @setPhotos()
   , (error) ->
     $rootScope.$broadcast('pageError', error)
 
+  @setPhotos = =>
```
gdpelican: This is the only thing not currently working; the `coverStyle` is being set, but not updated in the template. Apparantly I don't quite get how ng-style works or something.

gdpelican: Intended workflow is: 
- On pageload, set the photo styles
- As a callback from a photo being uploaded, reset the styles
(I couldn't quite get a simple two-way binding to work for me here)

lineman/app/services/form_service.coffee
```diff
@@ -13,23 +13,49 @@ angular.module('loomioApp').factory 'FormService', ($rootScope, FlashService, $f
       422: 'unprocessableEntity'
       500: 'internalServerError'
 
+    prepare = (scope, model, options) ->
```
gdpelican: I decided to split this out, because I want to keep all of the goodness of disabling the form, flash messages, broadcasting errors, etc. of submitting a form, but the submit function is a little different for upload.

This is slightly more complicated, but it still reads really well.

lineman/app/models/group_records_interface.coffee
```diff
@@ -6,3 +6,6 @@ angular.module('loomioApp').factory 'GroupRecordsInterface', (BaseRecordsInterfa
       @fetch
         path: "#{parentGroup.id}/subgroups"
         cacheKey: "subgroupsFor#{parentGroup.key}"
+
+    uploadPhoto: (group, file, kind) =>
+      @remote.upload("#{group.key}/upload_photo/#{kind}", file).then @remote.onSuccess, @remote.onFailure
```
gdpelican: I'm pretty sure this fix wants to go into the record store itself; in this case we want to import the response into the record store, so it can update the cover and logo url automatically.

lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
```diff
@@ -0,0 +1,21 @@
+.group-cover-photo-modal
+  .lmo-disabled-form{ng-show: 'isDisabled'}
+  .modal-header
+    %button{type: 'button', class: 'close', ng-click: '$close()'}
+      %span{aria-hidden: true} &times;
```
robguthrie: aria-hidden: 'true' (note the single quotes) or it won't work.
Yeah.. we're overdue for a directive for this.

lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
```diff
@@ -0,0 +1,21 @@
...
+      %span.sr-only{translate: 'common.action.cancel'}
+    %h1.lmo-h1.modal-title{translate: 'group_cover_modal.heading'}
+
+  .modal-body
+    %a.lmo-link{href: '', ng-click: 'selectFile()'}
```
robguthrie: another a11y think. Could you make this a button element with the .lmo-btn-link class

lineman/app/components/group_page/group_page.haml
```diff
@@ -1,32 +1,37 @@
...
-          .group-page__description-text.lmo-last.lmo-markdown-wrapper{ marked: 'groupPage.group.description' }
-        %discussions_card{group: 'groupPage.group'}
-      .col-sm-4
-        %membership_requests_card{group: 'groupPage.group'}
-        %members_card{group: 'groupPage.group'}
-        %subgroups_card{group: 'groupPage.group'}
+    .group-page__cover{ng-style:'groupPage.coverStyle()'}
+      %a.group-page__upload-photo{href: '', ng-click: 'groupPage.openUploadCoverForm()', ng-show: 'groupPage.canUploadPhotos()'}
+        %i.fa.fa-lg.fa-camera
+    %main.container.main-container
```
robguthrie: FYI - We had scratch our heads a little bit to remove the double %main element already.. this is probably what is currently conflicting. Can you please ensure we don't have two main elements when the merge happens

lineman/app/components/group_page/group_page.haml
```diff
@@ -1,32 +1,37 @@
...
-          %h2#description-card-title.lmo-card-heading{ translate: 'group_page.description'}
-          .group-page__description-placeholder.lmo-placeholder.lmo-hint-text{translate: 'group_page.description_placeholder', ng-if: 'groupPage.showDescriptionPlaceholder()' }
-          .group-page__description-text.lmo-last.lmo-markdown-wrapper{ marked: 'groupPage.group.description' }
-        %discussions_card{group: 'groupPage.group'}
-      .col-sm-4
-        %membership_requests_card{group: 'groupPage.group'}
-        %members_card{group: 'groupPage.group'}
-        %subgroups_card{group: 'groupPage.group'}
+    .group-page__cover{ng-style:'groupPage.coverStyle()'}
+      %a.group-page__upload-photo{href: '', ng-click: 'groupPage.openUploadCoverForm()', ng-show: 'groupPage.canUploadPhotos()'}
```
robguthrie: again, probably best that this is a button element

lineman/app/components/group_page/logo_photo_form/logo_photo_form.haml
```diff
@@ -0,0 +1,20 @@
+.group-logo-modal
+  .lmo-disabled-form{ng-show: 'isDisabled'}
+  .modal-header
+    %button{type: 'button', class: 'close', ng-click: '$close()'}
+      %span{aria-hidden: true} &times;
```
robguthrie: 'true'

lineman/app/components/group_page/logo_photo_form/logo_photo_form.haml
```diff
@@ -0,0 +1,20 @@
...
+      %span.sr-only{translate: 'common.action.cancel'}
+    %h1.lmo-h1.modal-title{translate: 'group_logo_modal.heading'}
+
+  .modal-body
+    %a.lmo-link{href: '', ng-click: 'selectFile()'}
```
robguthrie: button (lemme know if this is too fine detail)

lineman/app/models/group_records_interface.coffee
robguthrie: Cool. That's now part of 2.3.1

robguthrie: Also I think this should be on GroupModel.. since it is a method that works on a single group.. then there is one less argument to it also.

I'm hoping we'll be able to remove all the recordsInterfaces soon.. I'll come up with a nice way to put all the various fetch methods in a block on the model

lineman/app/services/form_service.coffee
```diff
@@ -13,23 +13,49 @@ angular.module('loomioApp').factory 'FormService', ($rootScope, FlashService, $f
...
+      FlashService.loading()
```
robguthrie: Could we please say "Uploading" rather than loading?

lineman/app/services/form_service.coffee
```diff
@@ -13,23 +13,49 @@ angular.module('loomioApp').factory 'FormService', ($rootScope, FlashService, $f
...
+      scope.isDisabled = true
+      model.setErrors
```
robguthrie: do you mean to call this method? (missing parens)

robguthrie: also I added a clearErrors() method because it makes more sense when reading stuff like this.

lineman/app/services/form_service.coffee
```diff
@@ -13,23 +13,49 @@ angular.module('loomioApp').factory 'FormService', ($rootScope, FlashService, $f
...
+
+    failure = (scope, model, options) ->
+      (response) ->
+        FlashService.dismiss()
+        model.setErrors response.data.errors                            if response.status == 422
+        $rootScope.$broadcast errorTypes[response.status] or 'unknownError',
+          model: model
+          response: response
+
+    cleanup = (scope) ->
```
robguthrie: clearErrors here?

robguthrie:
Thanks for this @gdpelican. How are you going with the e2e?

lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
gdpelican: Hm, yeah, and I copied this from some other place in the code, which means there's other places it's not right. There's that BaseFormModal class cropping up again...

lineman/app/services/form_service.coffee
gdpelican: Should have happened already when you submit the form, I think; that's what the `setErrors()` you mentioned above is trying to do. (I think it's not so good to potentially sit there and look at a validation error while the form is processing)

gdpelican:
Haven't started on it yet; hoping to charge through it today.

lineman/app/components/group_page/cover_photo_form/cover_photo_form.haml
robguthrie: oh I did a search and replace on all the source earlier in the week. Yes indeed on _something_ that allows reuse of the same template.

PR #2454, opened 2015-08-31
Add intercom to angular

robguthrie:
ready for review please. Tested and works.

### DIFF
--- a/app/serializers/group_serializer.rb
+++ b/app/serializers/group_serializer.rb
@@ -6,6 +6,7 @@ class GroupSerializer < ActiveModel::Serializer
              :key,
              :name,
              :created_at,
+             :creator_id,
              :description,
              :members_can_add_members,
              :members_can_create_subgroups,
--- a/app/views/layouts/angular.html.haml
+++ b/app/views/layouts/angular.html.haml
@@ -8,15 +8,8 @@
     %base{href: '/'}
     = stylesheet_link_tag lineman_css_path
 
-    -# google analytics code for production
-    - if Rails.env.production? and ENV['GOOGLE_ANALYTICS_ID']
-      :javascript
-        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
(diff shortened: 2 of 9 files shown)


app/serializers/group_serializer.rb
```diff
@@ -6,6 +6,7 @@ class GroupSerializer < ActiveModel::Serializer
              :key,
              :name,
              :created_at,
+             :creator_id,
```
rdbartlett: Note for another time: I think we actually want this to be something like :account_owner or :org_owner or something. Group creator can leave the group, but there should always be an org owner.

rdbartlett:
Looks good to me but I don't really know what I'm looking at :)

Should be relatively safe to test in production eh?

app/views/layouts/angular.html.haml
```diff
@@ -8,8 +8,7 @@
     %base{href: '/'}
     = stylesheet_link_tag lineman_css_path
 
-    -# google analytics code for production
-    - if Rails.env.production? and ENV['GOOGLE_ANALYTICS_ID']
+    - if ENV['GOOGLE_ANALYTICS_ID']
```
gdpelican: Is this something we want on in dev / test?

gdpelican:
Looks alright to me, although I'm starting to wish didn't need to be using ENV directly in the view for angular.html.haml, and am thus wondering if a `LoomioSettings` class with `intercom_enabled` / `analytics_enabled?` / etc methods should exist.

robguthrie:
I've thought about that a bit and honestly I think it would make things more complex rather than less.

I removed the production lines because of the same thing. If you have those ENV's set then you must want that stuff to run. Why worry about the Rails env too?

gdpelican:
It becomes useful when you want to start having services which have
slightly more complex checks to see if they can function or not (the inline
translation thing would be a good example, which needs a couple different
ENV variables to function correctly), or services which need to appear in
multiple places (less of a problem now that we're running a single page
app, although we'll still have other pages on vanilla rails for a long time
going forward - what if we want to change the criteria for displaying the
Loomio.org footer, but have to do it in a bunch of different templates and
the angular html?

It also makes an easier path towards making an admin panel which would
allow you to toggle features on and off, without having to know anything
about ENV variables.

```
def self.inline_translation_enabled?
  Rails.env.production? &&
  ENV['BING_TRANSLATION_API_KEY'] &&
  ENV['BING_TRANSLATION_SECRET'] &&
  !ENV['TRANSLATION_DISABLED'] # provided trigger from admin panel
end
```

robguthrie:
@gdpelican I've tidied this up a bit.. have a look now.

PR #2456, opened 2015-09-01
fix linking to comments

robguthrie:
there is still a weird html encoding issue: When these links are sent to the browser they look like this: http://localhost:8000/d/Fxdes1CL/what-star-sign-are-you-%3Fcomment=722017

### DIFF
--- a/lineman/app/services/lmo_url_service.coffee
+++ b/lineman/app/services/lmo_url_service.coffee
@@ -22,7 +22,7 @@ angular.module('loomioApp').factory 'LmoUrlService', ->
       @buildModelRoute('m', p.key, p.name, params, options)
 
     comment: (c, params = {}, options = {}) ->
-      @discussion c.discussion(), _.merge(params, {comment: c.key})
+      @discussion c.discussion(), _.merge(params, {comment: c.id})
 
     buildModelRoute: (path, key, name, params, options) ->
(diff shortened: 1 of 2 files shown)


robguthrie:
@gdpelican do you know how to prevent html encoding of these urls?

gdpelican:
I seem to remember doing something similar to this when doing the slack links.. i can take a look in a little bit.

gdpelican:
Bleh, I poked at this for a little bit; it's that the angular router is encoding the path string we pass it (It's trying to avoid situations like `/d/kjf48f/do-we-go-to-the-mall-?%20Or-the-zoo-?`)

Not sure what the perfect answer for this is just yet; will likely require a rethink of how we are trying to return comment urls

gdpelican:
All roads lead to the angular router:

https://github.com/angular/router/issues/176

gdpelican:
Well, a preliminary smoke test of changing `$location.path` to `$location.url` as suggested in that thread seems to be working alright, but the guy who tried it couldn't seem to get the tests passing with it, so seems like not the perfect fix.. not sure if I want to keep digging down this hole or not..

gdpelican:
Going to merge this partial fix in; will need some more thinking about how to tiptoe around the router again.

PR #2457, opened 2015-09-01
Add tabindex to h1s so we can focus them on page load

robguthrie:


robguthrie:
@gdpelican do you think I should be doing this? I'm starting to think we should make an lmo-h1 tag rather than h1

gdpelican:
It's the equivalent of Monkey patching in rails, but for some rreason this seems fine to me... no more than 1 h1 per page...

robguthrie:
I've come 180 - I really don't like invisible things. lmo-h1 could be
loaded up with more stuff and I think we should be explicit about when
there is a behaviour going on. It's worked really well in the css with
no/very little inheritance. I'll change it soon.

robguthrie:
Ok I've tried this as lmo_h1 tag with transclusion.. it means you cannot set the class of the h1 itself.
(you end up with <lmo-h1 class="classes you want on h1" ng-transclude><h1>title</h1></lmo-h1>

I've also tried it as a class directive on the .lmo-h1 class.. but then it does not work on the dashboard page as we use a different class there.

I'm coming back around to this way. Let's give this a go and evaluate as we learn.

PR #2459, opened 2015-09-01
Add trial card, pricing modal

HSalmon:
image


AaronThornton00:
These comments are more design based, but thought I would mention them here
anyway.
It feels to me like the call to action here is 'cancel' as that is the
color that stands out and it is located bottom right.
The select buttons are greyed out and positioned to the left, which is
generally a cancel or back position.
One solution might be to colour the select buttons green and grey the
cancel then swap their relative positions.  Just a comment thought take it
of leave it.

AaronThornton00:
Another option might be to make the whole tier div the button?

rdbartlett:
Good idea to make the tier the button. This is just a WIP commit, not
finished :)

HSalmon:
screen shot 2015-09-02 at 11 34 30 am

screen shot 2015-09-03 at 11 33 17 am

gdpelican:
This looks great @HSalmon

I'm seeing some missing translations 'Ok, Got it!', 'We're happy to talk', etc - any reason to not translate those?

I think I'm feeling some of @AaronThornton00 's design suggestions above, but not sure what the right channel for giving them is. I want 'cancel' to be 'btn-primary' (blue) and say something like 'Remind me later', and the background of the highlighted 'selected' plan to be bluey or greeny (let's go!) instead of yellowy (caution!)

HSalmon:
Todo:

- put subscription urls in environment
- send subscription urls in app/config

- on signup success user is redirected to subscriptions/signup_success
- we want to update subscription to be chargify (or paid) and redirect them to thank you page
- split return on comma to get user id and group id

HSalmon:
Hey @gdpelican, starting a subscription doesn't seem to work anymore:

screen shot 2015-09-07 at 5 26 04 pm

robguthrie:
try removing the oj gem it might be causing problems. (i'm not really here,
feel free to ignore)

config/routes.rb
```diff
@@ -1,27 +1,7 @@
...
-    get 'setup_proposal_with_votes'
-    get 'setup_proposal_closing_soon'
-    get 'setup_closed_proposal'
-    get 'setup_closed_proposal_with_outcome'
-    get 'setup_membership_requests'
-    get 'setup_user_email_settings'
-    get 'setup_all_notifications'
-    get 'setup_group_with_pending_invitation'
-    get 'setup_group_with_subgroups'
+    get ':action'
```
gdpelican: Nice!

lineman/app/components/group_page/gift_card/gift_card.coffee
```diff
@@ -0,0 +1,10 @@
+angular.module('loomioApp').directive 'giftCard', ->
+  scope: {group: '='}
+  restrict: 'E'
+  templateUrl: 'generated/components/group_page/gift_card/gift_card.html'
+  replace: true
+  controller: ($scope, $window, AppConfig) ->
+
+    $scope.makeDonation = ->
+      $window.open("#{AppConfig.chargify.donation_url}", '_blank')
+      return true # throws an exception without this line
```
gdpelican: No need for 'return', just 'true' should do.

lineman/app/components/group_page/group_actions_dropdown/group_actions_dropdown.coffee
```diff
@@ -24,4 +27,11 @@ angular.module('loomioApp').directive 'groupActionsDropdown', ->
     @archiveGroup = ->
       ModalService.open ArchiveGroupForm, group: -> $scope.group
 
+    @manageSubscriptionsLink = ->
+      "https://www.billingportal.com/s/#{AppConfig.chargify.appName}/login/magic"
```
gdpelican: We should probably have this open in a new tab as well.

lineman/app/services/modal_service.coffee
```diff
@@ -1,8 +1,12 @@
 angular.module('loomioApp').factory 'ModalService', ($modal, $rootScope) ->
+  currentModal = null
   new class ModalService
     open: (modal, resolve = {}) ->
       $rootScope.$broadcast 'modalOpened'
-      $modal.open
+      if currentModal?
```
gdpelican: This is cool.

HSalmon:
Hi @gdpelican, when I pulled this down to demo it I found that when you subscribe it doesn't update the client side to reflect it (i.e. Trial card is still there and there is no option to manage group subscription in the dropdown). Didn't have time to delve into it - I'm sorry, if I'd known (or remembered that there was still a little more to do?) I woulda taken a look earlier.

gdpelican:
Have to set up the webhook.

robguthrie:
merged into new groups start on angular

PR #2462, opened 2015-09-01
add greyscale to aaarrr table

rdbartlett:
Whoops, trying again. This makes the aaarrr table like:

image


rdbartlett:
@robguthrie do you think this is cool?

robguthrie:
yep!

PR #2465, opened 2015-09-01
Temp fix for groups not appearing after deactivate

gdpelican:
I just wanted to put this up for a little discussion before I worked on the record store itself @robguthrie 

I think the minimum viable fix here is to add a `vanish` (naming? Something which says 'remove from client database') method to `base_model`, and manually call it for dependant: destroy relationships.

Later on, when we have more use cases, we can put in a more automatic way of dealing with orphans.

```
@remote.patchMember(@key, 'archive').then =>
  @vanish()
  _.each @memberships(), (m) -> m.vanish()
```

### DIFF
--- a/lineman/app/components/groups_page/groups_page_controller.coffee
+++ b/lineman/app/components/groups_page/groups_page_controller.coffee
@@ -3,7 +3,7 @@ angular.module('loomioApp').controller 'GroupsPageController', ($rootScope, Curr
   $rootScope.$broadcast('setTitle', 'Groups')
 
   @parentGroups = =>
-    _.unique _.map CurrentUser.memberships(), (membership) =>
+    _.unique _.compact _.map CurrentUser.memberships(), (membership) =>
       if membership.group().isParent()
         membership.group()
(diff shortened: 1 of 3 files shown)


robguthrie:
James I don't understand why remove isn't what you want. If you remove a record from the collection it will be vanished.

gdpelican:
We could call it remove, 
```
remove: ->
  @recordsInterface.remove(@)
```

but there's no remove method on the base model itself, and I don't want to have to reach through the recordsInterface like
```
_.each @memberships(), (m) -> m.recordsInterface.remove(m)
```

I'm also not certain that 'remove' says to me that we're only removing a record from the client database, rather than from the server side.

robguthrie:
I don't think vanish helps though. I'm happy to add BaseModel.remove

gdpelican:
Cool, I think that's a good way to go.

robguthrie:
@gdpelican angular_record_store 2.3.0 has a mode.remove() method :+1:

gdpelican:
Hooray!

gdpelican:
Sweet, this looks great. I'll squash it up.

PR #2466, opened 2015-09-01
Random group image

gdpelican:
This one's ready for a look, @robguthrie @rdbartlett, @HSalmon but I think it needs to go in after the upload group cover photo stuff goes.

There's also a little bit of grunt work to do in prod to include the photos we want.. I waffled around making an admin panel for these uploads... I might still do that.


app/models/group.rb
```diff
@@ -604,4 +604,12 @@ def limit_inheritance
   def self.with_one_coordinator
     published.select{ |g| g.admins.count == 1 }
   end
+
+  def default_cover_photo
+    if self.default_group_cover
+      /^.*(?=\?)/.match(self.default_group_cover.cover_photo.url).to_s
```
gdpelican: Hi, this regex is confusing. Here's why it's there:

- By default, the paperclip url has an integer hash at the end of it, like `aws.com/test.jpg?10483202` (I assume this is a versioning / caching mechanism)
- When this gets transferred over to this url, it includes this hash, which then gets url encoded: `aws.com/test.jpg%3F10483202`
- This breaks everything

Loading the image seems to work okay without the version, and I'm honestly not that worried about whether we get the latest and greatest version of a default group logo; the point is that it's there and it's fine whatever it is.

Oh right, so this regex basically strips off everything after the first `?` in the url, leaving us with just `aws.com/test.jpg`, which loads up nicely.

app/serializers/group_serializer.rb
```diff
@@ -36,11 +36,11 @@ def include_logo_url_medium?
...
   def cover_url_desktop
-    object.cover_photo.url(:desktop)
-  end
-
-  def include_cover_url_desktop?
-    object.cover_photo.present?
+    if object.cover_photo.present? || object.default_group_cover.present?
+      object.cover_photo.url(:desktop)
+    else
+      'img/default-cover-photo.png'
```
gdpelican: This is because the default for angular and the default for beta are different at the moment (`img/default-cover-photo.png` vs `default-cover-photo.png`)

Is this a great place to put this? No. Are we going to make the move at some point and apply default cover photos to all existing groups which don't have cover photos? I think so, so can go away when that happens.

gdpelican:
Extra bonus points for not needing to upload a new cover photo every time a new group is created! :pill:

app/models/group.rb
robguthrie: Can we just migrate all groups without custom covers to have a DefaultGroupCover? I think that would be fine

rdbartlett: Agreed

robguthrie:
Sorry I took a while to review this. Looks good. rather "iffy" but understandable. Hopefully we can remove the old default cover and remove one set of those if's.

Also thinking about new installs.. is there an easy way we can provide seed data for some default covers? db/seeds.rb maybe?

app/models/group.rb
gdpelican: After a little experimentation, this is too slow to do in a migration, I think. The paperclip transfer is slow (~2/3 sec per record), and we've got over 20,000 groups in prod with no cover photo.

```
Group.joins('LEFT JOIN groups p ON p.id = groups.parent_id')
     .where('groups.cover_photo_file_name IS NULL AND p.cover_photo_file_name IS NULL')
     .count
=> 22220
```

That spins out to like a 7 hour migration.

We could do something like 'only groups with no cover photo created in the last 90 days' or something, which cuts down the numbers significantly, but then we've gotta keep the spare code in there anyway and I'm tempted not to bother.

robguthrie: What if they all have the same cover photo?.. ie: is there a database only migration I can help you write? `UPDATE groups set cover_photo_file_name = ? where cover_photo_file_name is null`
? Is it it more complex than that?

gdpelican: It is, because paperclip takes the id of the object into account to look for the object. So, given 
```
Group.new cover_photo_file_name: 'test.jpg', id: 15
```

it'll go and look in the Fog store for
```
groups/cover_photos/000/000/15/test.jpg
```
(or something like that), which it won't find because we haven't uploaded it.

Next step to fix that would be to override the paperclip default path to something which doesn't involve the id, which is really scary to me and I think would require a data migration of all of our existing groups. :/

gdpelican: Rob and I came up with a super easy and great way to do this, hooray!

gdpelican:
Closing to in favor of a master branch merge

PR #2468, opened 2015-09-01
add includes to improve our api response times

robguthrie:


robguthrie:
This will help a little bit when we see rails making lots of requests for assocated records when serializing. Looks like this:
image

PR #2474, opened 2015-09-03
small changes from a11y review 3 sept 2015

robguthrie:
dashboard - replace "starred" with "starred threads"?
Thread page - remove %article from Thread context region
Label (aria-labelledby) the comment textarea by the h2 above it
Question mark dangling in the Start proposal button
Current votes table try role=presentation, use caption element to label
it
add sr-only heading for undecided memebers and focus it when showing
undecided members


rdbartlett:
Nice!

PR #2477, opened 2015-09-03
Random default group images

gdpelican:


app/models/default_group_cover.rb
```diff
@@ -0,0 +1,14 @@
+class DefaultGroupCover < ActiveRecord::Base
+  has_attached_file :cover_photo
+  validates_attachment :cover_photo, content_type: { content_type: ["image/jpg", "image/jpeg", "image/png", "image/gif"] }
+
+  def self.sample
+    all.order('random()').first
+  end
+
+  def self.store(file)
+    file = open(file) if file.is_a? String
```
robguthrie: is this a IO open? If so you might need to close it. Also if so, use a do |file| end so you don't need to close it manually.

lineman/app/models/group_records_interface.coffee
```diff
@@ -6,3 +6,6 @@ angular.module('loomioApp').factory 'GroupRecordsInterface', (BaseRecordsInterfa
       @fetch
         path: "#{parentGroup.id}/subgroups"
         cacheKey: "subgroupsFor#{parentGroup.key}"
+
+    uploadPhoto: (group, file, kind) =>
```
robguthrie: I think you can drop onSuccess, onFailure

gdpelican: Cool, this is a merge artifact I think; the uploadPhoto method should be on group_model now.

app/models/default_group_cover.rb
gdpelican: Yep, it's open-uri; made that change to use a block.

gdpelican:
The merge :monkey: strikes!

PR #2482, opened 2015-09-10
remove start subgroup option from FAB

rdbartlett:
We don't want to give subgroups this level of prominence


gdpelican:
Cool, but I want to wait to deploy this until we fix the 'Add subgroup' button, otherwise there won't be a way to create a subgroup in the app that works.

PR #2483, opened 2015-09-10
link to markdown help page from comment form

rdbartlett:
Looks like this:

image

### DIFF
--- a/lineman/app/components/thread_page/comment_form/comment_form.haml
+++ b/lineman/app/components/thread_page/comment_form/comment_form.haml
@@ -17,4 +17,5 @@
       %button.btn.btn-default.comment-form-button{type: 'button', translate: 'comment_form.cancel_reply', ng-show: 'comment.parentId', ng-click: 'comment.parentId = null'}
       %attachment_form{comment: 'comment'}
       %button#post-comment-cancel.btn.btn-default{type: 'button', translate: 'common.action.cancel', ng-click: 'cancel($event)', ng-hide: 'comment.isNew()'}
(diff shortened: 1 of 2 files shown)


lineman/app/components/thread_page/comment_form/comment_form.haml
```diff
@@ -17,3 +17,4 @@
       %button.btn.btn-default.comment-form-button{type: 'button', translate: 'comment_form.cancel_reply', ng-show: 'comment.parentId', ng-click: 'comment.parentId = null'}
       %attachment_form{comment: 'comment'}
       %button#post-comment-cancel.btn.btn-default{type: 'button', translate: 'common.action.cancel', ng-click: 'cancel($event)', ng-hide: 'comment.isNew()'}
+      %a.comment-form-button{href: 'https://loomio.org/markdown', target: '_blank', title: "{{ 'common.formatting_help.title' | translate }}", translate: 'common.formatting_help.label' }
```
gdpelican: Quick primer on our lmo- directives using href (because Angular doesn't do nice things with it by default)

**lmo-href**: Simply takes a path which works like a normal href (but will allow ctrl-click, open in new tab, etc.)
**lmo-href-for**: Takes a model (a discussion, group, etc.) and links to the show page for that model. Optionally can also take an lmo-href-action directive, which will customize the path the link takes you to

eg:
`a{lmo-href: '/markdown'}` => `/markdown`
`a{lmo-href-for: 'group'}` => `/g/<group-key>`
`a{lmo-href-for: 'group', lmo-href-action: 'memberships'}` => `/g/<group-key>/memberships`

rdbartlett:
Swell, thanks for the primer. :santa:

PR #2490, opened 2015-09-15
Subscribe to all current user groups in a single request

gdpelican:


robguthrie:
brief look and I approve!

gdpelican:
I'm pretty confident with this, but would love to smoke test it with someone after deploying; maybe we could do that early next week @HSalmon?

PR #2492, opened 2015-09-16
fix vote button colours + make them pop on hover

rdbartlett:
Updates vote icons so they use the same colour as the pie chart.

Gives vote buttons a shadow + 1px lift so they grab your attention on hover:

image


AaronThornton00:
@rdbartlett  I haven't seen how your pop looks, but I think the usual way to do this is to raise the button (increase the box shadow). You can use the box shadow mixin I put in. The real ninja way is to create a ripple effect, but Inbox simply raises the button (check there FAB on press) :)

AaronThornton00:
Nice work on the buttons though man. That has been screening out for ever

gdpelican:
I reckon it's starting to feel like a really good time to get an icon designer on this, no?

rdbartlett:
The .btn class provides shadow on hover. I just added the 1px vertical
shift to make it really bombastic.

AaronThornton00:
Nice

PR #2503, opened 2015-09-18
Hide member actions from visitors

HSalmon:
Still writing tests but could maybe do with preliminary review?

[Trello card](https://trello.com/c/VdXrTc9p/677-hide-members-actions-from-non-members-e-g-comment-like-reply-on-public-threads)

### DIFF
--- a/lineman/app/components/notification_volume/notification_volume_dropdown.haml
+++ b/lineman/app/components/notification_volume/notification_volume_dropdown.haml
@@ -1,28 +1,29 @@
-.notification-volume{dropdown: ''}
-  %button.btn.lmo-btn-nude.notification-volume__button{dropdown-toggle: ''}
-    .screen-only{aria-hidden: 'true'}
-      %i.fa.fa-lg.fa-fw{ng-class: 'iconClassForLevel(model.volume)'}>
-      %span{translate: 'volume_levels.loud', ng-show: 'model.volume == "loud"'}
-      %span{translate: 'volume_levels.normal', ng-show: 'model.volume == "normal"'}
-      %span{translate: 'volume_levels.quiet', ng-show: 'model.volume == "quiet"'}
-      %span{translate: 'volume_levels.mute', ng-show: 'model.volume == "mute"'}
-    .sr-only
-      %span{translate: "{{translateRoot}}.volume.loud_aria_description", ng-show: 'model.volume == "loud"'}
-      %span{translate: "{{translateRoot}}.volume.normal_aria_description", ng-show: 'model.volume == "normal"'}
-      %span{translate: "{{translateRoot}}.volume.quiet_aria_description", ng-show: 'model.volume == "quiet"'}
-      %span{translate: "{{translateRoot}}.volume.mute_aria_description", ng-show: 'model.volume == "mute"'}
+.blank
+  .notification-volume{ng-if: 'showVolumeDropdown()', dropdown: ''}
+    %button.btn.lmo-btn-nude.notification-volume__button{dropdown-toggle: ''}
+      .screen-only{aria-hidden: 'true'}
+        %i.fa.fa-lg.fa-fw{ng-class: 'iconClassForLevel(model.volume)'}>
+        %span{translate: 'volume_levels.loud', ng-show: 'model.volume == "loud"'}
+        %span{translate: 'volume_levels.normal', ng-show: 'model.volume == "normal"'}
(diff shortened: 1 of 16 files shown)


lineman/app/components/notification_volume/notification_volume_dropdown.haml
```diff
@@ -1,4 +1,4 @@
-.notification-volume{dropdown: ''}
+.notification-volume{ng-show: 'canChangeVolume()', dropdown: ''}
```
gdpelican: Let's make these ng-if, since they're unlikely to change during the course of a page visit.

gdpelican: (You likely know this already but..)

ng-if doesn't render the inner block into the DOM at all until it is true, which means any directives, watchers, and nested scopes aren't part of the page. However it takes more time to render and digest all that stuff every time the value changes, so it's better to avoid ng-if if the value is likely to change often.
ng-show just slaps a 'display: none' style on the element when it's false, but still renders and watches for changes to all of the sub elements. It's super low-overhead to switch on and off, though.

lineman/app/services/ability_service.coffee
```diff
@@ -41,6 +50,7 @@ angular.module('loomioApp').factory 'AbilityService', (CurrentUser) ->
       (CurrentUser.isAuthorOf(proposal) or @canAdministerGroup(proposal.group()))
 
     canAdministerGroup: (group) ->
+      CurrentUser.isMemberOf(group) and
```
gdpelican: Think you can lose this one; being an admin implies being a member

HSalmon:
@gdpelican, @robguthrie, does this need anything further?

lineman/app/components/star_toggle/star_toggle.haml
```diff
@@ -1,4 +1,4 @@
-.star-toggle
+.star-toggle{ng-show: 'canStarThread()'}
```
robguthrie: Why would we prevent someone from starring a thread?

lineman/app/components/thread_page/activity_card/new_comment.haml
```diff
@@ -26,11 +26,11 @@
 
   .thread-item__footer
     .thread-item__actions
-      %button.thread-item__action--like.lmo-btn-link{type: 'button', translate: 'common.action.like', ng-show: '!currentUserLikesIt()', ng_click: 'like()'}>
+      %button.thread-item__action--like.lmo-btn-link{ng-if: 'showCommentActions()', type: 'button', translate: 'common.action.like', ng-show: '!currentUserLikesIt()', ng_click: 'like()'}>
```
robguthrie: Is it possible to wrap all of this in a single ng-if: showCommentActions() ?

lineman/app/components/thread_page/proposal_positions_panel/proposal_positions_panel.haml
```diff
@@ -25,7 +25,7 @@
         %span.proposal-positions-panel__name-and-position{translate: "new_vote_item.{{vote.positionVerb()}}", translate-value-actor: '{{vote.authorName()}}'}
         %span{ng-if: '!vote.hasStatement()'}.
         %span{ng-if: 'vote.hasStatement()'}: {{::vote.statement}}
-        %a.proposal-positions-panel__change-your-vote{href: '', ng-click: 'changeVote()', ng-show: 'proposal.isActive() && voteIsMine(vote)'}
+        %a.proposal-positions-panel__change-your-vote{ng-if: 'showChangeVoteOption(vote)', href: '', ng-click: 'changeVote()'}
```
robguthrie: Another 'a' that should be a button styled to look like an 'a'?

robguthrie:
This is looking good @HSalmon. If you address the comments I've written I think it'll be good to go.

lineman/app/components/star_toggle/star_toggle.haml
gdpelican: If they're logged out looking at a public discussion?

robguthrie: We don't support logged out people in the angular app yet. And I think you should be able to star any discussion so long as you're logged in.

lineman/app/components/thread_page/activity_card/new_comment.haml
HSalmon: The reason why I didn't is because I think we still want to show the timestamp.

HSalmon:
@robguthrie, I think I've addressed them

PR #2508, opened 2015-09-21
Inbox reliableness

gdpelican:
Fix unwieldy unread count for threads

### DIFF
--- a/app/controllers/api/events_controller.rb
+++ b/app/controllers/api/events_controller.rb
@@ -5,9 +5,7 @@ class API::EventsController < API::RestfulController
 
   def visible_records
     load_and_authorize :discussion
-    resource_class.
-      includes(:user).
-      where(discussion: @discussion).sequenced
+    @discussion.items.sequenced
   end
 
   def page_collection(collection)
@@ -21,10 +19,10 @@ class API::EventsController < API::RestfulController
   end
 
   def sequence_id_for(collection)
-    (sequence_id_for_comment(collection) if params[:comment_id].present?) || params[:from] || 0
+    sequence_id_for_comment(collection) || params[:from] || 0
   end
 
   def sequence_id_for_comment(collection)
-    collection.where(eventable_type: "Comment", eventable_id: params[:comment_id]).pluck(:sequence_id).first
+    collection.find_by(eventable_type: "Comment", eventable_id: params[:comment_id]).try(:sequence_id)
   end
 end
--- a/app/models/discussion.rb
+++ b/app/models/discussion.rb
@@ -1,7 +1,15 @@
 class Discussion < ActiveRecord::Base
   PER_PAGE = 50
   SALIENT_ITEM_KINDS = %w[new_comment new_motion new_vote motion_outcome_created]
-  THREAD_ITEM_KINDS = %w[new_comment new_motion new_vote motion_outcome_created motion_outcome_updated]
+  THREAD_ITEM_KINDS = %w[new_comment
+                         new_motion
+                         new_vote
+                         motion_closed
+                         motion_closed_by_user
+                         motion_edited
+                         motion_outcome_created
+                         motion_outcome_updated
+                         discussion_edited]
   paginates_per PER_PAGE
 
   include ReadableUnguessableUrls
@@ -48,8 +56,8 @@ class Discussion < ActiveRecord::Base
 
   has_many :events, -> { includes :user }, as: :eventable, dependent: :destroy
 
-  has_many :items, -> { includes(eventable: :user).where(kind: THREAD_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
-  has_many :salient_items, -> { includes(eventable: :user).where(kind: SALIENT_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
+  has_many :items, -> { includes(:user).where(kind: THREAD_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
+  has_many :salient_items, -> { includes(:user).where(kind: SALIENT_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
 
   has_many :discussion_readers
 
@@ -129,8 +137,10 @@ class Discussion < ActiveRecord::Base
   end
 
   def thread_item_created!(item)
-    self.items_count += 1
-    self.last_item_at = item.created_at
+    if THREAD_ITEM_KINDS.include? item.kind
+      self.items_count += 1
+      self.last_item_at = item.created_at
+    end
 
     if SALIENT_ITEM_KINDS.include? item.kind
       self.salient_items_count += 1
--- a/app/services/discussion_service.rb
+++ b/app/services/discussion_service.rb
@@ -1,11 +1,17 @@
 class DiscussionService
   def self.recount_everything!
     # I'm not sure anyone will need this .. but it's cool sql
+    # WHOA I TOTALLY NEEDED IT!
     ActiveRecord::Base.connection.execute(
       "UPDATE discussion_readers SET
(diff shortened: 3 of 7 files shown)


app/services/discussion_service.rb
```diff
@@ -1,11 +1,17 @@
...
-       read_items_count = (SELECT count(id) FROM events WHERE discussion_id = discussion_readers.discussion_id AND events.created_at <= discussion_readers.last_read_at ),
-       read_salient_items_count = (SELECT count(id) FROM events WHERE discussion_id = discussion_readers.discussion_id AND events.created_at <= discussion_readers.last_read_at AND events.kind IN ('#{Discussion::SALIENT_ITEM_KINDS.join('\', \'')}') )", )
+       read_comments_count = (SELECT count(id) FROM events WHERE discussion_id = discussion_readers.discussion_id AND events.created_at <= discussion_readers.last_read_at AND events.kind = 'new_comment'),
+       read_items_count = (SELECT count(id) FROM events WHERE discussion_id = discussion_readers.discussion_id AND events.created_at <= discussion_readers.last_read_at AND events.kind IN ('#{Discussion::THREAD_ITEM_KINDS.join('\', \'')}') ),
+       read_salient_items_count = (SELECT count(id) FROM events WHERE discussion_id = discussion_readers.discussion_id AND events.created_at <= discussion_readers.last_read_at AND events.kind IN ('#{Discussion::SALIENT_ITEM_KINDS.join('\', \'')}') )")
+    ActiveRecord::Base.connection.execute(
```
gdpelican: These updates run pretty slowly; ~13min for 90k discussions... but I think we need em in order to sync up the existing threads, I'm hesistant to pick out a subset of discussions or readers to update, and I don't know of a faster way to run a global update like this.

app/controllers/api/events_controller.rb
```diff
@@ -5,9 +5,7 @@ class API::EventsController < API::RestfulController
 
   def visible_records
     load_and_authorize :discussion
-    resource_class.
-      includes(:user).
-      where(discussion: @discussion).sequenced
+    @discussion.items.sequenced
```
robguthrie: can you keep includes(:user) ? that prevents a bit of n+1

robguthrie: p.s. love how this just got a lot easier to read

app/models/discussion.rb
```diff
@@ -129,8 +137,10 @@ def delayed_destroy
   end
 
   def thread_item_created!(item)
-    self.items_count += 1
-    self.last_item_at = item.created_at
+    if THREAD_ITEM_KINDS.include? item.kind
```
robguthrie: I've been thinking about ripping this out and replacing it with https://github.com/cedric/custom_counter_cache

lineman/app/components/thread_page/activity_card/activity_card.coffee
```diff
@@ -75,15 +75,7 @@ angular.module('loomioApp').directive 'activityCard', ->
...
-      _.contains ['new_comment',
-                  'new_motion',
-                  'new_vote',
-                  'motion_closed',
-                  'motion_closed_by_user',
-                  'motion_edited',
-                  'motion_outcome_created',
-                  'motion_outcome_updated',
-                  'discussion_edited'], kind
+      _.contains AppConfig.safeThreadItemKinds, kind
```
gdpelican: This way we can be sure that the events that come back in `/api/v1/events?discussion_id=:id` are the same as the ones being displayed in the activity card, because the definition of `discussion.items` is 'an event which has discussion id `<discussion.id>` and kind within `<Discussion::THREAD_ITEM_KINDS>`

(We also might be able to just lose this, but I'd rather keep it there for now)

db/migrate/20150921185430_reset_discussion_items_counts.rb
```diff
@@ -0,0 +1,5 @@
+class ResetDiscussionItemsCounts < ActiveRecord::Migration
+  def up
+    DiscussionService.recount_everything!
```
robguthrie: I'd rather run this manually though console than commit it to the migrations

app/controllers/api/events_controller.rb
gdpelican: Oh, sure! Because we include the author automatically with a discussion?

app/models/discussion.rb
gdpelican: That would be very great.

db/migrate/20150921185430_reset_discussion_items_counts.rb
gdpelican: Sounds good to meee!

app/controllers/api/events_controller.rb
robguthrie: yes

robguthrie: We include the actor of the event with each event we serialize.

Gemfile
```diff
@@ -66,6 +66,7 @@ gem 'rack-attack', '~> 4.2.0'
 gem 'ahoy_matey'
 gem 'oj'
 gem 'oj_mimic_json'
+gem 'custom_counter_cache', github: 'gdpelican/custom_counter_cache'
```
gdpelican: I was running into a bug in custom_counter_cache, see here: https://github.com/cedric/custom_counter_cache/pull/11

app/models/discussion.rb
```diff
@@ -48,8 +47,16 @@ class Discussion < ActiveRecord::Base
 
   has_many :events, -> { includes :user }, as: :eventable, dependent: :destroy
 
-  has_many :items, -> { includes(eventable: :user).where(kind: THREAD_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
-  has_many :salient_items, -> { includes(eventable: :user).where(kind: SALIENT_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
+  has_many :items, -> { includes(eventable: :user).where(kind: Event::THREAD_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
+  has_many :salient_items, -> { includes(eventable: :user).where(kind: Event::SALIENT_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
+
+  define_counter_cache(:items_count)         { |discussion| discussion.items.count }
```
gdpelican: This is the simplest implementation, but I'm not certain it's the most performant, since it means we'll be loading up all the items every time an event is created / destroyed for the discussion... I might like to do something like

```
if item_is_created? then discussion.items_count + 1 else discussion.items_count - 1 end
```
But not quite sure how to know whether the callback's being called on create or destroy from the block yet.

app/models/discussion.rb
```diff
@@ -129,32 +136,12 @@ def delayed_destroy
...
     self.last_sequence_id = item.sequence_id
-
     save!(validate: false)
   end
 
   def thread_item_destroyed!(item)
-    self.items_count -= 1
-    self.salient_items_count -= 1 if SALIENT_ITEM_KINDS.include? item.kind
-
     if item.sequence_id == first_sequence_id
```
gdpelican: The sequence ids could likely be counter cached too, but didn't bother with that yet; I'm still not 100% sold on this gem.

gdpelican:
I think I'm going to ditch custom_counter_cache for now; it doesn't work perfectly out of the box, doesn't seem to be supported or all that widely used, and seems to breaking tests that I don't know why it would break.

Might be a nice project for a rainy day, either forking that solution, or rolling our own, but going to leave the hooks in there for now.

app/controllers/api/events_controller.rb
gdpelican: Oh, right. `discussion.items` automatically includes the user:

```
  has_many :items, -> { includes(eventable: :user).where(kind: THREAD_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
```

gdpelican:
I feel itchy to squash and merge this this morning (for some reason). Any objections @robguthrie?

PR #2510, opened 2015-09-22
Update help card content

HSalmon:
New Help card content


HSalmon:
screen shot 2015-09-22 at 3 55 56 pm

robguthrie:
Looking at this.. it's not obvious what the video is. I wonder if we should
have a fake video place holder with a clear title in it, that opens a modal
with a bigger video

HSalmon:
What does this need to get merged @rdbartlett, @robguthrie?

robguthrie:
I'm happy for you to make the call on how much you want to do before deploying. I see it as a cool experiment and am not too fussed on details that delay it's deploy.

HSalmon:
screen shot 2015-09-25 at 1 48 44 pm

lineman/app/components/group_page/group_help_card/group_help_card.haml
```diff
@@ -0,0 +1,17 @@
+%section.group-help-card
```
gdpelican: Hmm... I wonder if we should display this only on loomio.org, and leave it off of future whitelabel versions?

HSalmon: I'm not sure...@rdbartlett, @robguthrie, thoughts?

robguthrie: No I think we just leave it for everyone. We want to see traffic from other installs hitting our docs.

PR #2511, opened 2015-09-25
Hide start proposal button when active proposal

HSalmon:
Eeeeep, sorry, I think I broke this when I hid the member actions from visitors


gdpelican:
Thanks @HSalmon !

PR #2512, opened 2015-09-26
More better invitations

gdpelican:
I had a great time on the :steam_locomotive:! 

List of improvements:
- API no longer returns multiple memberships for a single user
- Don't include the member's inviter in list of invitables
- Can now see members from groups you're a member of (not just groups you can invite people to)
- Invitation searching works after initial selection
- Fix avatars for invitables
- Fix layout for current invitations
- Improve performance of querying for visible members


app/extras/queries/visible_invitable_memberships.rb
```diff
@@ -17,15 +17,19 @@ def __setobj__(obj)
...
+              .where(user_id: visible_user_ids, archived_at: nil, is_suspended: false)
               .search_for(@query)
               .limit(@limit)
   end
 
-  def visible_membership_ids
-    @visible_member_ids ||= @user.groups.flat_map { |g| g.membership_ids }
-                                        .reject   { |id| group_membership_ids.include? id }
-                                        .uniq
+  def visible_user_ids
```
gdpelican: This is mildly fancy SQL, but oh-so-much faster than what we had before

It's saying
"Give me all the (distinct) user ids where if you look for a membership with that user and the given group, it doesn't exist."

This actually wants to be a query for (unique) users I think (the idea of querying invitables as memberships was a little misguided probably), but not quite ready to make that full switch just yet.

app/serializers/membership_serializer.rb
```diff
@@ -4,4 +4,9 @@ class MembershipSerializer < ActiveModel::Serializer
   has_one :group, serializer: GroupSerializer, root: 'groups'
   has_one :user, serializer: UserSerializer, root: 'users'
   has_one :inviter, serializer: UserSerializer, root: 'users'
+
+  def include_inviter?
+    [nil, true].include? (scope || {})[:include_inviter]
```
gdpelican: This is a little trick which means that we include the inviter if there is no scope (most of the time), or if we explicitly set 'include_inviter' to true, but not if we set `include_inviter` to be explicitly false

IE:
```
MembershipSerializer.new(membership, include_inviter: false).as_json
```

won't have the inviter included (we were having the inviter come back as an invitable user before, because it was included in the `user` query results)

gdpelican:
We don't have to use the Loomio bot for the default avatar, he was just the handiest thing lying around.

gdpelican:
@robguthrie This is ready for a look, and ideally to have someone pull it down and try it out... it's _way_ better than before.

app/controllers/api/memberships_controller.rb
```diff
@@ -9,7 +9,7 @@ def join_group
 
   def invitables
     @memberships = page_collection visible_invitables
-    respond_with_collection scope: { q: params[:q] }
+    respond_with_collection scope: { q: params[:q], include_inviter: false }
```
robguthrie: What does include_inviter: false do?

app/extras/queries/visible_invitable_memberships.rb
robguthrie: pretty flash sql there

app/controllers/api/memberships_controller.rb
gdpelican: We had a bug before where inviters were being included in the search results, so if I invited you, then my user would come back when searching for 'rob'.

This makes it so that the inviter (which, in this context, isn't something we really care about), doesn't get returned as a user with the json payload.

robguthrie: thank you. this looks great.

PR #2516, opened 2015-09-28
locales update, just es

mixmix:
@robguthrie 
:pineapple:


gdpelican:
Thanks @mixmix, let's get :beers:!

PR #2519, opened 2015-09-29
tell intercom which users are on beta

rdbartlett:
Minor tweak. Currently beta users turn up in Intercom as "angular=unknown".


rdbartlett:
( I have done no testing )

PR #2523, opened 2015-09-30
Fix inbox unread count

gdpelican:
The current bug:

- We serialize out discussions in the inbox query, and discussions in the notifications query on page load.
- Discussions on the inbox query have discussion readers attached to them, notifications do not
- [BUG]: the notifications query has discussion_reader_id keys on them set to null  (So, it's `{ id: 101, discussion_reader_id: null }` but should be `{id: 101 }`)
- This means that when notifications serialize out, they're overwriting the existing discussion_reader_id as null, which means we're not including those discussions in inbox (because the inbox won't include any discussions which have a null discussion reader id)

All due to a silly oversight on my part, heh; this is basically a typo fix. :heart:


robguthrie:
nice work! thank you

PR #2527, opened 2015-09-30
fix beta inbox count when angular. also errant emails

robguthrie:


app/queries/users_by_volume_query.rb
```diff
@@ -25,12 +20,16 @@ def self.mute(discussion)
...
   private
-  def self.volume_and_discussion(volume, discussion)
+  def self.base_query(discussion)
     User.
       active.
       distinct.
       joins("LEFT OUTER JOIN discussion_readers dr ON (dr.user_id = users.id AND dr.discussion_id = #{discussion.id})").
       joins("LEFT OUTER JOIN memberships m ON (m.user_id = users.id AND m.group_id = #{discussion.group_id})").
-      where('m.id IS NOT NULL AND dr.volume = :volume OR (dr.volume IS NULL AND m.volume = :volume)', { volume: DiscussionReader.volumes[volume] })
+      where('m.id IS NOT NULL AND m.archived_at IS NULL')
```
robguthrie: uh.. important change here is checking that membership is not archived. I want to know if there is a way I could have done .merge(Membership.active) 

Any ideas @gdpelican?

PR #2528, opened 2015-09-30
Build previous proposals card and page

HSalmon:


lineman/app/app.coffee
```diff
@@ -69,8 +69,9 @@ angular.module('loomioApp').controller 'ApplicationController', ($scope, $filter
     {path: '/m/:key/votes/new', component: 'proposalRedirect' },
     {path: '/g/:key/memberships', component: 'membershipsPage'},
     {path: '/g/:key/membership_requests', component: 'membershipRequestsPage'},
+    {path: '/g/:key/previous_proposals', component: 'previousProposalsPage'}
     {path: '/g/:key', component: 'groupPage' },
-    {path: '/g/:key/:stub', component: 'groupPage' }
+    {path: '/g/:key/:stub', component: 'groupPage' },
```
gdpelican: No trailing commas!

app/controllers/api/motions_controller.rb
```diff
@@ -24,6 +24,12 @@ def update_outcome
     respond_with_resource
   end
 
+  def closed
+    load_and_authorize :group
+    @motions = @group.closed_motions.order(closed_at: :desc)
```
gdpelican: Ah, I'd like to separate this out into `instantiate_collection`, so that we keep the pagination / timeframing capabilities that all the other endpoints have. I'll tap up a recommendation on how to do that in a bit.

lineman/app/models/group_model.coffee
```diff
@@ -17,6 +17,13 @@ angular.module('loomioApp').factory 'GroupModel', (BaseModel, AppConfig) ->
       @hasMany 'subgroups', from: 'groups', with: 'parentId', of: 'id'
       @belongsTo 'parent', from: 'groups'
 
+    proposals: ->
```
gdpelican: Pretty sure this wants to be a view, I think you can just go

```
@hasMany 'proposals`, from: 'proposals', with: 'groupId', of: 'id'
```

(You may need to serialize out the groupId in the motionSerializer)

lineman/app/components/group_page/group_page.haml
```diff
@@ -34,3 +34,4 @@
         %membership_requests_card{group: 'groupPage.group'}
         %members_card{group: 'groupPage.group'}
         %subgroups_card{group: 'groupPage.group'}
+        %group_previous_proposals_card{group: 'groupPage.group'}
```
gdpelican: Ah, so I think the way to do this cleanly is to have 

- a method on the group_model, which gives the closed proposals (you have this already I think)
- a `previous_proposal_collection` directive, which accepts a series of proposals
- a `previous_proposal` collection, which accepts a proposal

Similar to the `thread_preview_collection` and `thread_preview` directives

```
%proposals_collection{proposals: 'group.closedProposals()'}
```
```
%proposals_collection{proposals: 'discussion.nonActiveProposals()'}
```

Then, the proposal_collection.haml looks like
```
.thread-proposal-collection{ng-repeat: 'proposal in proposals'}
```

lineman/app/components/previous_proposals_page/previous_proposals_page.coffee
```diff
@@ -0,0 +1,11 @@
+angular.module('loomioApp').controller 'PreviousProposalsPageController', ($scope, $rootScope, $routeParams, Records) ->
+  $rootScope.$broadcast('currentComponent', { page: 'previousProposalsPage'})
+
+  Records.groups.findOrFetchById($routeParams.key).then (group) =>
```
gdpelican: We'll want a permissions check on here to make sure you can view stuff for this group; similar to `memberships_page_controller.coffee`, I reckon.

lineman/app/components/previous_proposals_page/previous_proposals_page.haml
```diff
@@ -0,0 +1,5 @@
+%main.previous-proposals-page.main-container
+  .previous-proposals-page__previous-proposals
+    %h2.lmo-h2 Previous proposals
```
gdpelican: Translation?

app/controllers/api/motions_controller.rb
gdpelican: alright, so `instantiate_collection` can take a block, which will modify the `visible_records` in whatever way you'd like. So, for example

```
def index
  instantiate_collection { |collection| collection.where(private: true) }
end

def visible_records
  Discussion.all
end
```

Will return all private discussions (it's magic, but not too magic)

So, here, we want something like

```
def index
  load_and_authorize :discussion
  instantiate_collection { |collection| collection.where(discussion: @discussion).order(:created_at) }
  respond_with_collection
end

def closed
  load_and_authorize :group
  instantiate_collection { |collection| collection.where(discussion_id: @group.discussion_ids).order(closed_at: :desc) }
end

def visible_records
  Queries::VisibleMotions.new(user: current_user, groups: current_user.groups)
end
```

app/serializers/motion_serializer.rb
```diff
@@ -18,6 +17,7 @@ class MotionSerializer < ActiveModel::Serializer
 
   has_one :author, serializer: UserSerializer, root: 'users'
   has_one :outcome_author, serializer: UserSerializer, root: 'users'
+  has_one :discussion
```
gdpelican: I'd like to not include this all the time, since it'll slow down a bunch of things, and make our json payloads much bigger (since including the discussion also includes the group, includes the author, etc.). The way we can include this conditionally:

```
has_one :discussion

def include_discussion?
  scope[:include_discussion]
end

def scope
  super || {} # (this simply ensures that the scope is not nil)
end
```
And in the controller:
```
respond_with_collection scope: { include_discussion: true }
```

lineman/app/app.coffee
```diff
@@ -69,8 +69,9 @@ angular.module('loomioApp').controller 'ApplicationController', ($scope, $filter
     {path: '/m/:key/votes/new', component: 'proposalRedirect' },
     {path: '/g/:key/memberships', component: 'membershipsPage'},
     {path: '/g/:key/membership_requests', component: 'membershipRequestsPage'},
+    {path: '/g/:key/previous_proposals', component: 'previousProposalsPage'}
```
gdpelican: Hmm, I'm not in love with this route name; what about `/decisions` or `/record` or `/decision_history`?

gdpelican:
Yep, so you're totally right that our primary hangup at the moment is about duplication in the previous proposals arena; I think a %previous_proposals directive which accepts a series of proposals will help solve that. Great work so far!

lineman/app/models/group_model.coffee
HSalmon: I tried this and it didn't work, @robguthrie can shed some light as to why. We had to do it manually in the end.

robguthrie: yea.. you're both right. Hannah and I discussed either doing a manual hasManyThrough or putting group_id in to the serializer. 

I worried that groupId might cause n+1 lookups so thought we should try this... the caveat was that ended up hasOneing discussion into the motionSerializer anyway.. so reflecting on this we could have also gone the other way... ie: it's worth trying that out @gdpelican...

app/controllers/api/motions_controller.rb
robguthrie: Thanks for explaining this James. For some reason I've had a hard time understanding this in the past.

Lemme see if I've got it right:
visible_records defines the collection
instantiate_collection applies any conditions you want

?

gdpelican: Yes!

The idea is that `visible_records` handles the privacy, ie, it's the subset of records that the user has access to, and the block in `instantiate_collection` defines 'what records I want right now', ie, all the motions that are closed.

lineman/app/models/group_model.coffee
gdpelican: Hm, the discussion gets included because the previous proposal card includes the discussion's title? I wonder if we could serialize out `discussionTitle` instead of having to include the whole discussion, the way we do with `parentAuthorName` on a comment..

lineman/app/components/group_page/group_previous_proposals_card/group_previous_proposals_card.haml
```diff
@@ -0,0 +1,17 @@
+%section.group-previous-proposals-card{aria-labelledby: 'group-previous-proposals-card__title'}
+  %h2#group-previous-proposals-card__title.lmo-card-heading{translate: 'group_previous_proposals_card.heading'}
```
HSalmon: So, @gdpelican what you've proposed would mean renaming proposal_collapsed to previous_proposal, using it everywhere, including on the group_previous_proposals_panel and modifying it according to what's displayed? My hesitation with that is that the group_previous_proposals_panel only has the pie and the title, and the proposals are not links. Modifying it conditionally sounds complex...but maybe it isn't? Would be keen to work with you and @robguthrie to work out the best way...

My first thought was to extract the pie with position (the css for the proposal_collapsed pie and thread_preview is pretty much the same I think). But I don't know how to structure the css so it looks ok by itself (rather than in the context of a proposal preview or thread preview)

HSalmon:
screen shot 2015-10-02 at 5 00 06 pm

screen shot 2015-10-02 at 5 00 15 pm

HSalmon:
@gdpelican, @robguthrie and I made a bunch of progress today. I'm happy to implement your other suggested changes over the weekend too and see this through.

app/serializers/motion_serializer.rb
robguthrie: Thank you. Other option is we take your first advice and put group_id in to MotionSerializer. I think we'll do that, but fall back to this if we need discussions in the end.

lineman/app/app.coffee
robguthrie: Hrrm.. we don't know if they were actually "decisions" given the way proposals are used..

gdpelican:
Cool! @HSalmon I'm thinking to add the group theme changes in #2530 to this branch, since you said you were having trouble dealing with the group theme and that will make it much easier.

gdpelican:
Hey @HSalmon I added the group theme to the page, and made the proposal accordion a little more flexible; maybe sometime this week we can have a go at generalizing it further and working out any further kinks.

rdbartlett:
Hey the only place I'm expecting a proposals accordion is on a Thread Page.

robguthrie:
currently also used on group_previous_proposals_page.

rdbartlett:
Design was intended to show that the group_previous_proposals_page is just
an index of links. I don't think the accordion makes sense anywhere other
than the Thread Page because there's not enough context around it.

robguthrie:
yep, happy to take more design lead on this. right now we've done it the
easiest way possible.

gdpelican:
This is very much a WIP. We can make it flexible enough to handle all the
cases, but the idea of a 'collection of previous proposals' is used in
enough places that we want to extract it out into its own thing so we're
not making 10 islands which are slightly different.

HSalmon:
What more do it need?

PR #2530, opened 2015-10-03
Generalize group theme

gdpelican:
This applies a directive to allow an easy way to put the group theme on new pages.

QUESTIONABLE BONUS:

I put the group theme on the thread page as well, per the mockups here:
https://projects.invisionapp.com/share/MV4B7FARB#/screens/103662393

In action:
screen shot 2015-10-03 at 10 44 43 am

I know we're thinking about revamping the thread page layout, but it was a small bit of work and looks quite nice imho. #IterateDontIncinerate


lineman/app/components/group_page/group_theme/group_theme.haml
```diff
@@ -0,0 +1,36 @@
+.group-page__theme
```
robguthrie: .group-theme and drop .group-page from everywhere

robguthrie:
oh man. this is so great. you rule

PR #2532, opened 2015-10-03
Remove puts messages from CI

gdpelican:


gdpelican:
Okay, this didn't work and I don't know why. It looks like we're pulling down the right / most recent version of loomio/private_pub now, but Travis is still spitting out those puts messages on build. ¯\_(ツ)_/¯

robguthrie:
i've fixed this buddy

PR #2533, opened 2015-10-03
New groups start on angular

gdpelican:


rdbartlett:
Shower thought: can you confirm this won't affect subgroups? I.e. Beta user creates subgroup will still work as normal, but beta user creates group will happen in angular.

gdpelican:
Cant confirm atm, but I'm pretty sure yes. This only affects the
StartGroupController#new action, and only if you're logged in (so you can
still create a non angular group while logged out from the try it page I
believe.. Didnt have a design for that bit.)

rdbartlett:
Try It page is decommissioned - you mean the logged out start_group page?

gdpelican:
Yeah the thing that happens when you click 'try out Loomio' on the marketing page.

gdpelican:
What's the source of all these rspec errors? Can't seem to make sense at a quick glance of why there's so many and what they're pointing at.

gdpelican:
FYI I just pushed a line onto this branch to reset the chargify_success parameter on modal open.

PR #2534, opened 2015-10-03
Don't skip booting angular UI when visiting /inbox

gdpelican:
Currently visiting `loomio.org/inbox` renders a weird half-baked version of the old beta interface, even if you're on angular.

This is what the protractor tests are complaining about... not sure when this broke?


gdpelican:
Ah, it's here:
https://github.com/loomio/loomio/commit/4941b3f132ae0b6f1cf0afa42cf55a4b6144decc

Why do we need to skip the angular UI boot there, @robguthrie ?

robguthrie:
oh this was my cheap hack to fix the inbox count turning into a webpage when visiting explore.

PR #2535, opened 2015-10-04
Use Loomio as feedback form instead of Google forms

gdpelican:
Switch back to beta form:
screen shot 2015-10-04 at 11 02 08 am

Give feedback form:
screen shot 2015-10-04 at 11 02 18 am

Admin panel:
screen shot 2015-10-04 at 11 03 36 am

### DIFF
--- a/lineman/app/app.coffee
+++ b/lineman/app/app.coffee
@@ -37,7 +37,7 @@ angular.module('loomioApp', ['ngNewRouter',
     $compileProvider.debugInfoEnabled(false);
 
 # Finally the Application controller lives here.
-angular.module('loomioApp').controller 'ApplicationController', ($scope, $filter, $rootScope, $router, KeyEventService, ScrollService, AnalyticsService, CurrentUser, MessageChannelService, IntercomService) ->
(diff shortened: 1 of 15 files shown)


rdbartlett:
exquisite

lineman/app/app.coffee
```diff
@@ -56,6 +56,9 @@ angular.module('loomioApp').controller 'ApplicationController', ($scope, $filter
 
   $scope.keyDown = (event) -> KeyEventService.broadcast event
 
+  if $location.search().feedback_given?
```
gdpelican: NB: this will cause a merge conflict with the new BootService in the 'new groups start on angular' PR.

rdbartlett:
We had a chat about this in planning yesterday. It's not a priority, but if you want to finish and deploy, it would be good to change the language so we get better bug reports. I'm thinking something like this, for both pages:

## Thanks for trying the new Loomio

Have some feedback for us? We'd love to hear your input!

If you're reporting a problem, the more detail you can give, the better. If you can tell us how we can reproduce the bug, we can fix it :)

gdpelican:
Gonna hold this for now, although if I see some more negative feedback about it come through in the google form (we've had ~3-4 comments about it so far), I may just merge this.

PR #2536, opened 2015-10-05
fixes after upgrade to el capitan

robguthrie:


robguthrie:
merging this now. tested on production... including rmagick test

PR #2537, opened 2015-10-05
Add contact us option to user dropdown

gdpelican:
This:
screen shot 2015-10-05 at 5 45 31 am

And this:
screen shot 2015-10-05 at 5 45 40 am

### DIFF
--- a/app/controllers/contact_messages_controller.rb
+++ b/app/controllers/contact_messages_controller.rb
@@ -1,5 +1,6 @@
 class ContactMessagesController < BaseController
   skip_before_filter :authenticate_user!
+  skip_before_filter :boot_angular_ui, only: :new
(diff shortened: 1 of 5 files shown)


rdbartlett:
Dear James. Please leave some jobs for everyone else. You are writing all the code.

robguthrie:
sokay with me. plenty more work to do

rdbartlett:
FYI James I just pushed up a little a11y improvement. I wonder if we should override the %i directive so we don't have to specify `aria-hidden: 'true'` all the time.

robguthrie:
great idea.

gdpelican:
@rdbartlett I pulled those aria changes out in favor of Rob's shiny new directive here: #2543

PR #2540, opened 2015-10-05
render proposal descriptions with Markdown in emails

rdbartlett:
I have not tested this but I think it is good. Can someone smoke it please?

### DIFF
--- a/app/views/thread_mailer/_motion.html.haml
+++ b/app/views/thread_mailer/_motion.html.haml
@@ -13,11 +13,11 @@
         - if motion.closed?
           %b= t(:closed) + " " + time_formatted_relative_to_age(motion.closing_at)
           - if motion.outcome.present?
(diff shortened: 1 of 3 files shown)


app/views/thread_mailer/_motion.html.haml
```diff
@@ -13,11 +13,11 @@
         - if motion.closed?
           %b= t(:closed) + " " + time_formatted_relative_to_age(motion.closing_at)
           - if motion.outcome.present?
-            = render_email_plaintext motion.outcome
```
robguthrie: looking over this.. I think you should be using render_rich_text(text, true) I think render_email_markdown is just for yesterday on loomio

rdbartlett: cool I updated this and then force pushed over the top. Should be good to go :)

PR #2541, opened 2015-10-06
rewrite invitations controller test

robguthrie:


gdpelican:
Cool. I think it might be good to have a notion of an invitation expiring as well (10 days from when it's created maybe?), so we check that you can't use an old invitation as well as one that has already been accepted.

gdpelican:
(I know that's scope creep so ignore at will, but would be nice-to-have)

gdpelican:
PS so nice seeing those protractor tests all on one line now, whee!

PR #2543, opened 2015-10-06
richards great idea

robguthrie:


gdpelican:
Looks good to meee!

Did you consider extracting out the `fa` class as well, so it could be

```
%icon{icon: 'plus', size: 'lg'}
```
?

robguthrie:
no, but that's also a good idea.

rdbartlett:
That is such a good idea.

robguthrie:
merging this for now. Happy to improve it i -> icon later.

PR #2544, opened 2015-10-06
repatriate navbar orphans

rdbartlett:
FYI there are a small number of predefined ["Landmark" roles](http://mcdlr.com/wai-aria-cheatsheet/), like banner, search, main, navigation, etc. 

When we have a region that doesn't make sense to be identified with a Landmark role, then we just use aria-label and call it whatever we like. 

That's why there's `role: 'banner'` and `aria-label: 'user-actions'` in this pull request.


robguthrie:
Thank you. I've now written a table of contents for the loomio-app gitbook. this will be added to the Accessiblity section of that book.

PR #2547, opened 2015-10-07
add help links and tidy up group page cards

rdbartlett:
# Changes

* add help link to user dropdown
* add help link to group page help card
* make group page previous proposals list items link to previous proposals page
* make group page membership requests list items link to membership requests page
* standardise all the lmo-card-minor-action links on the group page cards

# Before

before

# After 

after

### DIFF
--- a/lineman/app/components/utilities.scss
+++ b/lineman/app/components/utilities.scss
@@ -347,3 +347,15 @@ label.label-with-details{
   line-height: 35px;
   text-decoration: underline;
 }
+
+.lmo-card-minor-action{
+  @include fontHelveticaLight;
+  color: $grey-on-white;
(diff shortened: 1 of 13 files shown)


lineman/app/components/utilities.scss
```diff
@@ -347,3 +347,13 @@ label.label-with-details{
   line-height: 35px;
   text-decoration: underline;
 }
+
+.lmo-card-minor-action{
```
robguthrie: Heyhey @rdbartlett ... why does this around the a rather than on the a itself?

rdbartlett: Good point. Fixed now and overwritten.

PR #2548, opened 2015-10-07
reinstate threads card

rdbartlett:
Possibly not the best approach to this? it is a bit tricky because the threads list on Dashboard is it's own card, but the threads list on the Group page is inside a card. This works at least. Good enough to merge?

## Before

screen shot 2015-10-07 at 5 07 06 pm

## After

screen shot 2015-10-07 at 5 07 38 pm

### DIFF
--- a/lineman/app/components/group_page/discussions_card/discussions_card.scss
+++ b/lineman/app/components/group_page/discussions_card/discussions_card.scss
@@ -1,3 +1,7 @@
+.discussions-card{
+  @include card;
+}
+
 .discussions-card__new-thread-button{
   @include fontSmall;
   float: right;
--- a/lineman/app/components/group_page/discussions_card/discussions_card.scss
+++ b/lineman/app/components/group_page/discussions_card/discussions_card.scss
@@ -1,5 +1,9 @@
 .discussions-card{
-  @include card;
+  @include cardNoPadding;
+}
+
+.discussions-card__header{
+  margin: $cardPaddingSize;
 }
 
 .discussions-card__new-thread-button{
--- a/lineman/app/components/thread_preview_collection/thread_preview_collection.scss
+++ b/lineman/app/components/thread_preview_collection/thread_preview_collection.scss
@@ -3,3 +3,10 @@
   margin-bottom: $cardPaddingSize;
   margin-top: $thinPaddingSize;
 }
+
+.discussions-card__list .thread-previews-container{
+  @include cardNoPadding;
+  margin: $thinPaddingSize (-1 * $cardPaddingSize);
+  border-width: 1px 0 0 0;
+  box-shadow: none;
+}
--- a/lineman/app/components/thread_preview_collection/thread_preview_collection.scss
+++ b/lineman/app/components/thread_preview_collection/thread_preview_collection.scss
@@ -1,12 +1,5 @@
-.thread-previews-container {
+.inbox-page__groups .thread-previews-container, .dashboard-page__collections .thread-previews-container {
   @include cardNoPadding;
   margin-bottom: $cardPaddingSize;
   margin-top: $thinPaddingSize;
-}
-
-.discussions-card__list .thread-previews-container{
-  @include cardNoPadding;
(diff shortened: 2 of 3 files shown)


lineman/app/components/thread_preview_collection/thread_preview_collection.scss
```diff
@@ -3,3 +3,10 @@
   margin-bottom: $cardPaddingSize;
   margin-top: $thinPaddingSize;
 }
+
+.discussions-card__list .thread-previews-container{
+  @include cardNoPadding;
+  margin: $thinPaddingSize (-1 * $cardPaddingSize);
```
robguthrie: Is the negative margin the best thing to do? I've tried to keep negative margins out of the CSS so far.. and it's kind of jangy to run in to them sometimes and change how I think about doing CSS..

Happy to use em if you think it's the best thing to do but I don't want to end up with things breaking when moved between cards.

rdbartlett:
This version seems less ugly eh @robguthrie

lineman/app/components/thread_preview_collection/thread_preview_collection.scss
```diff
@@ -1,4 +1,4 @@
-.thread-previews-container {
+.inbox-page__groups .thread-previews-container, .dashboard-page__collections .thread-previews-container {
```
robguthrie: This is almost great. I'm wondering why there are qualifiers here. I mean, are there thread-previews-containers which won't work with this css? if so, why? If not.. then we should be able to drop the qualifiers here and just use .thread-previews-container as the selector

lineman/app/components/group_page/discussions_card/discussions_card.scss
```diff
@@ -1,3 +1,11 @@
+.discussions-card{
+  @include cardNoPadding;
+}
+
+.discussions-card__header{
```
robguthrie: have you seen .lmo-card-header-padding?

lineman/app/components/thread_preview_collection/thread_preview_collection.scss
rdbartlett: It's because the thread-previews-container on Inbox and Dashboard behave
like cards, but on the Group page it lives inside a card. Not sure the best
way to solve that

lineman/app/components/group_page/discussions_card/discussions_card.scss
robguthrie: actually... I'm not sure where it is.. but basically there are 3 or 4 main kinds of "padding on these sides of things" classes that I'm trying to group in utilities with the lmo-card- prefix. If what you're doing here is generic card padding that could be reused by other things consider making it a lmo-card-something

robguthrie:
Thanks for this Rich. .sorry If I'm getting boring with css enforcement

rdbartlett:
It's not boring. This is a bit of a tricky one because it is revealing the style inconsistency with how we've used cards on Dash/Inbox and everywhere else. 

I have had a few cracks at it and I think they are all compromised though, so I'm happy enough with this as an improvement.

lineman/app/components/group_page/discussions_card/discussions_card.scss
rdbartlett: .lmo-card-heading-padding doesn't quite work because it puts too much padding on the bottom.

lineman/app/components/thread_preview_collection/thread_preview_collection.scss
robguthrie: Don't need to change this now.. (i'll include it in the tech-manual) but I think with my preferred language "container" would be "panel" a panel is something that can be put inside a card.

Also I reckon we might need to come up with a a rule in the future... something like all cards have no padding by default and you need to add your own with one of these handy mixins.

For now this is super. If you're happy then we can merge

rdbartlett: Yeah dudey, merge away :)

PR #2549, opened 2015-10-07
Allow email interceptors to be defined

gdpelican:
Just want to see how this fares in test land; not quite ready for review yet.

### DIFF
--- a/app/mailers/base_mailer.rb
+++ b/app/mailers/base_mailer.rb
@@ -27,4 +27,17 @@ class BaseMailer < ActionMailer::Base
   def from_user_via_loomio(user)
     "\"#{user.name} (Loomio)\" <notifications@loomio.org>"
   end
+
+  def send_single_mail(locale: I18n.locale, to:, subject:, **options)
+    I18n.with_locale(locale) { mail options.merge(to: self.class.interceptor_email || to, subject: subject) }
+  end
+
+  def self.send_bulk_mail(to:)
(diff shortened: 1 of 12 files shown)


app/models/user.rb
```diff
@@ -233,6 +233,11 @@ def self.helper_bots
     where(email: ['contact@loomio.org', 'contact@loom.io'])
   end
 
+  def self.email_interceptors
```
robguthrie: How keen are you on this bit? I like what I see in the mailers but I don't think this feature should involve stuff in models and queries.

app/mailers/base_mailer.rb
```diff
@@ -27,4 +27,21 @@ def email_subject_prefix(group_name)
   def from_user_via_loomio(user)
     "\"#{user.name} (Loomio)\" <notifications@loomio.org>"
   end
+
+  def send_single_mail(locale: I18n.locale, to:, subject:, **options)
+    I18n.with_locale(locale) { mail options.merge(to: self.class.email_interceptor || to, subject: subject) }
+  end
+
+  def self.send_bulk_mail(to:)
+    to = Array(User.loomio_helper_bot) if email_interceptor.present?
```
gdpelican: This is a bit weird; it's because the blocks passed to `send_bulk_mail` always call `send_single_mail` under the hood, so I'm less concerned with who the recipients are (because all of them will be sent to the interceptor anyway), and more concerned with the number of recipients (we want one so that starting a thread in a group doesn't clog the inbox with 100 emails.)

PR #2550, opened 2015-10-07
update thread form

rdbartlett:
## Changes
* add detail to privacy message on thread form
* only show group select when you're starting a new thread outside of a group page
* expose edit thread option on thread context card
* make "show undecided members" link a lmo-card-minor-action

image

## TODO

discussion_form.haml needs a refactor. Currently it looks like discussion.private is returning different results for new threads vs edited threads. So starting a new thread in Enspiral says it is public, then you edit it and it says it is private. (This bug is live on prod)


rdbartlett:
@robguthrie here's the partial fix for the thread form

gdpelican:
@rdbartlett I've just put in a fix for it here: #2551, I can push it up on this branch I think.

gdpelican:
Actually, there's not a merge conflict, so they can just stay two separate PRs.

lineman/app/components/discussion_form/discussion_form.coffee
```diff
@@ -1,11 +1,11 @@
 angular.module('loomioApp').factory 'DiscussionForm', ->
   templateUrl: 'generated/components/discussion_form/discussion_form.html'
   controller: ($scope, $controller, $location, discussion, CurrentUser, Records, AbilityService, FormService, KeyEventService) ->
-    $scope.showGroupSelect = true
+    $scope.showGroupSelect = false
```
gdpelican: Should be able to just drop this line; it'll automatically start as undefined which is falsey

rdbartlett:
Thanks James. Updated now.

PR #2555, opened 2015-10-08
Two column thread page

gdpelican:
screen shot 2015-10-08 at 12 12 59 pm


lineman/app/components/thread_page/thread_page.haml
```diff
@@ -1,52 +1,53 @@
-.loading-wrapper.container.main-container
+.loading-wrapper.lmo-column-container.main-container
   %loading{ng-if: '!threadPage.discussion'}
   %main.thread-page{ng-if: 'threadPage.discussion'}
     %group_theme{group: 'threadPage.discussion.group()', compact: 'true'}
 
-    %section.thread-context{aria-label: "{{ 'thread_context.aria_label' | translate }}"}
```
gdpelican: This looks like a lot but it's just shifting the cards into divs with `.lmo-column-left` and `.lmo-column-right` classes

rdbartlett:
Looks great James. I'm looking forward to getting in there and doing some tweaks to the top of the context card and proposal card to make it look a bit more slick. 

I'm happy with the width of the right column, it's a bit wider than the mockup which is good.

PR #2561, opened 2015-10-09
use correct method to subscribe to message channel

robguthrie:
@gdpelican tiny typo in otherwise wonderful work. FYI then forget.


gdpelican:
Woof, sorry. Is that all it was to fix live updating?

PR #2562, opened 2015-10-09
Minor ui tweaks

HSalmon:


config/locales/client.en.yml
```diff
@@ -50,7 +50,7 @@ en:
 
   start_proposal_card:
     title: 'Start a proposal'
-    helptext: Ready to make a decision? Proposals let you see how everyone feels about a particular course of action.
+    helptext: Ready to make a decision?
```
robguthrie: I really like the "See how people feel about a course of action" part.. I wonder if there is room for that bit, @rdbartlett

robguthrie:
Can you squash then I'm happy to merge.

config/locales/client.en.yml
rdbartlett: That text is duplicated on the start proposal modal so I don't think we are losing anything.

HSalmon:
@robguthrie - done

robguthrie:
:ear: ChEARS

PR #2569, opened 2015-10-11
Muted threads

gdpelican:


lineman/app/components/flash/flash.scss
```diff
@@ -14,6 +14,25 @@
...
+  display: table;
+}
+
+.flash-root__message {
+  display: table-cell;
+  padding-right: 10px;
+}
+
+.flash-root__action {
+  display: table-cell;
```
robguthrie: You're really into table-cell eh? Does it do something display:flex cannot?

lineman/app/components/thread_preview/thread_preview.haml
```diff
@@ -6,10 +6,13 @@
     %button.btn.btn-default.btn-sm.thread-preview__mark-as-read{ng-click: 'thread.markAsRead()', ng-class: '{disabled: !thread.isUnread()}'}>
       %i.fa.fa-check>
       %span{translate: 'dashboard_page.mark_as_read'}
-    %button.btn.btn-default.btn-sm.thread-preview__mute{ng-click: 'thread.changeVolume("mute")', ng-class: '{disabled: thread.volume == "mute"}'}>
+    %button.btn.btn-default.btn-sm.thread-preview__mute{ng-click: 'changeVolume("mute")', ng-show: 'thread.volume != "mute"'}>
       %i.fa.fa-volume-off
       %i.fa.fa-times>
       %span{translate: 'volume_levels.mute'}
+    %button.btn.btn-default.btn-sm.thread-preview__mute{ng-click: 'changeVolume("normal")', ng-show: 'thread.volume == "mute"'}>
```
robguthrie: How does the suggestion to abstract the comparision into a function land with you? IE: thread.isMuted()

lineman/app/components/thread_preview/thread_preview.coffee
```diff
@@ -3,8 +3,17 @@ angular.module('loomioApp').directive 'threadPreview', ->
...
       thread.activeProposal().lastVoteByUser(CurrentUser)
 
+    $scope.changeVolume = (volume) ->
+      $scope.previousVolume = $scope.thread.volume
+      $scope.thread.changeVolume(volume).then ->
+        FlashService.success "discussion.volume.#{volume}_message",
+          name: $scope.thread.title
+        , 'undo', $scope.undo
+
+    $scope.undo = -> $scope.changeVolume($scope.previousVolume)
```
robguthrie: awesome

lineman/app/components/thread_preview/thread_preview.coffee
```diff
@@ -3,8 +3,17 @@ angular.module('loomioApp').directive 'threadPreview', ->
...
   templateUrl: 'generated/components/thread_preview/thread_preview.html'
   replace: true
-  controller: ($scope, Records, CurrentUser, LmoUrlService) ->
+  controller: ($scope, Records, CurrentUser, LmoUrlService, FlashService) ->
     $scope.lastVoteByCurrentUser = (thread) ->
       thread.activeProposal().lastVoteByUser(CurrentUser)
 
+    $scope.changeVolume = (volume) ->
+      $scope.previousVolume = $scope.thread.volume
+      $scope.thread.changeVolume(volume).then ->
```
robguthrie: We've had a few conversations in the office about how we want less critical user interactions to return immediately rather than wait for the server response. Eg: Mark as read and mute/change volume. Doing this encourages users to do stuff quickly and they feel like the app is nice and fast.

Are you comfortable with updating thread.changeVolume to update the record locally before firing the network request? The flash would still depend on the request but the other parts of the UI (ie: the thread list) would update immediately.

lineman/app/components/flash/flash.scss
gdpelican: Don't have a strong opinion of table-cell vs flex (due to ignorance more than anything else); I just wanted to make it work the way `media-left` and `media-body` do, which uses table-cell.

lineman/app/components/thread_preview/thread_preview.coffee
gdpelican: Yeah, I think it's as simple as adding `$scope.thread.volume = volume` there.

lineman/app/components/thread_preview/thread_preview.haml
gdpelican: Sounds good to me.

lineman/app/components/flash/flash.scss
robguthrie: Right ok. Update on media  .. we're actively trying to avoid it. I really like display: flex ... I think it's worth using.  I'm ok/happy with this, but I kinda wish it was display: flex and we adopted that everywhere it made sense.

lineman/app/components/thread_preview/thread_preview.coffee
robguthrie: probably within the thread_model but yea. cool you agree

PR #2570, opened 2015-10-11
Simplify my_votes endpoint

gdpelican:
Hey @robguthrie I'd be interested in hearing your thoughts on this, as it relates to side loading records we need from the store.

Here I'm loading up the current user's vote when the discussion a) is being loaded with a reader, and b) has a current proposal, but I wonder if there's a better way?

(This should, btw, fix the weird bug where people's votes aren't being loaded correctly on the dashboard sometimes... basically I don't trust the `my_votes` endpoint very much at the moment.)

(Also don't merge me; I'm just pushing this up now to see how the tests fare)

### DIFF
--- a/app/controllers/api/votes_controller.rb
+++ b/app/controllers/api/votes_controller.rb
@@ -4,15 +4,7 @@ class API::VotesController < API::RestfulController
   alias :update :create
 
   def my_votes
-    @votes = if params[:discussion_id]
-      load_and_authorize :discussion
-      @discussion.votes.includes({motion: [:author, :outcome_author]}, :user).for_user(current_user).most_recent
-    elsif params[:proposal_ids]
-      ids = params[:proposal_ids].split(',').map(&:to_i)
-      current_user.votes.where(motion_id: ids)
-    else
-      current_user.votes.most_recent.since(3.months.ago).most_recent
-    end
+    @votes = current_user.votes.includes(:motion, :user).where(motion_id: motion_ids).most_recent
     respond_with_collection
   end
 
@@ -21,6 +13,10 @@ class API::VotesController < API::RestfulController
   def visible_records
     load_and_authorize :motion
     @motion.votes.most_recent.order(:created_at)
+  end
+
+  def motion_ids
+    params[:proposal_ids].split(',').map(&:to_i) if params[:proposal_ids]
   end
 
 end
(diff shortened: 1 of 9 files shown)


app/controllers/api/votes_controller.rb
```diff
@@ -4,15 +4,8 @@ class API::VotesController < API::RestfulController
...
-    @votes = if params[:discussion_id]
-      load_and_authorize :discussion
-      @discussion.votes.includes({motion: [:author, :outcome_author]}, :user).for_user(current_user).most_recent
-    elsif params[:proposal_ids]
-      ids = params[:proposal_ids].split(',').map(&:to_i)
-      current_user.votes.where(motion_id: ids)
-    else
-      current_user.votes.most_recent.since(3.months.ago).most_recent
-    end
+    load_and_authorize :discussion
```
robguthrie: if there is no discussion does this fail silently?

app/controllers/api/votes_controller.rb
```diff
@@ -4,15 +4,8 @@ class API::VotesController < API::RestfulController
...
+    @votes = current_user.votes.where(motion_id: motion_ids).most_recent
```
robguthrie: includes or preload(:motion, :user) I reckon

app/serializers/discussion_serializer.rb
```diff
@@ -44,26 +44,26 @@ def self.attributes_from_reader(*attrs)
...
+    reader.present? && active_proposal.present?
+  end
+
+  def active_proposal_vote
+    active_proposal.votes.find_by(user_id: reader.user_id)
   end
 
   def active_proposal
-    object.current_motion
+    @active_proposal ||= object.current_motion
```
robguthrie: I'm not quite following the @active_proposal bit because I can't see it defined anywhere. So it's an instance variable? Where would it be set/defined?

robguthrie:
Overall I think it's very neat. Nice improvement. I wonder if there is any behaviour change due to this.

app/controllers/api/votes_controller.rb
gdpelican: Good question. I think so, but will test it manually; was actually hoping that an e2e test would catch something related to this, but they went green.

app/serializers/discussion_serializer.rb
gdpelican: I just wanted to cache the `object.current_motion` because we're using it more than once now, and not sure if we're preloading the current motion with the discussion so don't want it to fire off a request twice.

The active proposal's already serialized out with the discussion, this just makes it so that we serialize out the active proposal and your vote for that proposal (if we're serializing out reader info with the discussion)

robguthrie: Oh.. now I see you're using ||=  previously I did not.. that's why I sound confused.

gdpelican:
Okay, this is a PR I can be happy with now. Now we just have the endpoint doing one thing, which is accepting a series of proposal ids and returning the current user's votes for them.

**chanting:** _smart client, dumb server! smart client, dumb server!_

robguthrie:
really great.

PR #2578, opened 2015-10-13
Fix proposal legend for mobile, update design

HSalmon:
screen shot 2015-10-13 at 2 22 05 pm

screen shot 2015-10-13 at 2 22 54 pm

### DIFF
--- a/lineman/app/components/thread_page/proposal_outcome_panel/proposal_outcome_panel.haml
+++ b/lineman/app/components/thread_page/proposal_outcome_panel/proposal_outcome_panel.haml
@@ -4,4 +4,4 @@
     %p.lmo-hint-text{translate: 'proposal_outcome_panel.set_outcome_hint'}
     %a.btn.btn-success.lmo-btn-icon.proposal-outcome-panel__set-outcome-btn{href: '', ng-click: 'openProposalOutcomeForm()', translate: 'proposal_outcome_panel.set_outcome_btn'}
   %p.proposal-outcome-panel__outcome{ng-show: 'proposal.hasOutcome()'} {{ proposal.outcome }}
-  %a.lmo-card-minor-action.proposal-outcome-panel__edit-outcome-link{href: '', ng-if: 'canUpdateOutcome()', ng-click: 'openProposalOutcomeForm()', translate: 'proposal_outcome_panel.edit_outcome_link'}
+  %button.lmo-card-minor-action.proposal-outcome-panel__edit-outcome-link{ng-if: 'canUpdateOutcome()', ng-click: 'openProposalOutcomeForm()', translate: 'proposal_outcome_panel.edit_outcome_link'}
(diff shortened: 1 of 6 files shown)


lineman/app/components/thread_page/proposal_outcome_panel/proposal_outcome_panel.haml
```diff
@@ -4,4 +4,4 @@
     %p.lmo-hint-text{translate: 'proposal_outcome_panel.set_outcome_hint'}
     %a.btn.btn-success.lmo-btn-icon.proposal-outcome-panel__set-outcome-btn{href: '', ng-click: 'openProposalOutcomeForm()', translate: 'proposal_outcome_panel.set_outcome_btn'}
   %p.proposal-outcome-panel__outcome{ng-show: 'proposal.hasOutcome()'} {{ proposal.outcome }}
-  %a.proposal-outcome-panel__edit-outcome-link{href: '', ng-if: 'canUpdateOutcome()', ng-click: 'openProposalOutcomeForm()', translate: 'proposal_outcome_panel.edit_outcome_link'}
```
gdpelican: Think this wants to be an a11y button

lineman/app/components/thread_page/proposal_pie_chart/proposal_pie_chart.scss
```diff
@@ -50,8 +38,8 @@
   color: $grey-on-white;
 }
 
-@media (min-width: $medium-max-width) and (max-width: $large-max-width) {
-  .proposal-pie-chart__legend{
-    display: block;
+@media (max-width: $medium-max-width) {
+  .proposal-pie-chart__legend {
```
gdpelican: This doesn't work for me between 760px and 840px width, and should be floating right from the pie chart from 480px - 760px. Not sure why we need to change this, actually.

Might consider trying `display: flex` out here.

lineman/app/components/thread_page/proposal_pie_chart/proposal_pie_chart.scss
```diff
@@ -7,40 +7,28 @@
...
-.proposal-pie-chart__icon--no{
-  background-color: $disagree-color;
+.proposal-pie-chart__label--no{
   border-color: $disagree-border-color;
 }
 
-.proposal-pie-chart__icon--block{
-  background-color: $block-color;
+.proposal-pie-chart__label--block{
   border-color: $block-border-color;
```
gdpelican: I'd really like to use the actual colors of the pie chart here, instead of their darkened counterparts... it's looks confusing to me because 'disagree' looks the same as 'block', and 'block' looks maroonish instead of red.

lineman/app/components/thread_page/proposal_pie_chart/proposal_pie_chart.scss
rdbartlett: FYI I'm planning to get in tomorrow and put a bit more design into how the
2 columns work on different screens. I think we should be able to have the
pie and legend sitting comfortably side by side on any screen.

PR #2583, opened 2015-10-13
different margins on different screens

rdbartlett:
## Big screen:

image

## Small screen:

image


robguthrie:
good to merge, matey?

rdbartlett:
yep fire away

PR #2586, opened 2015-10-14
[not ready] first pass at cleaning up 2 column thread page

rdbartlett:
Pretty happy with how this is going. Thread page complete, next step is to bring all the other pages up to this spec and then remove

```
$main-max-width: 640px;
$medium-max-width: 767px;
$large-max-width: 960px;
```

### DIFF
--- a/lineman/app/components/base.scss
+++ b/lineman/app/components/base.scss
@@ -6,7 +6,22 @@ body {
 }
 
 h1, h2, h3, h4{
-  @include fontReset; /* Y U SO RESET? */
+  @include fontReset;
+
+  /*
+  We reset styles on headings for accessibility.
+
+  Semantically we want to enforce a strict hierarchy...
+  h1
+   - h2
+   - h2
+     - h3
+  -h2
+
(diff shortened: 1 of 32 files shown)


rdbartlett:
FYI @gdpelican this is where I got to with the 2 col thread page. I think it is pretty cool. 

I started trying to generalise this out to the group page but realised we actually want different behaviour. Group page right column should always be 320px.

So if you're happy, we can merge this now and come back to do the generalising job, or hold off merging until we've sorted out 2 cols everywhere. As you like :)

gdpelican:
I don't necessarily think we need to generalize the two column thing, even long-term. The primary screens for me, even long term, would be

- Inbox / dashboard / wall
- Network-related stuff
- Group-related stuff
- Discussion-related stuff

And I reckon those will all likely deserve distinct layouts.

robguthrie:
good call

rdbartlett:
Alright this is all done now, I'm really happy with the result. Check it out @gdpelican.

Ready to go.

PR #2587, opened 2015-10-14
fix proposal closing at issues

robguthrie:
validate proposal closing in future if open.
fix time select am/pm.
fix permission for canEditProposal so "change closing time" dropdown action is available at the right time


app/models/motion.rb
```diff
@@ -244,13 +245,21 @@ def store_users_that_didnt_vote
   end
 
   private
+    def closes_in_future_unless_closed
```
robguthrie: this ok @gdpelican

PR #2590, opened 2015-10-14
User profile page

gdpelican:
Sorry, this turned into a bit of a lot. Lots of good stuff though. Things in this branch:

- Adding the user profile page
- Add counter cache for group members count
- Memberships controller refactor
- Don't N+1 when subscribing to a series of groups
- Make @mentions work as links to profile page


app/extras/model_locator.rb
```diff
@@ -5,6 +5,8 @@ def locate
       resource_class.find id_param
     elsif key_param = params[:"#{model}_key"]
       resource_class.find_by_key key_param
+    elsif model == :user
```
gdpelican: I think it turns out we need to be able to search by username in order to make @mentions work, because the markdown renderer needs to come up with a url (or some user information), and has only the username to go off of.

Other option would be loading all the mentioned users with a comment and doing Records.find(username: (text)) somewhere, but I think that's too much for a mention-heavy thread.

app/serializers/group_serializer.rb
```diff
@@ -57,4 +57,8 @@ def has_discussions
   def has_multiple_admins
     object.admins.count > 1
   end
+
+  def cover_photo
```
gdpelican: This mitigates the cover_photo N+1 some, but I'll want to go back at one point and find the right place for an eager load. I'm still thinking I really only want to load the cover photo if we're loading a page with a group theme, though.

app/services/message_channel_service.rb
```diff
@@ -20,27 +20,25 @@ class UnknownChannelError < StandardError
...
-    PrivatePub.subscription(channel: channel, server: ENV['FAYE_URL'])
+  def self.subscribe_to(user:, model:)
+    raise AccessDeniedError.new unless can_subscribe?(user: user, model: model)
+    PrivatePub.subscription(channel: channel_for(model), server: ENV['FAYE_URL'])
   end
 
-  def self.can_subscribe?(user:, channel:)
-    key = channel_key(channel)
-    case channel_type(channel)
-    when 'group'         then user.ability.can? :see_private_content, Group.find(key)
```
gdpelican: This was causing an N+1 when loading all the user's groups. I think it makes more sense for the rest of the world to not know about the concept of a 'channel' anyway; pass in the model and have the service handle it correctly.

app/services/message_channel_service.rb
```diff
@@ -20,27 +20,25 @@ class UnknownChannelError < StandardError
...
-    when 'notifications' then key.to_i == user.id
-    else                      raise UnknownChannelError.new
-    end
+  def self.subscribe_to_notifications_for(user)
+    PrivatePub.subscription(channel: "/notifications/#{user.id}")
   end
 
-  def self.channel_type(channel)
-    /\/(\w+)-(\w+)/.match(channel)[1]
+  def self.can_subscribe?(user:, model:)
```
gdpelican: At some point I want to be able to pass in a User here, with
```
case User then user.ability.can? :get_notifications_for, model
```

but I'll have to change the channel name to '/user-#{user_id}' to do that and didn't want to tread that just yet.

lineman/app/app.coffee
```diff
@@ -15,6 +15,9 @@ angular.module('loomioApp', ['ngNewRouter',
   renderer = new marked.Renderer()
   renderer.link = (href, title, text) ->
     "<a href='" + href + "' title='" + (title || text) + "' target='_blank'>" + text + "</a>";
+  renderer.paragraph = (text) ->
+    text = text.replace(/([\s(\[]+|^)@([a-zA-Z0-9]+)([\s)\]\,\?]|$)/g, "$1<a class='lmo-user-mention' href='/u/$2'>@$2</a>$3")
```
gdpelican: This regex isn't perfect for mention highlighting, but it's doing really well.

Only thing that doesn't seem to work atm is a tricky case around mentions with a period afterwards:

```
clone@loomio.org
```
vs
```
let's ask @loomio.
```

app/controllers/api/memberships_controller.rb
```diff
@@ -43,8 +55,8 @@ def remove_admin
   private
 
   def visible_records
-    load_and_authorize :group
-    Queries::VisibleMemberships.new(user: current_user, group: @group)
+    resource_class.includes(:user, :group, :inviter)
+                  .where("group_id IN (#{current_user.group_ids.join(',')}) OR groups.is_visible_to_public = 't'")
```
gdpelican: A group is visible to a user if a) they're in it, or b) it is visible to the public. Seems legit?

app/models/group.rb
```diff
@@ -105,6 +105,7 @@ class MaximumMembershipsExceeded < Exception
   has_many :memberships,
            -> { where is_suspended: false, archived_at: nil },
            extend: GroupMemberships
+  define_counter_cache(:memberships_count) { |group| group.memberships.count }
```
gdpelican: This moves the N+1 problem from every time we load the Group to everytime we update the Group... which is better but still not the best.

I'm seriously considering writing a gem to handle this with increment / decrement.

app/extras/model_locator.rb
robguthrie: Sorry I don't follow you. Can you explain more?

app/serializers/group_serializer.rb
robguthrie: I don't get it. cover_photo_* attributes are all on the group table already. So it does not look like a n+1 problem to me.

app/services/message_channel_service.rb
robguthrie: I think you should go for it.

lineman/app/app.coffee
robguthrie: That regex is really great!
Pasting it into https://regex101.com/  is fun. Explains how it works really nicely.

It looks like you just need to add a `\.` in there..

```
([\s(\[]+|^)@([a-zA-Z0-9]+)([\s)\]\,\?\.]|$)
```

However I just looked at what we do on the rails side.. twitter-text library looks like it deserves some consideration.

https://github.com/twitter/twitter-text/tree/master/js

We already use it in the back end.. https://github.com/twitter/twitter-text

robguthrie: Arrr I've found the right Readme now. Please read this: https://github.com/twitter/twitter-text/tree/master/js

Seems like we'd be kind of mad not to use it

app/extras/model_locator.rb
gdpelican: Yep, so I want this route to work:
```
/u/:username
```
as well as this one:
```
/u/:key/:username
```

Because, when we construct the urls for @mentions, we don't necessarily have that user loaded into the record store, and I'm a bit loathe to try to, server side, figure out what users are mentioned in a window of activity. The thing we do have is the username though, so I can take this:
```
<p>I'd like to hear @loomiobot 's opinion!</p>
```
and turn it into this:
```
<p>I'd like to hear <a href="/u/loomiobot">@loomiobot</a> 's opinion</p>
```
but I can't turn it into this:
```
<p>I'd like to hear <a href="/u/botkey/loomiobot">@loomiobot</a> 's opinion</p>
```
because I don't know the key for the user.

app/serializers/group_serializer.rb
gdpelican: If it's using a default cover photo, which reach out to the DefaultCoverPhoto of the group.

robguthrie: So... how crazy do you think it would be to put something like

```
default_scope { includes(:default_cover_photo) }
```

into group.rb

I've always ended up regretting using default_scope.

gdpelican: Yeah, heh. Had a similar thought. There's lots of cases where we load a
group and don't care about the cover photo, but I think an 'includes'
default_scope is way less likely to come back and bite us (or it will take
longer to come back and bite us, heh) than a 'where' default_scope

lineman/app/app.coffee
gdpelican: Ah yeah, that's good. I didn't see an angular package for it, but I think we can still use it somehow

gdpelican: Hrm... the twitter-text libraries validate usernames as twitter usernames, which would be slightly different from how we do usernames currently. For example, I think there's a character limit (14?) on twitter usernames. Maybe we just conform our usernames to twitter, but I'm a little wary of that.

gdpelican: Hm, additionally, it seems that the twitter urls (http://www.twitter.com, etc.) are hard-coded into the gem; we could change them, but we'd have to overwrite a frozen constant, or maybe fork the repo. Starting to wonder if I can find whatever regex they're using and steal it instead.

gdpelican: Can't help but wonder if this is the reason we don't have the auto-linking functionality in beta?

robguthrie: Thanks for looking at that option. I'm not sure it's the right way to go either. Happy to proceed with the simple path until we find a reason why not.

app/controllers/api/memberships_controller.rb
```diff
@@ -24,7 +36,7 @@ def autocomplete
     @memberships = Queries::VisibleAutocompletes.new(query: params[:q],
                                                      group: @group,
                                                      current_user: current_user,
-                                                     limit: 10)
+                                                     limit: 5)
```
robguthrie: I tend to want to include more rather than less because then there is less lag as the user gets more accurate

app/controllers/api/message_channel_controller.rb
```diff
@@ -4,13 +4,13 @@ class API::MessageChannelController < API::RestfulController
...
   def subscribe
-    @subscriptions = [MessageChannelService.subscribe_to(user: current_user, channel: params[:channel])]
+    @subscriptions = [MessageChannelService.subscribe_to(user: current_user, model: subscription_model)] if subscription_model
     respond_with_subscriptions
   end
 
   def subscribe_user
-    @subscriptions = current_user.groups.map { |group| MessageChannelService.subscribe_to(user: current_user, channel: channel_for(group)) }
-    @subscriptions <<                                  MessageChannelService.subscribe_to(user: current_user, channel: "/notifications-#{current_user.id}")
+    @subscriptions = current_user.groups.map { |group| MessageChannelService.subscribe_to(user: current_user, model: group) }
```
robguthrie: have you managed to test the message channel since you've done this?

app/extras/model_locator.rb
robguthrie: hrrrmm that's really good. Makes me think we should not have a key on user but rather use username ..

robguthrie:
this is great. sorry for creating merge conflicts for you. but I've stuck with ccc so you should not have big issues. feel free to merge when you think it's ready. :+1:

app/controllers/api/memberships_controller.rb
gdpelican: Good call.

app/controllers/api/message_channel_controller.rb
gdpelican: Yeah, works like a charm; it's nicely designed in that the frontend just listens to what the server tells it to, it has no concept of what the channels are named.

PR #2593, opened 2015-10-14
suggested language change for first comment

rdbartlett:


config/locales/en.yml
```diff
@@ -1036,13 +1036,14 @@ en:
...
-      Here are some things you might want to try to get things under way:
+      - Invite people to join your group
+      - Set your [profile photo](https://www.loomio.org/profile) so people can see your lovely face
+      - Introduce yourself in the [Welcome to '%{group_name}' thread]('%{thread_url}')
+      - Configure your [email settings](https://www.loomio.org/email_preferences) to keep up to date with your Loomio activity
 
-      - Set your [profile photo](https://www.loomio.org/profile)
-      - Configure your [email settings](https://www.loomio.org/email_preferences)
-      - Introduce yourself to the group in the [Welcome to %{group_name} thread](%{thread_url})
+      If you want to get the attention of a specific person in your group, type @ into the comment field, followed by their name :)
```
robguthrie: Also consider: 
If you want to direct a message to someone, or just get their attention, you can __mention__ them by typing "@" then their name into the comment field.

Also.. we should include "Who should be notified button like yammer" long term

rdbartlett:
nice. Done now.

PR #2594, opened 2015-10-14
Possible fix for broken thread preview title truncation

HSalmon:


robguthrie:
thanks but not necessary. :recycle:

PR #2595, opened 2015-10-15
Add spec for api motions controller close

HSalmon:


HSalmon:
@robguthrie, is this correct?

gdpelican:
This looks great.

PR #2596, opened 2015-10-15
Allow people to contact us when Intercom is blocked

HSalmon:


robguthrie:
yay!

lineman/app/services/intercom_service.coffee
```diff
@@ -40,8 +40,10 @@ angular.module('loomioApp').factory 'IntercomService', ($rootScope, $http, AppCo
           created_at: group.createdAt
 
     contactUs: ->
-      return unless @available()
-      window.Intercom.public_api.showNewMessage()
+      if @available()
+        window.Intercom.public_api.showNewMessage()
```
gdpelican: We'll want to inject $window here, so that if we decide to write tests for this, we can control what's on the window object.

Annoying I know, but we've been using $window everywhere else and I don't want to start changing now.

lineman/app/services/intercom_service.coffee
```diff
@@ -1,4 +1,4 @@
-angular.module('loomioApp').factory 'IntercomService', ($rootScope, $http, AppConfig, CurrentUser) ->
+angular.module('loomioApp').factory 'IntercomService', ($rootScope, $http, AppConfig, CurrentUser, $location, LmoUrlService) ->
```
gdpelican: Why inject $location here?

lineman/app/services/intercom_service.coffee
robguthrie: yea cool. thanks for pointing it out

PR #2597, opened 2015-10-15
Remove API::BaseController, use API token for authentication

gdpelican:
Just giving this a quick hoon; will be nice to have app-wide token auth for our API soon.


gdpelican:
This passed, and will allow us to accept logged-in requests without a session, which is sweet. :balloon:

Hooray for a test suite which tests all permission things!

robguthrie:
This is great! However no changes to the client yet? Any thoughts there?

Also I think it's worth bringing up this:
https://coderwall.com/p/kb97gg/secure-tokens-from-rails-5-to-rails-4-x-and-3-x

I reckon we could clean up a bunch of token generation methods with that.

robguthrie:
Finally.. I wonder if this is secure. The email api key stuff was intended to only be used as a means to perform a couple of safeish actions.

Email api keys have been sent out in emails for a long time. Just so people can reply to comments via email.

Do we want to consider what features something like https://github.com/lynndylanhurley/devise_token_auth would bring us in terms of token expiry etc?

robguthrie:
also FYI: @euglazer and I have had an interesting time use ng-token-auth and devise-token-auth in Cobudget.

gdpelican:
Haven't needed to make clientside changes just yet because we're not dealing with logging in / out; all of the use cases through there are where the user already has a session.

I had a quick look at serializing out the token, but yeah it should expire if we're going to do that.

robguthrie:
close for now?

gdpelican:
I'd really like to get this some attention at some point... I've been wanting to get this public API thing sorted for a while now, and I think we're close.

Maybe we can do it around when we do the explore page (not too far off I hope)? Which is when we'll want to really start exposing public content.

robguthrie:
Ok, but I'm confused about what this branch actually does.  Would be good to walk over it with you

gdpelican:
Alright, this should be good to go; I'll squishy it.

- Slight refactor of the translationsController, which will make it easier to handle fallback languages when we get there
- Remove API::BaseController because it doesn't do anything
- Remove stray console logs

gdpelican:
@robguthrie is there a way we can test that this still works for creating a comment by email? (or are you confident in the automated tests around that?) Other than that this should be golden.

gdpelican:
Actually, looking at the tests, they seem pretty legit to me, and the code changes are super minor. I'm going to merge.

PR #2598, opened 2015-10-15
Replace outbox branch with outbox remote

Em-AK:
I think it is more appropriate to say:
*add your fork as a new remote* **than** *add your fork as a new branch.*


gdpelican:
Right; thanks!

PR #2599, opened 2015-10-15
New vote webhook

gdpelican:


robguthrie:
cool to have a SiteColors module.

gdpelican:
Hmm.. I'm not sure why this passed under our repo, but not theirs... it seems like travis isn't setting the Chargify environment variables correctly under their remote... which is distressing because it would mean that basically no one can get tests to pass..

I'll have to sit down at some point and ensure we're not relying on those ENV variables for those tests, that it's got a default in secrets.yml or something.

gdpelican:
Okay, here's the culprit:
http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests

Solution is to not have those tests rely on ENV; I'll twiddle that up now.

gdpelican:
(FYI @juliagra)

robguthrie:
Seems like using Rails.application.secrets and the secrets.yml is a good
way to centralise all the configuration and defaults.

gdpelican:
Closing; we'll get the other one merged in.

PR #2600, opened 2015-10-15
Don't mark discussions without readers as unread

gdpelican:
Fixes transitioning to a thread from a search result you haven't seen before.


robguthrie:
thanks!

PR #2603, opened 2015-10-15
use rails.application.secrets for intercom env

robguthrie:


robguthrie:
I've done this to try and stop the problem of development env's sending test data to intercom

PR #2604, opened 2015-10-15
finish chargify env stuff

robguthrie:


gdpelican:
Clearly whatever I'm doing at 1am, it shouldn't be coding.

PR #2610, opened 2015-10-19
clean up dropdowns

rdbartlett:
Standardise different dropdown versions:

### dropdown-menu

Used for the most simple context menus, e.g. thread context, proposal options, comment options.

---

image

---

These are small on desktop and expand out to full width when you are on a tiny screen:

---

image

---

### dropdown-menu--with-icons

Used when you want to illustrate each option with an icon.

---

image

image

---

### dropdown-menu--with-details

Used when you have to explain what the options mean

---

image

---

* all of them are aligned right
* Should search and notifications dropdowns use the same pattern? Or is this good enough to merge and we can resolve that when we fix search results.


rdbartlett:
Travis errors. Have restarted the build.

PR #2613, opened 2015-10-19
Fix cmd + enter for forms, expand submit on enter to group forms

gdpelican:
- Also applies FormService to comment form

### DIFF
--- a/lineman/app/components/thread_page/comment_form/comment_form.coffee
+++ b/lineman/app/components/thread_page/comment_form/comment_form.coffee
@@ -1,35 +1,40 @@
 angular.module('loomioApp').directive 'commentForm', ->
-  scope: {comment: '='}
+  scope: {discussion: '='}
   restrict: 'E'
   templateUrl: 'generated/components/thread_page/comment_form/comment_form.html'
   replace: true
-  controller: ($scope, $rootScope, FlashService, Records, CurrentUser, KeyEventService, AbilityService, ScrollService) ->
-    discussion = $scope.comment.discussion()
-    group = $scope.comment.group()
+  controller: ($scope, $rootScope, FormService, Records, CurrentUser, KeyEventService, AbilityService, ScrollService) ->
 
     $scope.formInFocus = ->   $rootScope.$broadcast 'hideFeedbackForm'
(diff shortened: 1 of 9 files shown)


lineman/app/components/thread_page/comment_form/comment_form.coffee
```diff
@@ -1,35 +1,40 @@
 angular.module('loomioApp').directive 'commentForm', ->
-  scope: {comment: '='}
+  scope: {discussion: '='}
```
gdpelican: I always felt it was kinda weird we were passing a comment to the comment form, then pulling the discussion and group out of it. Now we're not doing that anymore!

lineman/app/components/thread_page/comment_form/comment_form.coffee
```diff
@@ -1,35 +1,40 @@
...
+      if $scope.comment.isReply()
+        'comment_form.messages.replied'
+      else
+        'comment_form.messages.created'
+    successMessageName = ->
+      if $scope.comment.isReply()
+        $scope.comment.parent().authorName()
+
+    $scope.init = ->
+      $scope.comment = Records.comments.build(discussionId: $scope.discussion.id)
```
gdpelican: This pattern solves the issue of FormService not updating its model dynamically.

personally I don't think the model of the form should be able to change while the user is completing it, so I'm happy with this as a way to reuse the same form again.

lineman/app/services/key_event_service.coffee
```diff
@@ -28,5 +29,5 @@ angular.module('loomioApp').factory 'KeyEventService', ($rootScope) ->
     submitOnEnter: (scope) ->
       @registerKeyEvent scope, 'pressedEnter', scope.submit, (active, event) =>
         (event.ctrlKey or event.metaKey) and
-        angular.element(active).scope() == scope and
+        angular.element(active).scope().$$isolateBindings == scope.$$isolateBindings and
```
gdpelican: Because things like `<form>` and `<textarea>` are directives in angular, this scope comparison wasn't working. What I really want is `scope.isChildOf(angular.element(active).scope())`, but couldn't find that anywhere and don't want to write my own.

(This check is there at all in order to prevent you from, say, submitting your comment while creating a new thread in a model)

PR #2620, opened 2015-10-20
fix issues with async timing

robguthrie:
@gdpelican just wanted to share that we don't want to use angular single binding.. it always ends up ruining things in my experience, and the perf improvement is not measurable.

You probably knew this but wanted to be sure. 

Also I'm finding that putting records directly onto the scope is just as bad as single binding for causing these timing issues. If we access records via methods on the scope then we don't have to check that 'membershipsLoaded' has happened etc.


gdpelican:
@robguthrie I think a simpler fix is to put an ng-if onto the cards which require memberships to be loaded, maybe with a loading spinner.

robguthrie:
ng-if membership? yea.. that is better. i'll redo this tomorrow

PR #2621, opened 2015-10-20
prevent trial card from showing sometimes after subscribing

robguthrie:


robguthrie:
@gdpelican feel free to say if you think this is dumb. We're just finding that sometimes the webhook is slower than the user so we want to hide that case from the user.

gdpelican:
@robguthrie  I'd rather put a line into the subscriptions controller which publishes the event out over the group channel...

```
if SubscriptionService.available?
  # case statement...
  MessageChannelService.publish group_from_reference, GroupSerializer.new(group_from_reference).as_json
  head :ok
```

robguthrie:
hrrmm.. won't that also have synchronisation issues? The app might not be loaded when the message is sent out as the user is leaving the site to complete the transaction.

robguthrie:
p.s. I do really like the idea of using the message channel more.

gdpelican:
Ah... maybe? My thinking was that these two things could happen in either
order:

A: Webhook hits subscriptions controller and updates group
B: User is redirected to Loomio.org and the angular app is loaded

If A happens before B, then when the angular app is loaded, it will
serialize out the correct group data
If B happens before A, then the angular app will load old group data, but
will receive a group update soon after when the webhook hits

I guess there's the potential for some little corner case in there like
- User is redirected to loomio
- Request to load angular app begins, and within the request:
  - Group is loaded into memory
  - Webhook hits, which updates the group in the database
  - Angular app is served, with old group data
Not sure that I know ruby threading well enough to know if that's actually
a thing or not.

Maybe we do both?

PR #2623, opened 2015-10-20
Expand latest proposal on thread page

gdpelican:
With active proposal:
screen shot 2015-10-21 at 1 09 40 am

With recent closed proposal:
screen shot 2015-10-21 at 1 09 57 am

### DIFF
--- a/app/models/webhooks/slack/base.rb
+++ b/app/models/webhooks/slack/base.rb
@@ -34,9 +34,9 @@ Webhooks::Slack::Base = Struct.new(:event) do
   def motion_vote_field
     {
       title: "Have your say",
-      value: "#{proposal_link(eventable, "agree")} · " +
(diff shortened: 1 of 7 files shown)


gdpelican:
Hey I wonder if we should maybe put a time limit (2 weeks? A month?) on the latest closed proposal being open, since the primary objective is to expose the set outcome feature and at that point people will likely stop caring about something which happened forever ago.

rdbartlett:
Yeah cool. Stick a one month timeout on it and we'll see how it feels?

gdpelican:
This is good for a look @robguthrie; sorry about the mildly unrelated webhook fix in there.

robguthrie:
Looks great. thanks for giving me time to see. Merge if you're happy

PR #2625, opened 2015-10-21
link to comment via timestamp

robguthrie:
No styling..  it only shows underline on hover (pictured) @rdbartlett ?

image


rdbartlett:
beautiful. is that underline only visible on hover?

robguthrie:
yes

PR #2627, opened 2015-10-21
big performance win: Only animate elements with animated class (not ready)

robguthrie:
Now we need to find where animations occur so we can add the "animated" class.

Check http://www.bennadel.com/blog/2935-enable-animations-explicitly-for-a-performance-boost-in-angularjs.htm for details


gdpelican:
Do we have any animations? I suppose the modal appearing... I can't think of any others?

robguthrie:
Ok well I suggest we merge it in, knowing any animations will break, and just fix them as we go? @rdbartlett

rdbartlett:
agreed

PR #2628, opened 2015-10-21
Fix the error of my ways

mhjb:
Before
===
image

After
===
image


mhjb:
I see this will actually require a bit more attention from me today.

PR #2629, opened 2015-10-21
Preselect group on invitation form if only one

HSalmon:


rdbartlett:
can we do this on Start Thread too?

lineman/app/components/start_menu/start_menu_option.coffee
```diff
@@ -16,9 +16,15 @@ angular.module('loomioApp').directive 'startMenuOption', ->
     if $scope.hotkey
       KeyEventService.registerKeyEvent $scope, $scope.hotkey, $scope.openModal
 
+    availableGroups = ->
+      _.filter CurrentUser.groups(), (group) ->
+        AbilityService.canAddMembers(group)
+
     $scope.invitePeopleGroup = ->
       if $scope.group and AbilityService.canAddMembers($scope.group)
```
gdpelican: What do you think about not duplicating the AbilityService call and doing _.contains availableGroups() instead?

```
$scope.group = _.first(availableGroups()) if availableGroups().length == 1
if _.contains availableGroups(), $scope.group
  $scope.group
else
  Records.groups.build()
```

robguthrie: might be a good idea.. but I think I'll merge it now anyway.

PR #2630, opened 2015-10-22
A11y october

HSalmon:


config/locales/client.en.yml
```diff
@@ -200,6 +200,7 @@ en:
       comments: 'Comments'
 
   start_menu:
+    title: 'Start menu'
```
rdbartlett: In the help docs we call this the Action Button

rdbartlett:
Merged this one into #2634

PR #2631, opened 2015-10-22
upgrade angular ui bootstrap

robguthrie:
moving from 13.3 to 14.2 and they have namespaced their classes.. which highlights that we've been hijacking the dropdown-menu class..  I've not fixed that, this just removes all the deprecation warnings


rdbartlett:
This is cool. Dropdowns are the best thing about bootstrap!

robguthrie:
modals are right up there

PR #2632, opened 2015-10-22
Use angular promise implementation

gdpelican:


gdpelican:
Feel pretty good about it actually, although it makes me wonder if we should consider getting some of those angular e2ez running on IE.

PR #2633, opened 2015-10-23
lose extra padding on one col pages

rdbartlett:
This fixes the 48px extra padding seen on the dashboard and inbox pages. `.group-theme` and `.lmo-row` were overwriting each other so I nested one inside the other.

### DIFF
--- a/lineman/app/components/group_page/group_theme/group_theme.haml
+++ b/lineman/app/components/group_page/group_theme/group_theme.haml
@@ -1,43 +1,42 @@
 .group-theme
-  .lmo-row
-    .group-theme__cover{ng-style:'coverStyle()'}
-      .group-theme__upload-photo{ng-mouseover: 'themeHoverIn()', ng-mouseleave: 'themeHoverOut()'}
-        %button.lmo-btn-link-no-underline{ng-click: 'openUploadCoverForm()', ng-if: 'canUploadPhotos()'}
+  .group-theme__cover{ng-style:'coverStyle()'}
+    .group-theme__upload-photo{ng-mouseover: 'themeHoverIn()', ng-mouseleave: 'themeHoverOut()'}
+      %button.lmo-btn-link-no-underline{ng-click: 'openUploadCoverForm()', ng-if: 'canUploadPhotos()'}
+        %i.fa.fa-lg.fa-camera
+        %span.group-theme__upload-help-text{ng-show: 'themeHover'} new photo
+
+
+  %section.group-theme__header--compact{ng-if: 'compact', aria-label: "{{ 'thread_group.aria_label | translate'}}"}
+    .group-theme__logo--compact{aria-hidden: 'true'}
+      %a{lmo-href-for: 'group'}
+        %img{ng-src: "{{group.logoUrl()}}"}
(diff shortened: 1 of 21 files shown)


lineman/app/components/group_page/group_theme/group_theme.haml
```diff
@@ -1,42 +1,43 @@
...
-  .group-theme__cover{ng-style:'coverStyle()'}
-    .group-theme__upload-photo{ng-mouseover: 'themeHoverIn()', ng-mouseleave: 'themeHoverOut()'}
-      %button.lmo-btn-link-no-underline{ng-click: 'openUploadCoverForm()', ng-if: 'canUploadPhotos()'}
-        %i.fa.fa-lg.fa-camera
-        %span.group-theme__upload-help-text{ng-show: 'themeHover'} new photo
+.group-theme
+  .lmo-row
+    .group-theme__cover{ng-style:'coverStyle()'}
+      .group-theme__upload-photo{ng-mouseover: 'themeHoverIn()', ng-mouseleave: 'themeHoverOut()'}
+        %button.lmo-btn-link-no-underline{ng-click: 'openUploadCoverForm()', ng-if: 'canUploadPhotos()'}
```
robguthrie: FYI - I'm noticing that using utility classes is ruining the nice haml. I really want us to move to using a single BEM class if possible and just mixing in the utility at the css level. 

I think it starts by making all utility classes mixins with the same name. So the haml is:
```
%button.group-theme__upload-button{...}
```
and the css is

```
.group-theme__upload-button {
   @include btnLinkNoUnderline
}
```


Does that feel ok to you? I really think we need to steer away from the utility classes now before people copy them any more. 

2 questions I have are.. 
- should we stop using all the bootstrap utilities (even the .btn- ones?).. I think so.. and we make mixins with what we want for our own buttons
- Should we namespace the mixin names like we've namespaced the utility classes.. and I think probably yes to that too.

I don't want to ban the utility classes but just only use them when it really makes no sense to mixin them and keep the BEM naming strong.

rdbartlett: Yeah sounds good. I will do some exploration now and update the PR.

rdbartlett:
@robguthrie check it out. I think you're right, feels like a better separation of concerns doing it this way. Want me to keep going and rip out a bunch of other utility classes?

robguthrie:
Yea nice.. I wasn't so much thinking of the layout classes when I wrote it.. but agree it's an improvement and glad you're onboard. I think my pet peeve is the button classes.

Yes making stuff conform is worth it if you do it without breaking stuff, and feel good while doing it. So please go ahead if you're jazzed.

I reckon the current split between utilities and mixins.css is broken - it would be much nicer to have the utility class defined right after the mixin is defined. Meaning we need to breakup and rename the files around the purpose of the styles ie: components/styles/layout.scss, buttons.scss etc

robguthrie:
you can move anything within components around and it'll all be included just fine

rdbartlett:
I have just restarted Travis in case you care about that. 

I think probably merge this as is. I went exploring but my brain is too mushy to commit anything further today I think.

PR #2634, opened 2015-10-23
decoupling from bootstrap buttons

rdbartlett:


rdbartlett:
Ok I have merged in the A11y-october branch because we are both working in the same area. I think this is good to go now. Good to smoke it first though, there's many changes.

lineman/app/components/memberships_page/memberships_panel/memberships_panel.haml
```diff
@@ -11,6 +11,6 @@
             %div @{{::membership.user().username}}
   .row.memberships-page__table-row
     .members-card__invite-members.col-xs-6{ng-if: 'membershipsPage.canAddMembers()'}
-      %a.btn.btn-success.btn-block.lmo-btn-icon{href: '', ng_click: 'membershipsPage.invitePeople()'}
+      %a.lmo-btn--featured.lmo-btn--block.lmo-btn--icon{href: '', ng_click: 'membershipsPage.invitePeople()'}
```
rdbartlett: this should be a button

PR #2635, opened 2015-10-23
link membership request notifications to memberhship requests page

robguthrie:


lineman/app/services/lmo_url_service.coffee
```diff
@@ -12,6 +12,9 @@ angular.module('loomioApp').factory 'LmoUrlService', (AppConfig) ->
     routePath: (route) ->
       "/".concat(route).replace('//', '/')
 
+    membershipRequest: (mr, params = {}, options = {}) ->
+      @buildModelRoute('g', mr.group().key, 'membership_requests', params, options)
```
robguthrie: @gdpelican - checking if correct use of this method.. it works but is it right to put the path there?

gdpelican: Hm, I'm surprised it's not stubbing 'membership_request' to 'membership-request'

We could use the 'route' method here, which is what we use for lmo-href-for

```
membershipRequest: (mr, params = {}, options = {}) ->
  @route model: mr.group(), action: 'membership_requests', params: params
```

robguthrie: oh cool. I really like the way that reads. thank you!

PR #2639, opened 2015-10-24
Refactor MessageChannelService; add testing

gdpelican:


.travis.yml
```diff
@@ -40,7 +40,11 @@ before_script:
 
 script:
   - bundle exec rake travis
-  - RAILS_ENV=test bundle exec rails server > /dev/null &
+  #- RAILS_ENV=test FAYE_URL="http://localhost:9292/faye" PRIVATE_PUB_SECRET_TOKEN=abc123 bundle exec rails server > /dev/null &
```
gdpelican: (NB: I've merged in the faye channel changes to this branch.)

app/models/event.rb
```diff
@@ -48,6 +64,13 @@ def notify_webhooks!
 
   private
 
+  def channel_object
+    if    GROUP_KINDS.include?      self.kind then self.eventable.group
```
gdpelican: I think the MessageChannelService was a little bit too knowledgeable about what objects were getting passed in. This means the MCS has no idea what an event is, and only takes a model and some data to publish to that model's channel.

This switch statement is an artifact of the slightly wonky event system, but should work for now.

app/models/notification.rb
```diff
@@ -12,7 +12,7 @@ class Notification < ActiveRecord::Base
   scope :unviewed, -> { where(viewed: false) }
 
   def publish_message
-    MessageChannelService.publish(self.user, NotificationSerializer.new(self).as_json)
+    MessageChannelService.publish(NotificationSerializer.new(self).as_json, to: self.user)
```
gdpelican: I kinda fancy the `publish data to: model` format; I think it reads nicely.

gdpelican:
Looks like the initial implementation of running faye on travis is no good...

robguthrie:
yea.. it was a test push.. looks like it works now though.

robguthrie:
James.. under this model can you imagine how we'd create a channel for all users? (say for a message to upgrade/refresh the client when we release a fix?)

gdpelican:
Strikes me as a singleton class which gets passed in:

```
class GlobalMessage
  include Singleton
end
```
```
MessageChannelService.publish({version: '0.15.0}, to: GlobalMessage.instance)
```
```
#message_channel_service.rb
def can_subscribe?(model)
  case model
  ...
  when GlobalMessage then true
  end
end

def channel_for(model)
  case model
  ...
  when GlobalMessage then '/global'
  end
end
```
```
#message_channel_controller.rb
def subscribe_user
  ...
  @subscriptions << MessageChannelService.subscribe_to(user: current_user, GlobalMessage.instance)
end
```

Although I'm having a ponder on whether the classes themselves should be responsible for telling the message channel service what channel they belong to, maybe with something like

```
class Discussion / Group / User
  include Publishable
end
```
```
module Publishable
  def channel_name
    "#{self.class.to_s}-#{self.key}"
  end
end
```
```
class GlobalMessage
  include Singleton
  def channel_name
    "/global"
  end
end
```
Which would get rid of the `channel_for` method and its big ugly switch statement:
```
#message_channel_service.rb
def self.publish(data, to:)
  raise UnknownChannelError.new unless to.respond_to?(:channel_name)
  PrivatePub.publish_to(to.channel_name, data)
end
```

robguthrie:
Sorry for slow reply. I think that GlobalChannel solution looks great. Just gonna re-mention here that we agreed discussion channels were a dumb idea.

PR #2642, opened 2015-10-27
Fix dashboard race condition

robguthrie:
Fix problem with dashboard loading and saying "you have no recent threads" because it ran before memberships were loaded.

Problem would have started being noticed when we added "only_threads_in_my_groups"


robguthrie:
branched from wrong place. closing

PR #2645, opened 2015-10-27
Fix at mention inconsistency on angular

gdpelican:
Hi, this is a much saner way to detect and render @mention links.

### DIFF
--- a/app/services/comment_service.rb
+++ b/app/services/comment_service.rb
@@ -49,12 +49,17 @@ class CommentService
   end
 
   def self.update(comment:, params:, actor:)
+    new_mentions = Comment.new(params).mentioned_usernames - comment.mentioned_usernames
+
     comment.edited_at = Time.zone.now
     comment.body = params[:body]
+
     return false unless comment.valid?
     actor.ability.authorize! :create, comment
     ThreadSearchService.index! comment.discussion_id
     comment.save!
+
+    User.where(username: new_mentions).each { |user| Events::UserMentioned.publish!(comment, user) } if new_mentions.any?
     Memos::CommentUpdated.publish!(comment)
     true
   end
--- a/lineman/app/app.coffee
+++ b/lineman/app/app.coffee
@@ -21,7 +21,7 @@ angular.module('loomioApp', ['ngNewRouter',
   renderer.link = (href, title, text) ->
     "<a href='" + href + "' title='" + (title || text) + "' target='_blank'>" + text + "</a>";
   renderer.paragraph = (text) ->
-    text = text.replace(/([\s(\[]+|^)@([a-zA-Z0-9]+)([\s)\]\,\?]|$)/g, "$1<a class='lmo-user-mention' href='/u/$2'>@$2</a>$3")
+    text = text.replace(/\[\[@([a-zA-Z0-9]+)\]\]/g, "<a class='lmo-user-mention' href='/u/$1'>@$1</a>")
     "<p>#{text}</p>"
 
   markedProvider.setOptions
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -93,9 +93,12 @@ class Comment < ActiveRecord::Base
     save!
   end
 
+  def mentioned_usernames
+    extract_mentioned_screen_names(self.body).uniq
+  end
+
   def mentioned_group_members
-    usernames = extract_mentioned_screen_names(self.body)
-    group.users.where(username: usernames).where('users.id != ?', author.id)
+    group.users.where(username: mentioned_usernames).where('users.id != ?', author.id)
   end
 
   def likes_count
--- a/app/serializers/comment_serializer.rb
+++ b/app/serializers/comment_serializer.rb
@@ -2,7 +2,7 @@ class CommentSerializer < ActiveModel::Serializer
   include Twitter::Autolink
 
   embed :ids, include: true
-  attributes :id, :body, :created_at, :updated_at, :parent_id, :parent_author_name
+  attributes :id, :body, :mentioned_usernames, :created_at, :updated_at, :parent_id, :parent_author_name
 
   has_one :author, serializer: UserSerializer, root: 'users'
   has_one :discussion, serializer: DiscussionSerializer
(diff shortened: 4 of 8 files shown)


robguthrie:
Not sure I'm convinced. Won't this break editing comments?

gdpelican:
Ah yeah, but I can go in and fix that so it parses it back correctly.

I'm way more interested in doing it this way than coming with some crazy cthulu regex to handle mentions, though... and this has the benefit of not @mentioning people who aren't actually people (like github, heh.)

robguthrie:
I'm really super dubious of modifying the body like this. I would really like to explore other options. Chat about this today too? 

I note that we also want to be able to support mentioning when editing comments. Which means we need to associate mentions with comments, maybe knowing that we can do something similar but without rewriting the body?

robguthrie:
So I'm thinking that if we create an atMention model then we can do something just like this but client side, and we can also detect when someone adds new mention when editing a comment, and ensure that behaviour is handled correctly. 

I do wonder if now is the time to plan out how to support mentions properly in contexts, proposals and votes and get this right.

@rdbartlett is now the time?

gdpelican:
I'm pretty sure that an at mention model is overkill; are you saying to store this info in the database? What info would you store? 

The functionality you're looking for can be accomplished really easily with the Twitter::Extractor as we use it now:

```
include Twitter::Extractor
#comment_service.rb
def self.update(comment:, params:, actor:)
  new_mentions = extract_mentioned_screen_names(params[:body]) - extract_mentioned_screen_names(comment.body)
  # update the comment
  mentions.each { |m| (publish a new mention event) }
end
```

And, when we want to generalize it for other models, we make a MentionService
```
include Twitter::Extractor
def new_mentions(current, updated)
  extract_mentioned_screen_names(updated) - extract_mentioned_screen_names(current)
end
```
```
new_mentions(comment.body, params[:body]).each { |mention| (publish mention event) }
```

lineman/app/app.coffee
```diff
@@ -21,7 +21,7 @@ angular.module('loomioApp', ['ngNewRouter',
   renderer.link = (href, title, text) ->
     "<a href='" + href + "' title='" + (title || text) + "' target='_blank'>" + text + "</a>";
   renderer.paragraph = (text) ->
-    text = text.replace(/([\s(\[]+|^)@([a-zA-Z0-9]+)([\s)\]\,\?]|$)/g, "$1<a class='lmo-user-mention' href='/u/$2'>@$2</a>$3")
+    text = text.replace(/\[@mention\|([a-zA-Z0-9]+)\]/g, "<a class='lmo-user-mention' href='/u/$1'>@$1</a>")
```
gdpelican: I want to call `text = MentionService.prepareForDisplay(text)`, but I can't inject the MentionService here. Grrn, silly problem to have.

lineman/app/services/mention_service.coffee
```diff
@@ -0,0 +1,7 @@
+angular.module('loomioApp').factory 'MentionService', ->
```
gdpelican: I think this is a nice solution for the display issues... whenever we're looking to use a body of text for display, we call `prepareForDisplay` on it, whenever we want to edit that text, we call `prepareForEdit`

app/services/comment_service.rb
```diff
@@ -50,12 +50,15 @@ def self.destroy(comment:, actor:)
 
   def self.update(comment:, params:, actor:)
     comment.edited_at = Time.zone.now
+    new_mentions = User.where username: Comment.new(params).mentioned_usernames - comment.mentioned_usernames
```
gdpelican: Think I prefer instantiating a new comment to having the CommentService know what a Twitter::Extractor is.

Also, maybe I should throw an empty check in there so we're not hitting the DB if there's not new mentions.

app/services/comment_service.rb
```diff
@@ -50,12 +50,15 @@ def self.destroy(comment:, actor:)
...
+
     comment.body = params[:body]
     return false unless comment.valid?
     actor.ability.authorize! :create, comment
     ThreadSearchService.index! comment.discussion_id
     comment.save!
     Memos::CommentUpdated.publish!(comment)
+    new_mentions.each { |user| Events::UserMentioned.publish!(comment, user) }
```
gdpelican: Also, this is lightweight and readable enough that I don't really mind duplicating it across a few services (VoteService, MotionService, DiscussionService)

But we could do

```
MentionService.publish_new_mentions(discussion, params[:description], discussion.description])
```
```
def publish_new_mentions(model, new_value, old_value = "")
  return unless mentions = extract_mentioned_screen_names(new_value) - extract_mentioned_screen_names(old_value)
  User.where(username: mentions).each { |user| Events::UserMentioned.publish!(model, user)
end
```

lineman/app/services/mention_service.coffee
robguthrie: I think I'm down with where you're going, particularly on not making a relation in the db. 

I'm happy to go with this, but still wish the server did not have to do a text transform, and that this transforming could be entirely client side. But.. I guess we can always change it later :)

gdpelican: I think the transforming is all client side at the moment?

server "I am a @user" ->
client display "I am a [@mention|user]" ->
client edit "I am a @user" ->
server "I am a @user"

If you give the branch a quick hoon I think it will be clear what's going
on.

robguthrie: Yep, I think I know what's going on. I was talking about the serializer modifying the body before the client gets it.

If we serialized the mentioned usernames into a comment property then this might be able to be done client side without modifying the text for edit.. IE: you could just have renderMentions and edit would not need anything.

Aside:
Will this work ok with the change detection, "are you sure you want to cancel?" stuff?

gdpelican: Yep, because we're only preparing the clone for editing, no touching actual
records.

gdpelican:
Okay, this should be good to go @robguthrie; let me know your thoughts and I'll squash it up if it's good.

robguthrie:
oh wow. Sooo good!!!

PR #2647, opened 2015-10-27
Search improvements

gdpelican:
These are minor improvements, but have made me excited about building the search results page, and putting in some advanced search functionality as we find it necessary.

I've tweaked the search weights so that discussion title is king, and comment bodies are very low valued (primarily so that they show up if there's nothing else, but they won't outweigh anything else if it exists)

### DIFF
--- a/app/models/model_cache.rb
+++ b/app/models/model_cache.rb
@@ -17,15 +17,15 @@ class ModelCache
   end
 
   def discussion_for(discussion_id)
-    discussions.select { |d| d.id == discussion_id.to_i }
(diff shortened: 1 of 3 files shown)


app/models/model_cache.rb
```diff
@@ -17,15 +17,15 @@ def comments
   end
 
   def discussion_for(discussion_id)
-    discussions.find_by id: discussion_id
```
gdpelican: This was N+1'ing because `find_by` runs a DB query, but we already have the records we want stored in memory

PR #2648, opened 2015-10-27
check for last read at on missed yesterday

robguthrie:


robguthrie:
@gdpelican  does this look ok to you?

app/views/user_mailer/missed_yesterday/_discussion.html.haml
```diff
@@ -13,7 +13,8 @@
     = render 'thread_mailer/motion', motion: motion
 
   .activity-feed
-    - discussion.items.where('created_at > ?', @time_start).order('created_at').each do |item|
+    - last_read_at = DiscussionReader.for(user: @user, discussion: discussion).last_read_at
```
gdpelican: I know we're in a delayed job here, but I'm a little worried about running that query for every discussion in every group in every email we send out.

Maybe we could use DiscussionReaderCache here somehow?

robguthrie:
Thanks for the great feedback @gdpelican . Implemented

gdpelican:
Yeah, that looks great.

PR #2649, opened 2015-10-27
Big performance improvement for discussion readers on dashboard

gdpelican:
Don't merge me yet, I have a couple more improvements I'd like to tackle first.


lineman/app/models/discussion_model.coffee
```diff
@@ -9,6 +9,7 @@ angular.module('loomioApp').factory 'DiscussionModel', (BaseModel, AppConfig) ->
     afterConstruction: ->
       if @isNew()
         @private = @privateDefaultValue()
+      @volume = @discussionReaderVolume or @recordStore.memberships.find({userId: AppConfig.currentUserId, groupId: @groupId })[0].volume
```
robguthrie: At first look I think this should be a function.. but now I understand you're doing it so we can query on this. Seems like the right thing to do, but if membership volume changes, what do we do?

app/models/discussion_reader.rb
```diff
@@ -42,6 +42,11 @@ def volume
     end
   end
 
+  def discussion_reader_volume
+    # Crazy James says: necessary in order to get a string back from the volume enum, rather than an integer
+    self.class.volumes.invert[self[:volume]]
```
robguthrie: with hindsight like this.. would you say that enum is worth it?

lineman/app/models/discussion_model.coffee
gdpelican: I looked at making it a function, but found it to be a little hairy because
in places we're doing things like `thread.volume = 'quiet'`, and I'd rather
deal with this problem than that problem.

That being said, not sure how to handle an updated membership just yet.

app/models/discussion_reader.rb
gdpelican: Yeah... the only reason we need this is because we've already overloaded the `volume` method, which would return the thing we want. I had a thought about doing this:

```
def volume(reader_only: false)
  if reader_only
    super
  elsif persisted?
    # etc
   end
end
```
```
def discussion_reader_volume
  volume(reader_only: true)
end
```

...But decided against it because I'd just rather not touch existing code at all.

app/extras/queries/visible_motions.rb
```diff
@@ -7,7 +7,7 @@ def initialize(user: nil, groups: nil, group_ids: nil)
       group_ids = Array(groups).map(&:id)
     end
 
-    @relation = Motion.joins(:discussion => :group).merge(Group.published).preload(:discussion)
+    @relation = Motion.joins(discussion: :group).where('groups.archived_at IS NOT NULL').preload(:discussion)
```
gdpelican: Does this seem okay to you, @robguthrie?

We were getting errors around `default_group_cover was not found on Motion` because we were merged a group relation and a motion relation.

gdpelican:
Okay, one failing spec left, and it's, of course, updating the discussion volume correctly.

I'm gonna table this for the moment and have a think about it later.

app/extras/queries/visible_motions.rb
```diff
@@ -7,7 +7,7 @@ def initialize(user: nil, groups: nil, group_ids: nil)
       group_ids = Array(groups).map(&:id)
     end
 
-    @relation = Motion.joins(:discussion => :group).merge(Group.published).preload(:discussion)
+    @relation = Motion.joins(discussion: :group).where('groups.archived_at IS NULL').preload(:discussion)
```
gdpelican: Does this seem ok @robguthrie ? We were getting errors with the default include because we were merging Motion relations with Group relations.

robguthrie: That's a little bit sad but I guess you gotta do it and move on. 

Seems weird though, running those lines in the console makes it look easy. However the rest of the query does get a bit scary sometimes.

gdpelican: I wouldn't put it past there being a bug in ActiveRecord, since it's pretty unexpected behaviour, but don't have the desire to track it down.

gdpelican:
Turned into a bit of a lot (dashboard tends to do that, eh?) but I'm happy with it at the moment. Could definitely use a smoke test, though.

Potentially risky areas:

- Muting / unmuting threads from the dashboard
- Changing the volume on groups and threads (both from the card and the dropdown)
- Removing the last admin of a group (not being able to, that is)
- Displaying a default group cover

I've tested these manually, and the tests pass, but would be really great to get another set of eyes or two on this before it goes.

robguthrie:
I'll smoke test this this morning with Hannah.

robguthrie:
spinning this up now.. first thing I notice is some conflicts in migrations. I'll test the migrations on a production copy and let you know how they go

robguthrie:
First problem I've noticed is that muting a group does not cause the threads to show as muted. IE: They still appear in recent threads. ( We should put an e2e on this)

robguthrie:
Changing thread volume √
Leaving group when no other coordinator prompts add coordinator √
Can leave group when another coordinator is present √
Deactivate group works √
Threads from muted groups show in recent threads X

Still need to test group cover photo on loomio-test

gdpelican:
I can't repro that muted group thingy; I added an e2e for it though.

Also, is there a reason we're displaying muted groups on inbox?

gdpelican:
OH. Hm. Okay, so threads which have a volume populated in the DB override the muted group setting. How do we want that to work? Ie,

- Thread set to quiet, group set to mute = thread appears on dashboard because it's quiet

robguthrie:
Thanks James. If thread has a volume setting it is preferred over group volume. No matter what the value. So group mute does not override thread mute. 

I hope this  will be way less confusing when we develop "Use this setting for all threads in the group" option for volume.

PR #2650, opened 2015-10-27
Decoupling from bootstrap buttons (squash)

gdpelican:
Squash version of #2649


gdpelican:
Hrm... these line addition / subtraction numbers aren't the same as the original branch... I reckon maybe just squash the old branch up, since we know that one works, unless someone wants to smoke this.

@robguthrie

robguthrie:
thanks for trying

PR #2653, opened 2015-10-28
Add group to start thread modal launched from group page

HSalmon:
[Trello card](https://trello.com/c/xbJCPPNB/846-when-i-open-the-start-thread-modal-from-the-group-page-i-want-to-see-the-group-name-so-i-am-confident-it-is-going-to-start-in-th)


gdpelican:
Smoked this and looks good to me! :cloud:

PR #2654, opened 2015-10-28
Add USD

mhjb:
Before
=====
image
image

image

After
===
image
image

image

### DIFF
--- a/app/assets/stylesheets/pricing_table.scss
+++ b/app/assets/stylesheets/pricing_table.scss
@@ -50,10 +50,11 @@
   border: 1px solid #b16a1a;
 }
 .pricing-table__price {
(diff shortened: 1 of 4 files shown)


robguthrie:
@mhjb is this ready to go?

mhjb:
Yes I think so. But @rdbartlett as product owner should have a glance over it.

mhjb:
Oops — I squashed, according to the tech manual instructions — but now there are 20-odd commits instead of one

robguthrie:
let's run through the squash page when we meet today and I'll try to make
it clearer

mhjb:
bonza

rdbartlett:
:+1:

PR #2656, opened 2015-10-28
improve notfications indexing

robguthrie:


robguthrie:
Before: 
image

gdpelican:
Yeah, good stuff.

The notifications stuff is a nest performance wise because there's so many events and eventables with relations that are tough to keep track of; there's a bunch of unneeded queries going on there I think.

PR #2657, opened 2015-10-28
remove motions_count method as it is a column anyway

robguthrie:
Looks like this method should not be on group, as I added it as a counter cache.

This means we'll use the CC rather than causing a query:
image


gdpelican:
This fix is in the dashboard improvements branch already, I think :)

robguthrie:
@gdpelican any objections to removing this discussions_controller_spec? It's too much work to fix because it's written with a mock_model. We have cukes covering the important functionality, we're not changing the file under test in the future.

gdpelican:
I fixed those specs in the other branch (with a miserable, miserable hack...), when I merge this in there I'd just as soon keep em in there if they pass easily, if not I don't mind dumping it.

PR #2659, opened 2015-10-29
Ensure group welcome modal works for logged out group creation

HSalmon:


HSalmon:
@robguthrie, is this ready to go d'you reckon?

app/models/invitation.rb
```diff
@@ -13,6 +13,8 @@ class InvitationAlreadyUsed < StandardError
   belongs_to :invitable, polymorphic: true
   belongs_to :canceller, class_name: User
 
+  update_counter_cache :invitable, :invitations_count
```
gdpelican: Hey @robguthrie do you have a strong sense of when this `update_counter_cache` fires? Is it after saves? Only saves which change the relevant information? Only after create / delete, and not update?

Also, should this be `update_counter_cache :group, :invitations_count`?

lineman/app/components/group_page/group_page_controller.coffee
```diff
@@ -37,4 +35,15 @@ angular.module('loomioApp').controller 'GroupPageController', ($rootScope, $loca
   @openUploadLogoForm = ->
     ModalService.open LogoPhotoForm, group: => @group
 
+  @handleSubscriptionSuccess = ->
```
gdpelican: This is great for now, but this 'do stuff when query parameters are there' feels like a separate concern to me, and I like the idea of separating it out. Haven't got a great idea for how yet tho.

gdpelican:
This looks wonderful!

app/models/invitation.rb
robguthrie: Yea, I am also keeping an eye out for if these are good value.

It adds hooks for create, update and destroy. So there is a count query per invitation operation. It's a bigger cost that what you see with an increment/decrement approach but it's less problematic.

lineman/app/components/group_page/group_page_controller.coffee
robguthrie: I imagine that if we had a router we really used then this would be state triggered and would feel less clunky.

I think I'm keen to explore using ui-router

PR #2661, opened 2015-10-29
Fix javascript error on profile page

gdpelican:


gdpelican:
To repro:

- Go to user profile page
- Make changes and hit 'save'

PR #2662, opened 2015-10-29
Search improvements

gdpelican:
Just throwing this up to see which tests I need to write / rewrite; this works hella well already though.


gdpelican:
I've greatly simplified search, yay! I'm going to annotate this PR so it's a little easier to understand.

app/models/discussion.rb
```diff
@@ -65,6 +65,17 @@ class Discussion < ActiveRecord::Base
   pg_search_scope :search, against: [:title, :description],
     using: {tsearch: {dictionary: "english"}}
 
+
+  scope :search_for, ->(query, user, opts = {}) do
```
gdpelican: So, searching for discussions has been reduced to 3 scopes, one on Discussion and two on SearchVector

- SearchVector#search_without_privacy returns all of the matches across the system, and returns their discussion_id and search rank
- SearchVector#search_for applies the Queries::VisibleDiscussions privacy option
- Discussion#search_for takes all of the discussion ids in SearchVector#search_for, and selects the fields we're interested in (title, description, key, a blurb, etc.) based on them.

Note that these queries are all knit together in ActiveRecord, so it all happens in one query that's pretty darn fast.

app/models/search_vector.rb
```diff
@@ -1,8 +1,76 @@
 class SearchVector < ActiveRecord::Base
+
+  WEIGHT_VALUES = [
```
gdpelican: These are the weights applied to different sections of the discussion. For example, a discussion title is ranked A (1.0), while a motion title is ranked B (0.3), a comment body is ranked D (0.03), etc. They're pulled out here so they can be easily tweaked (maybe even as an ENV?)

app/models/search_vector.rb
```diff
@@ -1,8 +1,76 @@
...
+    handle_asynchronously :index!
+
+  end
+
+  def self.index_everything!
+    index_without_delay! Discussion.pluck(:id)
+  end
+
+  def self.index_thread_sql(discussion_id)
+    "INSERT INTO discussion_search_vectors (discussion_id, search_vector)
```
gdpelican: This hasn't changed from the previous implementation (although I might be tempted to simplify it some, it works just fine.)

We're taking each of the different parts of the discussion and weighting them based on our weights above, creating a multi-weighted vector which gets stored in SearchVector for searching later.

Note the string_agg functions, which knit together all of the, say, motion names, into a single blob of text for vectorizing.

app/models/search_vector.rb
```diff
@@ -1,8 +1,76 @@
...
+       WHERE  discussion_id = #{discussion_id}) comments ON #{discussion_id} = id
+     WHERE    id = #{discussion_id}"
   end
+
+  def self.discussion_field_weights
+    DISCUSSION_FIELD_WEIGHTS.map { |field, weight| "setweight(to_tsvector(coalesce(#{field}, '')), '#{weight}')" }.join ' || '
+  end
+
+  scope :search_for, ->(query, user, opts = {}) do
+    Queries::VisibleDiscussions.apply_privacy_sql(
```
gdpelican: The great thing about having these as scopes is that we can really easily apply weirdo filters to them. (Although we have to do it at the SearchVector level, not the Discussion level)

So, something like
```
SearchVector.search_for('bunnies', current_user).where(group: )
```

will search in a single group. This works because we're joining the discussion and group fields on the search vector query, so we have access to loads of info about the things we're searching.
(We might have to do `where('groups.id = ?', group.id)`, but who cares that still rocks)

This allows for things like per-group queries, timeframing, and searching a single field
(Search Discussion Titles Only = `.where('discussions.description ilike %#{query}%')`

app/queries/thread_search_query.rb
```diff
@@ -1,147 +0,0 @@
...
-              #{rank_sql(motion_field_weights)} as rank,
-              #{field_as_blurb_sql('description')} as blurb
-     FROM     motions
-     WHERE    motions.discussion_id IN (#{top_ids.join(',')})
-     AND      #{field_weights_sql(motion_field_weights)} @@ plainto_tsquery(:query)
-     ORDER BY discussion_id, rank DESC"
-  end
-
-  def self.relevant_comments_sql(top_ids)
-    "SELECT   DISTINCT ON (discussion_id)
```
gdpelican: I got rid of the relevant comments / relevant motions stuff for now. It can go back in if it needs to, but it doesn't need to at the moment; I'd rather opt for getting it blazing fast and then make speed sacrifices for quality then if we need to.

app/serializers/search_result_serializer.rb
```diff
@@ -1,10 +1,6 @@
 class SearchResultSerializer < ActiveModel::Serializer
   embed :id, include: true
-  attributes :id, :priority, :query, :discussion_blurb, :motion_blurb, :comment_blurb
-
-  has_one :discussion
-  has_one :motion, root: :proposals
```
gdpelican: This serialization was killing us as well, especially given that the discussion serializer had some perf issues.

Now we just get a slightly-dumbed-down discussion back as the result, with no bells or whistles.

robguthrie: Looks really great James.  Thanks for explaining it all too.

PR #2665, opened 2015-10-30
make /development an index of development actions you can use

robguthrie:


gdpelican:
Nice.

Hm, I wonder if there'd be a way to use the existing rails Routing Error page to display this, but no point spending time on it.

PR #2667, opened 2015-10-30
Add slack webhook for new discussion

juliagra:


juliagra:
Hi @gdpelican, it took awhile to get my dev env set up again with the new angular frontend, but I finally manged to test this slack notification feature for a new discussion. 

screen shot 2015-10-29 at 6 09 37 pm

gdpelican:
This looks pretty great.

Only tweak I'd suggest is that we don't need the discussion title twice, so something like

Patrick Swayze started a new discussion in Fresh Group
  Lunch Options
  This will be a place...

would do just fine.

juliagra:
I updated the text to remove a double mention of the discussion title, now it looks like this: 

screen shot 2015-11-02 at 11 37 01 pm

gdpelican:
Looks great @juliagra, thanks!

PR #2669, opened 2015-10-30
Mention pass two: lists & tables

gdpelican:


gdpelican:
It kinda blows my mind that there's not a global rendering option.

robguthrie:
thanks james!

PR #2670, opened 2015-11-02
tag cohort_id after create

robguthrie:


robguthrie:
this is pretty embarassing. We're going to do a fix on cohorts/analytics soon. Sorry to everyone who loves loomio

PR #2671, opened 2015-11-02
add default cover photo to public

robguthrie:


robguthrie:
we already have a copy of this file in assets. I'm not going to use this.

PR #2673, opened 2015-11-02
Blue mentions

robguthrie:


lineman/app/components/user_page/user_page.haml
```diff
@@ -5,7 +5,7 @@
       .user-page__left
         %user_avatar{user: 'userPage.user', size: 'featured'}
       .user-page__right
-        %h2.user-page__name {{userPage.user.name}}
+        %h1.user-page__name {{userPage.user.name}}
         %h4.user-page__username @{{userPage.user.username}}
```
rdbartlett: why are we using an h4 here?

PR #2676, opened 2015-11-02
set is_referral for second and more groups

robguthrie:


rdbartlett:
Do you think there's any value in running a migration to make sure all groups match this definition?

robguthrie:
yes

PR #2683, opened 2015-11-03
Soft drafts

gdpelican:
Here's a PoC for the drafts backend; I haven't touched anything on the frontend yet, but it'd be nice to see this pass the specs anyhow.

Basic idea, we've got a 'Draft' class, which belongs_to a user and polymorphs to a 'draftable', which can be any model (currently, Group, Discussion, Motion). It stores a 'payload' json object, which stores the draft values.

The idea is that we call `$scope.comment.updateFromJSON(draft['comment'])` on the frontend; I don't want to get into the business of storing full-on models and marking them as 'draft' in the backend.

The specs in `drafts_controller_spec` describe the rest of the behaviour pretty well, I think. We're in a weird case where the update and create actions should do the same thing on the controller, so I've currently only got a `create` (read: upsert) action on the controller.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -70,7 +70,7 @@ class Ability
       (user_is_admin_of?(group.id) && group.enabled_beta_features.include?('export'))
     end
 
-    can [:members_autocomplete, :set_volume, :see_members, :move_discussions_to, :view_previous_proposals], Group do |group|
+    can [:members_autocomplete, :set_volume, :make_draft, :see_members, :move_discussions_to, :view_previous_proposals], Group do |group|
       user_is_member_of?(group.id)
     end
 
@@ -134,7 +134,7 @@ class Ability
       not user_to_deactivate.adminable_groups.published.any? { |g| g.admins.count == 1 }
     end
 
-    can [:update, :see_notifications_for], User do |user|
+    can [:update, :see_notifications_for, :make_draft], User do |user|
       @user == user
     end
 
@@ -212,7 +212,7 @@ class Ability
       user_is_member_of?(comment.group.id)
     end
 
-    can :add_comment, Discussion do |discussion|
+    can [:add_comment, :make_draft], Discussion do |discussion|
       user_is_member_of?(discussion.group_id)
     end
 
@@ -232,7 +232,7 @@ class Ability
         user_is_admin_of?(discussion.group_id) )
     end
 
-    can [:vote], Motion do |motion|
+    can [:vote, :make_draft], Motion do |motion|
       discussion = motion.discussion
       motion.voting? &&
       ((discussion.group.members_can_vote? && user_is_member_of?(discussion.group_id)) ||
@@ -275,6 +275,11 @@ class Ability
 
     can [:show], Vote do |vote|
       can?(:show, vote.motion)
+    end
+
+    can :update, Draft do |draft|
+      draft.user_id == @user.id &&
+      can?(:make_draft, draft.draftable)
     end
 
   end
--- a/lineman/app/models/draftable_model.coffee
+++ b/lineman/app/models/draftable_model.coffee
@@ -1,0 +1,18 @@
+angular.module('loomioApp').factory 'DraftableModel', (BaseModel) ->
+  class DraftableModel extends BaseModel
+    @draftParent: 'undefined'
+
+    draft: ->
+      @recordStore.drafts.findOrBuildFor(@[@constructor.draftParent]())
+
+    fetchDraft: =>
+      @recordStore.drafts.fetchFor(@[@constructor.draftParent]())
+
+    restoreDraft: ->
+      @update @draft().payload[@constructor.singular]
+
+    resetDraft: ->
+      @draft().updateFrom(@recordStore[@constructor.plural].build())
+
+    updateDraft: ->
+      @draft().updateFrom(@)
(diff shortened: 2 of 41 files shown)


rdbartlett:
shit you're good at programming

gdpelican:
Ugh. Sometimes I go back and read things that I write on the internet and all I can think is "neeeerd!" :neckbeard:

rdbartlett:
That's what I meant...

app/models/ability.rb
```diff
@@ -70,7 +70,7 @@ def initialize(user)
       (user_is_admin_of?(group.id) && group.enabled_beta_features.include?('export'))
     end
 
-    can [:members_autocomplete, :set_volume, :see_members, :move_discussions_to, :view_previous_proposals], Group do |group|
+    can [:members_autocomplete, :set_volume, :add_discussion, :see_members, :move_discussions_to, :view_previous_proposals], Group do |group|
```
robguthrie: these add_ methods are different to :create, discussion because you don't pass the discussion object eh?

gdpelican: Yeah, the ability is 'given a group, can I make a discussion in it'?
Alternative might be do do something like `can? :create,
group.discussions.build`

robguthrie:
Looks good james. It's how I imagined, fwiw

gdpelican:
Hooray! Good to hear we're thinking similarly. Will be interesting to see
how the frontend handles it

lineman/app/models/draftable_model.coffee
```diff
@@ -0,0 +1,15 @@
+angular.module('loomioApp').factory 'DraftableModel', (BaseModel) ->
+  class DraftableModel extends BaseModel
```
gdpelican: This is the first attempt at generalizing draftable objects.

idea:
- The model defines the name of the method to call for it's draft model (so, discussion calls group(), motion calls discussion(), etc.)
- Based on that, we update the payload of that object, with the model name, ie

```
# a discussion draft
payload: {
  motion: { ...update here for a new proposal draft... }
  comment: { ...update here for a new comment draft... }
}
```

```
# a group draft
payload: {
  discussion: { ...update here for a new thread... }
}
```

lineman/app/services/form_service.coffee
```diff
@@ -41,6 +41,7 @@ angular.module('loomioApp').factory 'FormService', ($rootScope, FlashService, $f
         scope.isDisabled = false
 
     submit: (scope, model, options = {}) ->
+      DraftService.applyDrafting(scope, model) if options.allowDrafts
```
gdpelican: This makes the interface for using drafts dead easy. If we're using the FormService, all we need to do is pass the `allowDrafts: true` option, and the DraftService will take care of the rest.

gdpelican:
Hm, I'm a little torn on this at the moment, because I'm finding this solution to be _really_ tight and elegant, but also a little difficult to explain. I wonder if you'd be willing to let me know your thoughts, and what parts don't make intuitive sense @robguthrie ?

gdpelican:
NB: I've only implemented things for new discussions at the moment, but if all goes to plan then allowing other forms to use the drafting system will be as easy as passing `allowDrafts: true` into the FormService.

gdpelican:
Also a thing that's missing at the moment is when the right time to load drafts up is. Do fire out a request to load up all drafts on pageload? Or load them as we hit the pages which have drafts?

One particularly possibly annoying case is the new thread form, where you can switch between all of your groups, which may or may not have drafts loaded for them yet.

gdpelican:
A pretty thorough testing of these has gone quite well. Things I tested:

- Typing into the thread form stores a draft (it comes back after you click 'cancel' and re-open the form)
- Typing into the group form stores a draft
- Typing into the subgroup form stores a draft, which does not interfere with any drafts you've created for other groups.
- Typing into the comment form stores a draft for that discussion.
- Drafts are not stored (and no endpoints are touched) for editing comments, groups, or threads.
- Threads, groups, and comment drafts store and restore line breaks and weird characters (like ßå@#!) correctly
- Threads, groups, and comments clear their draft data after the form has been submitted
- Switching between groups on the thread form loads up the correct draft for that group.

The weakest part of this at the moment is that if you a) have a draft stored in the DB for a group (ie, you've created or started to create a discussion for it before), switching to it from another draft will cause it to load the 'empty' draft, clearing out the existing text (this draft is still available by returning to the previously selected group)

The use case where it sucks is 'Type up a big long thing... oh wait crap this is supposed to be in 'Enspiral Services', not 'Enspiral'... change group to Enspiral, draft disappears, perhaps panic (although you can change group back to 'Enspiral' and the draft will come back... and then copy/paste is an option).

This is a small and non-destructive enough use case that I want to ignore it for now... if we get a couple complaints about it I'm happy to revisit.

robguthrie:
Thanks James, I'll test it with Hannah today. Im interested in how this
works if you have a couple of tabs open.

gdpelican:
Expected behaviour is that the last thing you typed is the draft that gets
saved, so typing 'draft A' in window 1, then 'draft B' in window 2, will
save 'draft B' as the draft. If someone goes back and types 'draft A2' into
window 1, then 'draft A2' will be saved as the draft.

HSalmon:
@robguthrie, @gdpelican, tested today. Works as expected :) @rdbartlett mentioned that it could be good for us to provide some kind of lightweight feedback that the draft has been saved, i.e. "draft saved X minutes ago”. Thoughts?

robguthrie:
Actually I'm really against telling the user anything. I don't think this
feature should make any noise or take any space.. it should be great,
completely silently.

robguthrie:
like github

rdbartlett:
The specific case I am concerned about is: user opens Thread modal, selects
group, starts typing, switches group, loses content. Actually the content
is still over in the other group but that is not obvious. I don't think
there is an equivalent scenario in Github.

gdpelican:
As mentioned above, I'd rather wait and get that complaint for users and identify what they're expecting, rather than speculate about it now.

rdbartlett:
I'm happy with that :)

PR #2686, opened 2015-11-04
tweaks

mhjb:
@rdbartlett pls review

Before | After
------------ | -------------
image | image
image | image
image | image
image | image

### DIFF
--- a/app/assets/stylesheets/pricing_table.scss
+++ b/app/assets/stylesheets/pricing_table.scss
@@ -72,5 +72,5 @@
 }
 
 .pricing-table__get-started {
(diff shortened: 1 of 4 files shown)


rdbartlett:
what's the last one?

rdbartlett:
Oh, right, the `button + button` margin. That was in there to make the spacing good on any modals that have 2 adjacent buttons. But we don't have any of them anymore eh?

mhjb:
I looked and couldn't find any

rdbartlett:
all good then

mhjb:
ready when you are @robguthrie

PR #2687, opened 2015-11-04
remove outline from user page h1

rdbartlett:
## before

image

## after 

image

### DIFF
--- a/lineman/app/css/1_utilities.scss
+++ b/lineman/app/css/1_utilities.scss
@@ -109,11 +109,15 @@ label.label-with-details{
   padding: 10px;
 }
 
-.lmo-h1{
+@mixin lmoH1{
   @include fontHelveticaBold;
   @include fontGrande;
   padding: 0;
   margin: 0;
+  outline: 0;
+}
+.lmo-h1{
+  @include lmoH1;
 }
 .lmo-h1-medium{
   @include fontHelveticaMedium;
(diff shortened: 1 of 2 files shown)


lineman/app/css/1_utilities.scss
```diff
@@ -109,11 +109,15 @@ label.label-with-details{
   padding: 10px;
 }
 
-.lmo-h1{
+@mixin lmoH1{
```
gdpelican: I want to flag that I'm not a fan of this BEM mixin strategy, because it makes us bury through another layer of abstraction before getting to the actual styles.

ie,

.user-page__name . o (WTH is that?) -> @mixin lmoH1. o (WTH is that?) -> actual styles

instead of

.lmo-h1 . o (WTH is that?) -> actual styles

rdbartlett: Fair enough. 

I just went through and applied this pattern to all the button classes and it felt like a great improvement. See what you think: #2634 

I like the abstraction because it reminds me, 'hey this collection of style declarations is reused in a whole bunch of places'. In many places it meant I could remove all the utility classes from the view and only be left with a single semantic class names on each element, and all the style information relegated to a css file where it should be. (I only like to use utility classes in the view when we don't already have a css file for that block.)

gdpelican:
Whoa, that build 0% checked itself, and 100% wrecked itself.

PR #2689, opened 2015-11-04
Fix for public API errors; logged out user API views

gdpelican:


gdpelican:
Here's the cheapy fix: use `accessible_records` instead of `visible_records` for controllers which are using a query object.

Later I can go back in and fiddle with a VisitorUser object and flesh out some object queries if we need em, but this should stop those errors, which is good.

I'll write an e2e for this before merge.

gdpelican:
Oops, looks like VisibleDiscussions isn't quite set up to handle the nil user case. Will circle back on this one.

app/controllers/api/discussions_controller.rb
```diff
@@ -46,13 +47,15 @@ def set_volume
   private
 
   def visible_records
```
gdpelican: This is the original intended usage of visible / public records, I think:

```
def visible_records
  current_user.groups
end

def public_records
  Group.visible_to_public
end
```

whereas in a controller where they want to be the same thing (like, ExploreController, maybe?), we just write `accessible_records`:

```
def accessible_records
  Discussion.visible_to_explore # whether you're logged in or not
end
```

Could pull this out of the gem, but overwriting `accessible_records` is just as good, and I reckon that otherwise we'd just have a lot of `visible_records` methods which look like

```
def visible_records
  current_user.is_logged_in? ? current_user.groups : Group.visible_to_public
end
```
anyway.

app/controllers/api/motions_controller.rb
```diff
@@ -25,8 +25,11 @@ def update_outcome
   end
 
   def index
-    load_and_authorize :discussion
-    instantiate_collection { |collection| collection.where(discussion: @discussion).order(:created_at) }
+    load_and_authorize :discussion if params[:discussion_id] || params[:discussion_key]
+    instantiate_collection do |collection|
+      collection = collection.where(discussion: @discussion) if @discussion
```
gdpelican: This'll be a common pattern, I reckon, in the index action... either 'give me all the discussions for this group' or 'all the groups for this user' or 'all the motions for this discussion' or whatever... I reckon pulling that behaviour out somehow might be nice.

app/controllers/popolo/motions_controller.rb
```diff
@@ -1,4 +1,4 @@
-class Popolo::MotionsController < API::RestfulController
+class Popolo::MotionsController < API::MotionsController
```
gdpelican: This is what makes me really excited about doing some API interoperability work: write a new controller, inherit from the Restful version of it, define a new serializer. Bam, new interoperability option, which works exactly like the one we have, in a different format.

app/extras/notification_items/membership_requested.rb
```diff
@@ -6,15 +6,7 @@ def initialize(notification)
...
-    requestor = notification.eventable.requestor
-    if requestor
-      requestor
-    else
-      visitor = LoggedOutUser.new(name: notification.eventable.name,
-                                 email: notification.eventable.email)
-      visitor.set_avatar_initials
-      visitor
-    end
+    notification.eventable.requester || LoggedOutUser.new
```
gdpelican: >.>

Gemfile
```diff
@@ -68,7 +68,7 @@ gem 'oj_mimic_json'
 # might need this
 #gem 'activeuuid', '>= 0.5.0'
 
-gem 'snorlax'
+gem 'snorlax', '~> 0.1.3'
```
gdpelican: This updates the gem to use

`if current_user.is_logged_in? then (thing) else (other thing)`

instead of

`if current_user then (thing) else (other thing)`

which naturally doesn't work with the null user pattern

app/controllers/api/discussions_controller.rb
```diff
@@ -46,13 +47,15 @@ def set_volume
...
-  def visible_groups
-    return current_user.groups unless params[:group_id] || params[:group_key]
-    load_and_authorize :group
-    [@group, @group.subgroups].flatten
+  def public_records
+    Queries::VisibleDiscussions.new(user: current_user, groups: organisation_groups || Group.visible_to_public)
+  end
+
+  def organisation_groups
+    [@group, @group.subgroups].flatten if @group
```
gdpelican: Here's another one where we're going 'if there's a group, give me just the ones in that group, otherwise give me everything for this user / everything public'

app/controllers/api/restful_controller.rb
```diff
@@ -17,6 +17,10 @@ def destroy_action
 
   private
 
+  def current_user
+    super || LoggedOutUser.new
```
gdpelican: I was too afraid to change `current_user_or_visitor` to `current_user` in the rest of the app, and didn't want to write `current_user_or_visitor` everywhere in the API controller, so I remade the `current_user_or_visitor` helper here, now called `current_user` (so we can guarantee that there's always an object there for the current user)

Happy to hear different thoughts on a good way to do this.

app/extras/queries/visible_discussions.rb
```diff
@@ -117,7 +116,7 @@ def self.apply_privacy_sql(user: nil, group_ids: [], relation: nil)
                     (discussions.group_id IN (:user_group_ids)) OR
                     (groups.parent_members_can_see_discussions = TRUE AND groups.parent_id IN (:user_group_ids)))',
                    group_ids: group_ids,
-                   user_group_ids: user_group_ids)
+                   user_group_ids: (user || LoggedOutUser.new).group_ids)
```
gdpelican: (Because beta calls this and I don't trust that it's always passing in a `current_user_or_visitor` value.)

app/models/discussion_reader.rb
```diff
@@ -51,10 +51,6 @@ def first_read?
     last_read_at.blank?
   end
 
-  def user_or_logged_out_user
-    user || LoggedOutUser.new
-  end
-
```
gdpelican: Can't quite fathom why this was here, and I don't see references to it anywhere else.

app/controllers/popolo/motions_controller.rb
robguthrie: yusss

app/controllers/api/discussions_controller.rb
robguthrie: Yaa.. I reckon this should be abstracted so the controller does not know about public/private and logged out cases.

Queries::VisibleDiscussions should hold def organisation_groups for instance.
Don't 'cha think?

gdpelican:
Hey @robguthrie what do you think about switching on angular for visitors? I'm going to have a play with a few public groups on this branch and see how it holds up.

robguthrie:
We need to have a think about search engine discoverability, when we're
happy about that, then we can turn it on.

gdpelican:
Should be

```
# logged_out_user.rb
def angular_ui_enabled?
  is_not_a_robot # matched from 
end
```
There's a gem, which looks okay (it's a Blade Runner reference): https://github.com/biola/Voight-Kampff
And a stackoverflow on the topic: http://stackoverflow.com/questions/5882264/ruby-on-rails-how-to-determine-if-a-request-was-made-by-a-robot-or-search-engin

Also, it'd be really nice to have even a back-of-napkin design for the logged out navbar... I was thinking we lose dashboard / inbox / groups, keep search (search works for logged out users, wee!), and replace the user menu with a link to log in / create an account.

robguthrie:
Hrrmmm.. Serving different content to google is likely to drop our score.
Google indexes JS apps now.. we just need to make sure that it can find all
the links within the thing.

app/controllers/api/memberships_controller.rb
```diff
@@ -54,13 +54,11 @@ def remove_admin
...
 
-  def visible_records
-    visible = resource_class.includes(:user, :group, :inviter)
-    if current_user.group_ids.any?
-      visible.where("group_id IN (#{current_user.group_ids.join(',')}) OR groups.is_visible_to_public = 't'")
-    else
-      visible.where("groups.is_visible_to_public = 't'")
-    end
+  def accessible_records
+    resource_class.joins(:group)
```
robguthrie: This is cool. Lets make a next step ticket for VisibleGroupsQuery

app/controllers/api/restful_controller.rb
robguthrie: I like the approach.. but how does the super call work in this case?

Where do we return the real current user when they're logged in?

(I think I'm finding that I don't ever like the use of super.. but that's totally an aside and I'm happy for you to use it once I understand this)

app/extras/queries/visible_discussions.rb
```diff
@@ -1,17 +1,14 @@
 class Queries::VisibleDiscussions < Delegator
-  def initialize(user: nil, groups: nil, group_ids: nil)
-    @user = user
-
-    if group_ids.nil? and groups.present?
-      group_ids = Array(groups).map(&:id)
-    end
+  def initialize(user:, groups: nil, group_ids: nil)
+    @user = user || LoggedOutUser.new
+    @group_ids = group_ids.presence || Array(groups).map(&:id).presence || user.group_ids
```
robguthrie: is presence the same as compact?

robguthrie: never mind I looked it up. Cool trick. thanks

db/schema.rb
```diff
@@ -16,21 +16,22 @@
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
   enable_extension "hstore"
+  enable_extension "pg_stat_statements"
 
   create_table "active_admin_comments", force: :cascade do |t|
-    t.string   "resource_id",   null: false
-    t.string   "resource_type", null: false
+    t.string   "resource_id",   limit: 255, null: false
```
robguthrie: It seems like we you and me have different configuations... mine seems to want to remove limits and yours is adding them.. I wonder what the difference is?

robguthrie:
This looks good. Feels ready to merge to me.

PR #2690, opened 2015-11-05
Slack webhooks new comment

BenjamH:
screen shot 2015-11-04 at 4 58 22 pm

Hi! Sorry been away for a while. I've gotten the slack notifications for `new_comment` to work.

I've been having trouble squashing my commits. Tried for a couple hours but to no avail. I've tried doing git rebase -i and commits are still showing up. I've also tried the branch-squash and the same thing is happening.


BenjamH:
There's also no sure-fire way for universal markdown compatibility between the loomio markdown and slack markdown, they are a little different.

gdpelican:
Hey @BenjamH: 

The simplest possible way I've found to squash:
(given a feature branch `feature`, and starting on branch `master`)
```
git checkout -b feature-squash # create a new squash branch and switch to it
git merge feature --squash # squash all commits into one on the new branch
git commit -m "Commit message" # make a commit of the squashed commits
git push origin feature-squash:feature -f # push the squashed commit to the feature branch on origin (-f forces the history overwrite.
```

spec/.ruby-version
```diff
@@ -0,0 +1 @@
+2.2.2
```
gdpelican: We want to make sure these stray version files don't appear in our commit

app/models/webhooks/slack/base.rb
```diff
@@ -20,13 +20,11 @@ def attachments
       text:        attachment_text,
       fields:      attachment_fields,
       fallback:    attachment_fallback,
-      color:       attachment_color
+      color:       attachment_color,
+      mrkdwn_in: [:text]
```
gdpelican: Not sure what this is intended to do?

app/models/webhooks/slack/base.rb
```diff
@@ -20,13 +20,11 @@ def attachments
...
     }]
   end
 
-  def attachment_color
```
gdpelican: I don't think we want to drop this from the base, otherwise we have to define it in each of the sub classes, which is repeating ourself, which is bad.

app/models/webhooks/slack/new_comment.rb
```diff
@@ -0,0 +1,25 @@
+class Webhooks::Slack::NewComment < Webhooks::Slack::Base
+
+  def text
+    I18n.t :"webhooks.slack.new_comment", author: author.name, name: discussion_link
```
gdpelican: I reckon we'd want to find a way to link to the exact comment which the message is referring to; the url would be something like

```
/d/<key>?comment=<commentId>
```
Which should scroll you to that comment automagically.

gdpelican:
Going to close this for now; let me know if you're interested in picking it back up, @BenjamH !

PR #2698, opened 2015-11-10
move feedback card to footer

rdbartlett:
This is an urgent fix to get the feedback card out of the way. We are getting lots of messages from people saying it is in the way. Soon this will be in the sidebar but I wanted to get this fix out in a hurry.

image


gdpelican:
This looks pretty not great and I'd be keen to at least put a card around it so that it's not text floating in space. Looks especially bad when the dashboard / inbox are loading.

rdbartlett:
I think it has always looked bad, this is just different bad. Looking
forward to a) getting a sidebar and b) figuring out what to do with the
'contact us'/'give feedback' overlap

gdpelican:
Okay, rephrasing: I think it looks much worse and is more annoying than before. Permission to fix?

rdbartlett:
hehe yes for sure. so long as the fix doesn't put it in front of key
interaction elements :)

PR #2699, opened 2015-11-10
Welcome thread language tweak

mhjb:
Provoked by user test revealing user wasn't yet clear that there was such a thing as the comment form.

Before | After
-------- | ----
image | image

@rdbartlett please review


rdbartlett:
Drastic improvement and reduction in flaccid Kiwish

PR #2701, opened 2015-11-11
pull out data for massey research

robguthrie:


app/admin/groups.rb
```diff
@@ -158,6 +158,13 @@ def collection
     f.actions
   end
 
+  collection_action :massey_data, method: :get do
+    rows = Group.visible_to_public.map do |group|
```
gdpelican: Think `pluck` is the way to go here; love that they changed it to take a list of columns.

```
Group.visible_to_public.pluck(:id, :parent_id, :name. :description)
```

robguthrie: oh nice!

PR #2702, opened 2015-11-11
Make subscription success language more appropriate

mhjb:
I can't figure out how to test this. I think it's right, but it does need testing.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -497,11 +497,11 @@ en:
 
   subscription_success_modal:
     heading: Thanks for setting up your subscription!
-    social_mission: 'Your contribution helps us to deliver on our social mission: making it easy for anyone, anywhere to participate in decisions that affect them.'
-    receipt: "If you ever want to change or cancel your subscription, or if you have any other queries,"
-    contact_link: ' please contact us.'
+    receipt: You’ll receive an email with a receipt.
+    change_subscription: "If you ever want to change or cancel your subscription, or if you have any other queries, "
+    contact_link: please contact us.
     sign_off: ❤ from the Loomio team
-    ok_got_it: 'Ok, got it!'
+    ok_got_it: Ok, got it!
 
   edit_group_form:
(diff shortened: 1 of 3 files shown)


mhjb:
Ready, I reckon. @rdbartlett 

Before | After
----- | -----
image | image
image | image

rdbartlett:
Nice

robguthrie:
What about mentioning the manage your subscription link as well as
contacting us?

mhjb:
I wondered about that — but I'm not sure enough thought has yet gone into what kinds of things customers will be able to do in their billing portal. Better to just talk to us at this stage imho.

robguthrie:
but but but! yea ok

lineman/app/components/group_page/group_page_controller.coffee
```diff
@@ -38,7 +38,7 @@ angular.module('loomioApp').controller 'GroupPageController', ($rootScope, $loca
     ModalService.open LogoPhotoForm, group: => @group
 
   @handleSubscriptionSuccess = ->
-    if AppConfig.chargify and $location.search().chargify_success?
+    if (AppConfig.chargify or AppConfig.environment != 'production') and $location.search().chargify_success?
```
gdpelican: I'd rather not change behaviour based on environment here.

robguthrie: Would you be ok with removing the chargify check then? It's impossible to
test this otherwise.

gdpelican: Not sure I understand? For testing in dev we could set `.env` to contain the chargify test creds, for automated testing we could stub out `AppConfig.chargify = true`?

robguthrie: I just wanted an easy way for mhjb to test his css / html changes. I guess
he can put chargify stuff into his env, but that seems like more hassle
than just making it easier to trigger.

PR #2703, opened 2015-11-11
Remove create account link from log in page

mhjb:
I recorded a [2.5 minute screencast](https://youtu.be/ga4rQq9TZSk) to look at what happens when you create an account from the log in page. It's an un-designed flow. It's not welcoming. It makes 'start thread' and 'invite people' non-sensical, but relatively prominent actions. It doesn't validate email addresses. It doesn't set the user's locale. You get no email trail that you set up an account. Also if you are invited to a group, and choose 'If you already have a Loomio account, log in here' it's a bit confusing on the next page seeing the option 'Don't have an account? Create one now'.

Before | After
--- | --- 
image | image
image | image

What chu reckon e.g. @rdbartlett @robguthrie ?

### DIFF
--- a/app/views/devise/sessions/new.html.haml
+++ b/app/views/devise/sessions/new.html.haml
@@ -16,10 +16,6 @@
               - else
                 = t :sign_in
 
-          %h2= t :"common.welcome_back"
-          %p
-            = t(:'dont_have_an_account')
-            = link_to t(:'create_one_now'), new_user_registration_path, id:'create_account'
           .signin_container
             = render "form"
             -unless omniauth_authenticated_and_waiting?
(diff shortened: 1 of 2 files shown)


rdbartlett:
How cool that you can get into the app without confirming email though!

I agree we should rip it out. 

I quite like the "welcome back" header though because it is an additional hint that you are on the wrong screen, if this is your first time on the log in page. I copied that off some other site once. Don't have a strong opinion

rdbartlett:
Also if you are removing unique strings from the UI make sure you remove them from the i18n too.

mhjb:
I kinda like welcome back too. Just looked overkillish having two headings.

gdpelican:
Er, how does someone join our website then? Are we forcing people to join with a third party account?

rdbartlett:
They join by starting a group or being invited to join one. Creating an
account without a group is undesigned and quite broken.

gdpelican:
Ah, okay. I use it quite a bit, but probably only because I'm devvin'.

PR #2709, opened 2015-11-13
Bring it closer

mhjb:
@rdbartlett


rdbartlett:
Very good

robguthrie:
thanks @mhjb !

PR #2713, opened 2015-11-16
Perf improvements

gdpelican:
Fix N+1 / unused eager loading queries as suggested by bullet

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -42,4 +42,10 @@ Loomio::Application.configure do
   }
 
   config.action_controller.action_on_unpermitted_parameters = :raise
+
+  config.after_initialize do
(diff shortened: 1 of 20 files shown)


config/environments/development.rb
```diff
@@ -42,4 +42,10 @@
   }
 
   config.action_controller.action_on_unpermitted_parameters = :raise
+
+  config.after_initialize do
```
robguthrie: Are you sure this will not be loaded in production?

gdpelican: It's in development.rb, so I don't think it's loaded in prod unless we've got some super weirdo rails setup.

robguthrie: ok cool. thanks

PR #2714, opened 2015-11-16
set application locale in api controller

robguthrie:
fixes bug where welcome message could be generated in wrong language


gdpelican:
Can we close this @robguthrie ?

robguthrie:
@gdpelican  I think that this should fix #2851 I'm just refreshing it now.

PR #2715, opened 2015-11-16
move to ruby 2.2.3 and update gemfile

robguthrie:


gdpelican:
Excite for ruby 2.3 soon!

robguthrie:
Personally I'm really excited to see if we can run upon Rubinius one day.
What is it about ruby 2.3 that excites you?

gdpelican:
No massive wins; I don't care about the safe navigation operator so much, but Hash/Array#dig is cool, and Hash#select_values is cool, and greb_v could be cool.

robguthrie:
Thanks. I agree with that!

PR #2718, opened 2015-11-17
Max image width for group description

gdpelican:


gdpelican:
Oh, I put a colon in there. stupid me. Anyway we'll just got with #2720, which is the same fix.

PR #2729, opened 2015-11-18
Add stories to marketing page

HSalmon:
[Trello card](https://trello.com/c/ON6rsMTN/954-5-display-stories-on-homepage)

screen shot 2015-11-18 at 3 08 01 pm

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,6 +65,7 @@ gem 'rack-attack', '~> 4.2.0'
 gem 'ahoy_matey'
 gem 'oj'
 gem 'oj_mimic_json'
+gem 'simple-rss'
 # might need this
 #gem 'activeuuid', '>= 0.5.0'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -596,6 +596,7 @@ GEM
     sexp_processor (4.5.1)
     shoulda-matchers (2.8.0)
       activesupport (>= 3.0.0)
+    simple-rss (1.3.1)
     simple_form (3.1.0)
       actionpack (~> 4.0)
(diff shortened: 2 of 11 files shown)


robguthrie:
@rdbartlett look at this. Tell us what needs to be improved!

robguthrie:
@mhjb do you have any designy feedback on this?

rdbartlett:
Awesome!! 

Images need to be cropped to be the same dimensions. Wordpress will serve you cropped images like this: https://loomio.files.wordpress.com/2015/10/17-tory-st.jpg?w=250&h=155&crop=1'

Or you can CSS, specify the size of the div, set the background to the src, and use `background-size: cover` to make it crop nicely.

robguthrie:
note to ourselves: add cron job to update stories hourly

rdbartlett:
Can we get this onto loomiotest plz?

robguthrie:
yes!

rdbartlett:
@robguthrie can you run the cron or whatever it is that we need to do to get loomiotest.org to display the stories?

rdbartlett:
Cool, it's working now. 

I'm not super happy with the resolution of the images. It's kinda awkward because we want to balance low file size with nice aesthetics, taking the responsive layout into consideration. Currently on iPad it looks stink:

image

As an easy halfway solution, can we just change thumb_url so it is loading images at say 500x310? That looks like this:

image

Also notice the aspect ratio (500x310) is the same as what's used on the blog.

rdbartlett:
(oops accidental close)

robguthrie:
yes we can do that rich

HSalmon:
image

PR #2733, opened 2015-11-18
Logged out angular

gdpelican:
Not ready for merging yet.


gdpelican:
Erb, there's a bunch of cukes that suggest there's quite a bit more work to be done here; things like subscribing to an rss feed, modifying email preferences via unsubscribe link, etc. I'll keep plugging away though; the logged out view looks pretty nice so far.

gdpelican:
I feel pretty good about this, but want to hold off until the e2ez are running properly, since there's some stuff in there that should get a proper pass.

features/users/login.feature
```diff
@@ -15,19 +15,3 @@ Feature: Login
...
-
-  Scenario: Redirected to original page on login
-    Given I am a logged out user
-    And I am a member of a hidden group
-    When I visit the group page
-    And I log in
-    Then I should see the group page
-
-  @javascript
-  Scenario: Redirected to previous public page on login
```
gdpelican: There's a couple of things here which are actually lack of feature parity; redirecting you to the right place after login and unsubscribing via email link being the two I remember off the top of my head.

lineman/app/components/navbar/navbar.coffee
```diff
@@ -11,6 +11,8 @@ angular.module('loomioApp').directive 'navbar', ->
 
     $scope.hostName = parser.hostname
 
+    $scope.isLoggedIn = ->
+      AbilityService.isLoggedIn()
```
gdpelican: I wondered if this should be `AppConfig.isLoggedIn()`, but it felt more like an ability of the user than a configuration thing.

lineman/app/components/navbar/navbar_search.coffee
```diff
@@ -21,6 +21,9 @@ angular.module('loomioApp').directive 'navbarSearch', ->
       return if $element[0].contains document.activeElement
       $scope.closeSearchDropdown()
 
+    $scope.showDropdown = ->
```
gdpelican: Only show the groups dropdown bit if the user is a member of a group, otherwise it shows a weird disembodied floaty thing when you first click on search.

lineman/app/components/thread_page/activity_card/activity_card.coffee
```diff
@@ -46,7 +46,7 @@ angular.module('loomioApp').directive 'activityCard', ->
 
     $scope.threadItemVisible = (item) ->
       addSequenceId(item.sequenceId)
-      $scope.discussion.markAsRead(item.sequenceId)
+      $scope.discussion.markAsRead(item.sequenceId) if AbilityService.isLoggedIn()
```
gdpelican: I think I actually want to base this on whether the discussion model has a reader or not, so the activity card doesn't care if you're logged in or not.

```
# discussion_model.coffee
markAsRead: (sequenceId) ->
  return unless @discussionReaderId
```

lineman/app/services/thread_query_service.coffee
```diff
@@ -43,7 +43,7 @@ angular.module('loomioApp').factory 'ThreadQueryService', (Records, CurrentUser)
     applyFilters = (view, filters, queryType) ->
       filters = [].concat filters
 
-      view.applyFind(discussionReaderId: { $gt: 0 })
+      view.applyFind(discussionReaderId: { $gt: 0 }) if AbilityService.isLoggedIn()
```
gdpelican: Otherwise the logged out user never sees a list of discussions (on the group page for example)

gdpelican:
I'm keen to merge this in, which will get rid of a lot of that error noise we're seeing in prod.
Any objections, @robguthrie ? (This doesn't actually render angular to logged out users anymore)

robguthrie:
Hrrmm The only concerning thing is removing the cukes around requesting membership.

I won't have much loomio time until Sunday now.

gdpelican:
Kay, I've put those cukes back in; going to merge and smoke this on clone.

PR #2736, opened 2015-11-18
run all e2es

robguthrie:


robguthrie:
@gdpelican do you have any idea why this last failure is happening? It seems to only happen on travis. Maybe it's related to memberships not being loaded?

gdpelican:
Hrm. I tried halting membership loading and running the test, and got a different error locally:

```
NoSuchElementError: No element found using locator: By.cssSelector(".proposal-outcome-panel__set-outcome-btn")
```
rather than
```
NoSuchElementError: No element found using locator: By.cssSelector(".proposal-form__outcome-field")
```

That error indicates the modal isn't loading correctly for some reason? ... I don't know, I don't think I have any idea.

PR #2738, opened 2015-11-19
invite to group

HSalmon:


rdbartlett:
How's it going? I see more `-` than `+`, which is always a good sign :)

gdpelican:
I think the 'WIP' in this case means we haven't put the +'s back in yet :smile:

gdpelican:
But yeah this will be much simpler and looks good so far.

lineman/app/components/invitation_form/add_members_modal/add_members_modal.coffee
```diff
@@ -0,0 +1,35 @@
...
+      _.filter group.parent().members(), (user)->
+        !user.isMemberOf(group)
+
+    $scope.canAddMembers = ->
+      $scope.members().length > 0
+
+    LoadingService.applyLoadingFunction($scope, 'load')
+    $scope.load()
+
+    $scope.submit = ->
```
gdpelican: Couldn't get FormService to fit here?

I believe there's a successFn option which is called after success which passes the data through... maybe the addUsersToSubgroup submitFn was tricky?

lineman/app/components/invitation_form/invitation_form.coffee
```diff
@@ -1,98 +1,18 @@
...
-  controller: ($scope, $rootScope, $q, group, Records, CurrentUser, AbilityService, LoadingService, FlashService) ->
+  controller: ($scope, group, CurrentUser, AbilityService, FlashService, RestfulClient, ModalService, TeamLinkModal, AddMembersModal) ->
     $scope.group = group
-    $scope.invitations = []
-    $scope.messageFieldHidden = true
+    $scope.form = { emailAddresses: '' }
+    $scope.showCustomMessageField = false
+    $scope.isDisabled = false
+    $scope.noInvitations = false
+    remote = new RestfulClient('invitations')
```
gdpelican: Shouldn't we have a records interface of some kind instead of manually creating a restful client in the controller?

lineman/app/components/invitation_form/add_members_modal/add_members_modal.coffee
robguthrie: I'm not sure how best to use FormService when you're not passing a single model to save. 

Reading over it (FormService) now, I wonder if we want to create a general FormService that can handle any kind of submit, and a more specific ModelFormService that does all the model specific stuff.

robguthrie:
Unless anyone objects I'll merge this in tomorrow morning.

app/controllers/api/memberships_controller.rb
```diff
@@ -1,5 +1,16 @@
 class API::MembershipsController < API::RestfulController
 
+  def add_to_subgroup
+    parent_group = current_user.groups.find_by_id!(params[:parent_group_id])
+    group = current_user.groups.find_by_id!(params[:group_id])
```
gdpelican: I believe `load_and_authorize(:group)` does the same thing here, and will respond with the appropriate 403 instead of 500'ing if the user isn't a member of the specified group.

app/controllers/api/memberships_controller.rb
```diff
@@ -1,5 +1,16 @@
...
+
+    users = parent_group.members.where('users.id IN (?)', params[:user_ids])
```
gdpelican: Could be `parent_group.members.where('users.id': params[:user_ids])` for slightly less SQL (this makes it so it doesn't error when params[:user_ids] is blank)

app/models/ability.rb
```diff
@@ -95,8 +102,9 @@ def initialize(user)
       # inwhich case we need to confirm membership and permission
 
       group.is_parent? ||
-       ((user_is_member_of?(group.parent.id) && group.parent.members_can_create_subgroups?)) ||
-       user_is_admin_of?(group.parent.id)
+      user_is_admin_of?(group.parent.id) ||
```
gdpelican: (feel free to ignore if annoying :) )

We should be doing `parent_id` instead of `parent.id`, which'll save us a DB request

lineman/app/components/invitation_form/team_link_modal/team_link_modal.coffee
```diff
@@ -0,0 +1,20 @@
+angular.module('loomioApp').factory 'TeamLinkModal', ->
+  templateUrl: 'generated/components/invitation_form/team_link_modal/team_link_modal.html'
+  controller: ($scope, Records, LoadingService, group, AppConfig, FlashService) ->
+    $scope.group = group
+    $scope.loading = true
+
+    $scope.load = ->
+      Records.invitations.fetchShareableInvitationByGroupId(group.id)
```
gdpelican: This doesn't seem to be a method anywhere?

gdpelican: http://g.recordit.co/LTesn0kbum.gif

gdpelican:
Hm, the team link still isn't working for me, and there's a bunch of failing tests; could we resolve those before we merge?

Looks grrreat overall.

lineman/app/components/invitation_form/team_link_modal/team_link_modal.coffee
robguthrie: that is sooo super weird.. must have fallen through the git cracks somehow. Thank you!

app/models/ability.rb
robguthrie: na that's great. thank you.

app/controllers/api/memberships_controller.rb
robguthrie: yesyes

app/controllers/api/memberships_controller.rb
robguthrie: Oh look.. now I don't have to pass parent group id at all. Nice

robguthrie:
This is now running on loomio clone and works well.

PR #2739, opened 2015-11-19
Update client.en.yml

rdbartlett:
Saw this and felt sad:

image

Propose we switch out "position" for "vote".


robguthrie:
Super! Really agree it's the more correct word.

PR #2748, opened 2015-11-23
Fix admin subscription handling

gdpelican:
@mhjb Watch the PR, once it gets merged into master I'd like to smoke test it on clone / test to make sure that it meets your needs alright before we deploy

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -4,8 +4,14 @@ class Subscription < ActiveRecord::Base
   validates :group, presence: true
   validates_inclusion_of :kind, in: ['trial', 'gift', 'paid']
 
+  delegate :id, to: :group, prefix: true, allow_nil: true
+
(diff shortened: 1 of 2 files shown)


app/models/subscription.rb
```diff
@@ -4,8 +4,14 @@ class Subscription < ActiveRecord::Base
   validates :group, presence: true
   validates_inclusion_of :kind, in: ['trial', 'gift', 'paid']
 
+  delegate :id, to: :group, prefix: true, allow_nil: true
```
gdpelican: We don't usually create subscriptions independently of groups, so this is awkward (we're basically defining a wannabe foreign key for when we create a subscription so we can define the group... making me wonder if we have this `has_one`/`belongs_to` relationships backwards?)

Also I just wanted to get this out the door; let me know if I'm missing something obvious @robguthrie

PR #2750, opened 2015-11-24
Exclude deactivated accounts from memberships page

HSalmon:


app/controllers/api/memberships_controller.rb
```diff
@@ -55,7 +55,7 @@ def remove_admin
   private
 
   def accessible_records
-    visible = resource_class.joins(:group).includes(:user, :inviter, {group: [:parent, :subscription]})
+    visible = resource_class.joins(:group).includes(:user, :inviter, {group: [:parent, :subscription]}).active
```
gdpelican: This'll make it tough if we have the desire to make an endpoint which displays deactivated memberships... not sure if that's a real concern or not.

Anyway I wonder if this would work just as well in the `index` action:

```
# memberships_controller.rb#5
collection.active.where(group_id: @group.id).order('users.name')
```

gdpelican:
:ok_hand:

PR #2761, opened 2015-11-27
Group form tweaking

gdpelican:
Hey @robguthrie let me know what you think of this fix; the current issue as I understand it:

Currently, we can change from say 'closed' to 'open' privacy without updating the underlying 'discussionPrivacyOptions', meaning it's possible to submit a form which has 'open' privacy and 'private_only' discussions, which = 422.


robguthrie:
Looks good!

PR #2762, opened 2015-11-27
Fix AWOL welcome modal

gdpelican:
Break out the conditions for the welcome modal to make them a little easier to read, and only show welcome modal to the creator (we don't have memberships loaded at this point, so we don't know that the current user is an admin or not. Also, the welcome modal goes away after anyone's invited to the group, so it's really just the creator that needs to see it.)


robguthrie:
gah! thank you!

PR #2766, opened 2015-12-01
Load proposals correctly on thread page

gdpelican:
Fix for votes not loading on previous proposals card for thread page

PR #2775, opened 2015-12-02
More proper fix for my votes not appearing

gdpelican:
Allow filtering by group and discussion for my_votes action on votes_controller

More better than #2766


robguthrie:
that's awesome

gdpelican:
Hmm.. giving this a smoke on clone, I see that the previous proposals page now includes _all_ proposals you've voted in, because we're loading up all of your proposals when we load up all of your votes.

I think there's no need to do `has_one :proposal` on a vote anymore, but maybe it's better to filter or limit them client side? @robguthrie

PR #2776, opened 2015-12-02
Live update mark as read

gdpelican:
Am I nuts? I think this might be all we need to make live updating of read content work across tabs.

I'm going to spin up a faye server later hopefully and give this a spin.

### DIFF
--- a/app/models/discussion_reader.rb
+++ b/app/models/discussion_reader.rb
@@ -92,6 +92,13 @@ class DiscussionReader < ActiveRecord::Base
       self.last_read_at = read_at
       reset_counts!
     end
+
+    publish!
+  end
+
+  def publish!
+    publish_data = Simple::DiscussionSerializer.new(discussion, scope: { reader_cache: reader_cache })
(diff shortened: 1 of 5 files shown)


robguthrie:
yea nah this is what I've always thought was required. thing is you made message channels so great it's this easy now.

robguthrie:
I wonder if we could use a light weight serializer for this..  something that just updates the specific attributes about reading discussions.

gdpelican:
Whoa, so much yes.

gdpelican:
I believe this should fix #2774 as well, the issue being that we're marking things as read in the DB when we create a comment, but not telling the client.

robguthrie:
Yea it sure will. Are you happy to make a discussion_marked_as_read_serializer before we merge this?

gdpelican:
Should be already in there as Simple::DiscussionSerializer

robguthrie:
Arrr cool. Happy to merge this when you're happy.

PR #2787, opened 2015-12-03
Hide panel when no undecided members

gdpelican:
For #2727


robguthrie:
Hrrmm.. should this be looking at counter cache _count columns rather than expecting records to be loaded locally?

robguthrie:
@gdpelican any thoughts about my above question? I think undecidedMembers assumes all the records have been loaded already.. If we move to _count based system we don't need to load all the records first.

gdpelican:
Yep, happy to make that change.

Issue #2740, opened 2015-11-19
Nonsense timestamps on notifications


rdbartlett:
![image](https://cloud.githubusercontent.com/assets/970124/11283477/cd26f41e-8f69-11e5-821c-1dba11d2dd73.png)

gdpelican:
@rdbartlett Sorry, I'm not seeing the nonsense here?

rdbartlett:
Sorry, bad explanation. I get a lot of timestamps that say "just now", but
they are actually hours old. It's like the clock gets frozen when the event
first fires.

gdpelican:
I believe the auto-updating timestamps will fix this.

PR #2788, opened 2015-12-03
Auto-updating timestamp

gdpelican:
This pulls in a directive to allow the timestamps to auto-update as they sit there on the frontend. Pretty neat stuff.

For #2740


robguthrie:
You rule.

PR #2790, opened 2015-12-07
Bootstrap task: installing dependencies and lunching project with one line (hopefully)

skatkov:
Good day, 

James invited me to contribute to project and I got annoyed by project setup process. As a developer, I can managed with most of those annoyances, but it's impossible to create user from interface and there is not a single mention about that in a documents. Dropping to console just to do that seemed like a horrible hack. 

So, i've decided to create this 'bootstrap' task. In ideal world it would automatically set everything up and lunch the project. But current version has some limitations (installing ruby, which is managed by rbenv/rvm, bundler gem, npm and PG database).

i've hacked this together in couple of hours i've had free. And decided to put this for discussion as a Pull Request, i'm not really sure if this is needed for a project and if i should continue my work on this.

### DIFF
--- a/lib/tasks/bootstrap.rake
+++ b/lib/tasks/bootstrap.rake
@@ -1,0 +1,96 @@
+namespace :bootstrap do
+
+  desc 'Retrieve dependencies needed for project to run'
+  task :dependencies do
+    unless bundler_installed?
+      puts "Project requires bundler to run: 'gem install bundler'"
+      puts "Quitting .."
+      return
+    end
+
+    unless npm_installed?
+      puts "Project requires npm to run: 'brew install npm' "
+      puts "Quitting .."
+      return
+    end
+
+    sh 'bundle install'
+    sh 'npm install -g lineman' unless lineman_installed?
+    sh 'npm install -g bower' unless bower_installed?
+    sh 'bower install'
+  end
+
+  desc "Create database.yml file"
+  # don't add :enviroment here, you can't load Rails without database file
+  task :config_files do
+    return if  File.exists?(File.join(Dir.pwd, "config", "database.yml"))
+
+    source = File.join(Dir.pwd, "config", "database.example.yml")
+    target = File.join(Dir.pwd, "config", "database.yml")
+    FileUtils.cp_r source, target
+  end
+
+  desc "Setup database for test and development enviroment"
+  task :db => :environment do
+    %w(development test).each do |env|
+      db = ActiveRecord::Base.configurations[env]['database']
+      user = ActiveRecord::Base.configurations[env]['username']
+      pass = ActiveRecord::Base.configurations[env]['password']
+
+      create_db db, user, pass
+      puts "Database #{user}@#{db} created for #{env}"
+    end
+
+    system('rake db:migrate') ? 'database migrated' : 'database migration failed'
+  end
+
+  desc 'Create user (option arguments email, password)'
+  task :create_user, [:email, :passcode] => :environment do |t, args|
+    args.with_defaults[:email => 'default@loomio.com', :passcode => 'passcode1']
+    User.create(email: args[:email], password:args[:passcode])
+    puts "Created user with email #{args[:email]} and password '#{args[:passcode]}'"
+  end
+
+  task :run => :environment do
+    #lunch rails here
+    #lunch lineman here
+    # https://loomio.gitbooks.io/tech-manual/content/using_development.html
+  end
+
+  private
+
+
+  def npm_installed?
+    `which npm`
+    $?.success?
+  end
+
+  def bundler_installed?
+    `which bundle`
+    $?.success?
+  end
+
+  def lineman_installed?
+    `which lineman`
+    $?.success?
+  end
+
+  def bower_installed?
+    `which bower`
+    $?.success?
+  end
+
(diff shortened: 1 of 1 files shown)


gdpelican:
This looks pretty great!

Am I right in thinking that we've got 4 rake tasks here that we can run independently, plus a fifth which attempts to run them all in order? (ie, something which takes you from a freshly cloned git to a working dev env in one command?)

lib/tasks/bootstrap.rake
```diff
@@ -0,0 +1,96 @@
...
+  desc 'Create user (option arguments email, password)'
+  task :create_user, [:email, :passcode] => :environment do |t, args|
+    args.with_defaults[:email => 'default@loomio.com', :passcode => 'passcode1']
+    User.create(email: args[:email], password:args[:passcode])
+    puts "Created user with email #{args[:email]} and password '#{args[:passcode]}'"
+  end
+
+  task :run => :environment do
+    #lunch rails here
+    #lunch lineman here
```
gdpelican: I reckon we could do something like

```
cd lineman && lineman run &
rails s
```
, (the trailing & telling linux to spin it off to a process other than the one in the shell.)

I'd love to run Loomio in one command, but my experience has been that it's useful to have access to the lineman log so you can see when it updates (and when it breaks, which it does sometimes). Thoughts on this @robguthrie ?

lib/tasks/bootstrap.rake
```diff
@@ -0,0 +1,96 @@
...
+      puts "Database #{user}@#{db} created for #{env}"
+    end
+
+    system('rake db:migrate') ? 'database migrated' : 'database migration failed'
+  end
+
+  desc 'Create user (option arguments email, password)'
+  task :create_user, [:email, :passcode] => :environment do |t, args|
+    args.with_defaults[:email => 'default@loomio.com', :passcode => 'passcode1']
+    User.create(email: args[:email], password:args[:passcode])
```
gdpelican: Likely want to make this `is_admin: true` as well

I also like hash#fetch syntax for this sort of thing:

```
User.create email: args.fetch(:email, 'default@loomio.org'), 
                    password: args.fetch(:password, 'mypassword')
```

lib/tasks/bootstrap.rake
```diff
@@ -0,0 +1,96 @@
...
+    $?.success?
+  end
+
+  def bower_installed?
+    `which bower`
+    $?.success?
+  end
+
+  def create_db( db, user, pass )
+    %x{/usr/bin/creatuser -d -l -R -S #{user} &>/dev/null}
```
gdpelican: `createuser` typo here?

rdbartlett:
Wow awesome contribution thanks @skatkov! Simple genius :gem:

skatkov:
@gdpelican your right.
There are 5 tasks that you can run independently, but if you just run 'rake bootstrap' it should take you from zero to hero (but only if you have ruby, bundler, pg and npm currently) by running those tasks in proper order.

Logically, I would put copy database.yml file to bootstrap:db task. But you can't really lunch rake task with environment loaded if database.yml file is missing, so i've had to make a separate task for that.

lib/tasks/bootstrap.rake
skatkov: @gdpelican i can lunch rails in separate process and leave lineman running in this one so logs will be visible.

skatkov:
@gdpelican 
It looks like i'm finished, all tasks i wrote work separately, i didn't had a chance to test it with a clean system.

I removed self written task that creates database and used a standard one (db:setup). And bootstrap:run lunches both rails s and lineman and prints all logs :)

lib/tasks/bootstrap.rake
```diff
@@ -0,0 +1,89 @@
...
+      FileUtils.cp_r source, target
+      puts "Database.yml file created"
+    else
+      puts "Database.yml file already exists"
+    end
+  end
+
+  desc 'Create user (optional arguments email)'
+  task :create_user, [:email, :password] => :environment do |t, args|
+    args.with_defaults[email: 'default@loomio.com', password: 'passcode1']
```
gdpelican: This isn't working for me with `[]`, doesn't it want to be

```
args.with_defaults(email: '', password: '')
```
?

lib/tasks/bootstrap.rake
```diff
@@ -0,0 +1,89 @@
...
+    if User.find_by(email: args[:email]).empty?
```
gdpelican: Running `empty?` on `nil` is going to throw an error, so this is no good. Either switch up the logic to use present, or use `nil?` here instead?

lib/tasks/bootstrap.rake
```diff
@@ -0,0 +1,89 @@
...
+      User.create(args.to_hash)
+      puts "Created user with email #{args[:email]} and password '#{args[:password]}'"
+    else
+      puts "User with #{args[:email]} already exist"
+    end
+  end
+
+  desc "Lunch project"
```
gdpelican: Minor nit: should be 'launch' :)

gdpelican:
Merging this; I'll apply the user tweaks in another PR

gdpelican:
Thanks @skatkov, this rocks!

PR #2796, opened 2015-12-09
Add ability to mark all discussions in group as read

HSalmon:


lineman/app/models/group_model.coffee
```diff
@@ -148,6 +148,12 @@ angular.module('loomioApp').factory 'GroupModel', (DraftableModel, AppConfig) ->
         @remove()
         _.each @memberships(), (m) -> m.remove()
 
+    markAsRead: =>
+      @remote.postMember(@key, 'mark_as_read')
+      _.map @discussions(), (discussion) ->
+        return unless discussion.isUnread()
+        discussion.update(lastReadAt: moment(), readSalientItemsCount: discussion.salientItemsCount)
```
robguthrie: Is there a way we can reuse discussion.markAsRead() and not fire a network request when doing that?

lineman/spec-e2e/inbox_spec.coffee
```diff
@@ -13,3 +13,8 @@ describe 'Inbox Page', ->
 
     expect(inboxHelper.anyThreads()).not.toContain('Muted discussion')
     expect(inboxHelper.anyThreads()).not.toContain('Old discussion')
+
+  it 'can mark all threads in a group as read', ->
```
robguthrie: I've been pushing for a new way of writing our e2es.

Using the PageHelper module, you could rewrite this as:
```
page.expectElement('.inbox-page__group')
page.click('.inbox-page__mark-as-read')
page.expectNoElement('.inbox-page__group')
```

robguthrie:
Looks Good!

lineman/app/models/group_model.coffee
gdpelican: Actually the more I look at this the more I don't like it. We're getting out of sync with the server here, don't have a good way of handling failure, and are left with weirdo guessed-at values designed to make `isUnread()` false, which don't get updated because we're not responding with anything useful from the server. We also don't update other attributes like `readCommentsCount` and `readItemsCount` here, meaning they're out of sync on the record itself.

I think we need to just wait for the request to come back, perhaps marking the threads with something like `readOnClient` which could exclude them from the unread count in the meantime.

robguthrie: What failures are possible?

I think the weird guess at values could be legitimate.. If we use the
values from the last known discussion item. Any newer data can just arrive
down the message channel. I can't see why it's a problem to update to
latest known values client side.

We really do want this to update immediately, user testing tells us that
it's a really bad time to click "go away now" and have to wait. Also
marking many discussions as read is really slow on the server side.

gdpelican: Could be 404 bad network connection, could be you logged out elsewhere, your token expired, we 500... Who knows, servers are weird, heh. I have a strong suspicion we could end up with records where the activityCount is 0, but readCommentCount < commentCount, and that just seems weird and fragile.

I think the solution here is to do `discussion.update(readOnClient: true)` (reads better anyway), and then isUnread becomes
```
isUnread: ->
  @readOnClient or # current isUnread logic
```

Then when Faye updates the record, it'll populate the correct values from the server, and set readOnClient back to false (we'll serialize out readOnClient as false with each discussion)

robguthrie: cool sounds great

gdpelican: Whoa, now that I'm looking at the code, your suggestion of using discussion_model's mark as read is perfect. There's even that really cool `invoke` function on underscore we can use.

HSalmon:
@robguthrie, @gdpelican, I'm not sure what's happening with this failure. Thoughts?

robguthrie:
@HSalmon if you get errors on your e2e specs in travis you can expand the last section and find a link to the screenshots of the failures. See "report available at:" after I clicked on the expand arrow beside "travis upload screenshots"

image

So looking at that we get to this screenshot of the failure:
https://s3-us-west-2.amazonaws.com/loomio-travis2/20151217_0158-2796/838dce2c38181add24bc5f98397e4019.png

Which tells us that the page did not load.. which is strange.. I'm still not sure what's going on, happy to debug with you tomorrow.

gdpelican:
I've patched up the tests so this passes; last thing would be to update the e2e to how you like it, if you want @robguthrie @HSalmon

PR #2797, opened 2015-12-10
Legacy trial expired

robguthrie:


robguthrie:
image

and on a thin screen:
image

lineman/app/components/group_page/trial_card/choose_plan_modal/choose_plan_modal.coffee
```diff
@@ -2,6 +2,7 @@ angular.module('loomioApp').factory 'ChoosePlanModal', ->
   templateUrl: 'generated/components/group_page/trial_card/choose_plan_modal/choose_plan_modal.html'
   size: 'choose-plan-modal'
   controller: ($scope, group, ModalService, ConfirmGiftPlanModal, CurrentUser, AppConfig, $window, IntercomService) ->
+    console.log group
```
gdpelican: Can we remove this before merge?

lineman/app/components/legacy_trial_expired_modal/legacy_trial_expired.coffee
```diff
@@ -0,0 +1,23 @@
+shownToGroup = {}
+angular.module('loomioApp').factory 'LegacyTrialExpiredModal', (ModalService, AppConfig)  ->
+
+  appropriateToShow: (group, user) ->
+    !shownToGroup[group.id] and
```
gdpelican: This user is always CurrentUser, yeah?

I think this wants to live in something like `AbilityService.canSeeExpiryModalFor(group)`?

robguthrie:
This needs:
smoke test on loomio clone to confirm that donation and subscription flows work
only show for groups included in list of 100 id's to be provided by mhjb and set via ENV.

Will pick it up again on Sunday sometime.

lineman/app/components/legacy_trial_expired_modal/legacy_trial_expired.coffee
robguthrie: If the logic is never shared anywhere else I think it feels better to keep it all in the same place.

lineman/app/components/group_page/trial_card/choose_plan_modal/choose_plan_modal.coffee
robguthrie: yes, will do!

robguthrie:
As it looks now:
image

PR #2799, opened 2015-12-11
Red notifications badge, border around both badges

mhjb:
before
=====

image

after
====

image

### DIFF
--- a/lineman/app/components/navbar/navbar.scss
+++ b/lineman/app/components/navbar/navbar.scss
@@ -112,9 +112,10 @@
   font-size: 10px;
   position: absolute;
   left: 16px;
-  top: 1px;
+  top: 2px;
   background: #4A90E2;
   border-radius: 10px;
+  border: 2px solid white;
   padding: 5px;
   color: white;
   line-height: 5px;


mhjb:
@rdbartlett

gdpelican:
I am a fan of the badge borders, but am not a fan of the red. Feels very
unharmonious with the rest of the interface to me.

rdbartlett:
Yep borders are nice, but why the red? In Loomio, red means block or
delete.

mhjb:
Read means important

image

image

image


This PR stimulated by watching users not noticing their Loomio notifications.

rdbartlett:
I abstain.

gdpelican:
I'm maintaining my disagree. We can up the visibility of notifications with an animation on update and with not having two versions of notifications on opposite sides of the screen, but having yellow, blue and red on the navbar makes this looks really bad imho.

robguthrie:
feels like da cops

gdpelican:
Gonna merge this in without the red for now, could be a good thing to circle back on a notification color scheme strategy at some point... Discourse has a rainbow of notification colors and I think they're a little tough to grok, but maybe there's some color scheme out there that makes sense.

PR #2800, opened 2015-12-11
Add OAuth2 provider support, accept access_token in API

Connoropolous:
see this comment for todo list https://github.com/loomio/loomio/pull/2800#issuecomment-164912379


config/locales/doorkeeper.en.yml
```diff
@@ -0,0 +1,123 @@
+en:
```
rdbartlett: @robguthrie can we exclude this from Transifex? I think it will confuse the living eff out of most of our translators :)

rdbartlett:
Connor it is so exciting to have this underway!! Yeeeeeee :pineapple: :nail_care:

gdpelican:
Phew such firehose, but this looks fantastic so far. Thanks for the great work!

Connoropolous:
step 1, as an admin, create an application
screen shot 2015-12-10 at 10 49 34 pm

step 2, as a user, authorize an application

screen shot 2015-12-10 at 11 02 49 pm

step 3, as an app developer, use the received auth code to grab an access_token from loomio
```
curl -F grant_type=authorization_code \
-F client_id=31d4b4fdfe2b791108259ae13ab40e94429757717dd64d39e99b020f5ab27fc3 \
-F client_secret=7896a43f8f4bd47bc6a6f1ea82abd974b52386b08dc02726c294378cbe524df6 \
-F code=c2590347524f6956dace1e54409ff2b62034047d2280fc1800e92c2c2f53d015 \
-F redirect_uri=urn:ietf:wg:oauth:2.0:oob \
-X POST http://localhost:3000/oauth/token
```

step 4, start making requests as a user over api
```
curl -H "Content-Type: application/json" \
-X POST \
-d '{"access_token": "d2f4beff551e6beb89b9d45f7c8a826f1ac1bde0f5f25e19acf0733b2ce5a119", "group": { "name": "test" }}' \
http://localhost:3000/api/v1/groups
```

elimisteve:
This is OAuth between Loomio and... what?  Does this make Loomio an OAuth provider/server that people can log into and have other apps interface with users' Loomio account/data?!

gdpelican:
It's the first step in that direction, yes.

elimisteve:
Awesome!!!  Very important work!  Yay @Connoropolous @gdpelican @robguthrie!!!

config/locales/doorkeeper.en.yml
robguthrie: Good spotting Rich!

It's weird to me that it's in here at all. I guess it's here so you can easily modify them.

 If we delete it, I wonder if doorkeeper will fallback to strings within it's own gem? They would have this translated already.

robguthrie:
Sorry I'm so late to the party! This looks super duper great.

So with this, users can generate an API token to use?

What are your thoughts about when we could merge this in?

Connoropolous:
still needed...

- [ ] lock down the creation of applications to admins?
- [ ] make the application creation page UI look like loomio, not boilerplate UI.
- [ ] if signed out when attempting to authorize an app, set up follow up redirect (currently not happening)
- [ ] define scopes and actions available to those scopes?

and that's just for the oauth stuff itself... I am wondering how does development happen, maybe we could merge this in to some branch on loomio and then the follow up development on the api routes themselves could happen as separate PRs? 

@robguthrie "So with this, users can generate an API token to use?" yes.

config/initializers/doorkeeper.rb
```diff
@@ -0,0 +1,109 @@
...
+  #   http://tools.ietf.org/html/rfc6819#section-4.4.2
+  #   http://tools.ietf.org/html/rfc6819#section-4.4.3
+  #
+  # grant_flows %w(authorization_code client_credentials)
+
+  # Under some circumstances you might want to have applications auto-approved,
+  # so that the user skips the authorization step.
+  # For example if dealing with a trusted application.
+  # skip_authorization do |resource_owner, client|
+  #   client.superapp? or resource_owner.admin?
```
Connoropolous: I think if we were going to try to make the loomio front end use token based auth we would use this.

I believe what it would let us do is retrieve a token automatically for an authed user without them hitting the 'authorize' page, if they are logged in or if they hit the sign_in page

robguthrie:
Just caught up on doorkeeper readme, and the [rails cast on it](http://railscasts.com/episodes/353-oauth-with-doorkeeper).
 
I'm thinking about permissions design and use cases. How do we handle the per group authorisation case? 

Is it intended that you send the group id as a scope when asking for an access token?

Connoropolous:
I wonder if, for V1, we could just treat the generated token as the authed user and assume they have the same read write access as regular usage. 

Not totally sure how important, how soon, it would be to have that 'per group' case, nor whether implementing without initially poses an unhealthy constraint towards enabling that down the road.

Connoropolous:
all loomio activity takes place within a context of a group, correct?

robguthrie:
yea... right now it all does.

Connoropolous:
@robguthrie 
"I'm thinking about permissions design and use cases. How do we handle the per group authorisation case?
Is it intended that you send the group id as a scope when asking for an access token?"

when you said you were thinking about use cases, did you have something solid in mind with the 'per group' authorization? 

I am thinking that one way that this could get solved is by actually modifying the 'app authorization' page, to include the selection of which groups the token has access to, rather than passing it beforehand as an id.

robguthrie:
We've just had another support request for data export. 

I wonder if, instead of targeting MVP as "loomio client uses OAuth", we aim to build a "backup group script", which writes a file with all the json encoded data for a group.

This would mean writing fairly exampley code for others to write their own interesting scripts against the API.

What do you think?

gdpelican:
I don't think group json dump will get us where we need to go as far as interoperability goes. I also think we're overcomplicating our thinking on this; my primary use case is:

"I want to be able to use the Loomio API as my user, without using a browser"

So, with a series of curl commands, I should be able to do anything I'd be able to do in-app. Our existing permissions (which are quite good), should handle whether an authed user can do certain things or not. Happy to be missing something, but I don't think we need scopes to accomplish this? I can see the use case for a group-only token, but don't think it needs to happen in this run.

Also, what's the reasoning behind restricting admins to creating applications?

Connoropolous:
Yes, with you on your last comment @gdpelican ! 

"Also, what's the reasoning behind restricting admins to creating applications?"
um, I put a question mark there because I feel like that's up to the loomio team to decide whether you want this to work like
1. anybody can create applications, which would require a well developed UI and functionality
2. only admins can create applications, which would require minimal UI development and could start to be tested with sooner rather than later.

ahdinosaur:
@Connoropolous @gdpelican @robguthrie wow this is really great. :)

i'm interested in building a [micro-service to log-in using a loomio auth token and generate a live stream of people, group, and membership models](https://github.com/holodex/loomio-data), this is super helpful for that, cheers.

Connoropolous:
Great @ahdinosaur!

gdpelican:
I reckon UI's out of scope. To give an example, we developed a slack integration and straight up didn't tell anyone about it at all... we started using it internally, and it slowly leaked out to Enspiral and a few other orgs, and now we pretty much just give it to anyone who's heard about it for 'beta' testing. I'd be totally happy for a similar thing to happen here like

1. Mikey tells us he wants to make an application
2. We do a task in the command line / barebones admin panel
3. Mikey has an application, yay!

And we'll just let anyone who asks about it use it as they will - this has the benefit of a) letting people do the thing, and b) not having them be upset if it's a bit wonky to start (maybe they'll even help fix it! ;|)

Connoropolous:
Solid, ok.

I'll try and hack the doorkeeper boilerplate UI just a bit to be
Loomio-ized for the applications registration pages, as well as the
'authorize app' pages.

Let me know if the application creation should be limited to admins for
now, or whether we just don't publicize the route.

:)

gdpelican:
No restrictions, free and easy. :palm_tree:

Connoropolous:
Ok! I've got the templates moved over into the loomio layouts, but at least two pages need a little bit of styling/UI work. Could someone familiar with that take on that basic task?

The pages to do are the `authorize

Connoropolous:
sorry about the half comment. edited it with the complete message

Connoropolous:
- [ ] do a bit of styling
- [ ] finish enabling api routes to make use of token as authed user
- [ ] rebase onto master, in preparation for merge

Connoropolous:
Note that doorkeeper doesn't have the 'implicit' grant type flow enabled by default, we would have to enable it if we want. The default is the 'authorization code' type. You can read something explaining the differences here:
https://www.instagram.com/developer/authentication/

robguthrie:
maybe add to that list: reject non https requests when production env

Connoropolous:
adding

robguthrie:
Connor you are awesome.

Connoropolous:
thanks Rob :)
@robguthrie @gdpelican where could we turn regarding this question? 
"I've got the templates moved over into the loomio layouts, but at least two pages need a little bit of styling/UI work. Could someone familiar with that take on that basic task?
The pages to do are the authorize page and the authorized page."

gdpelican:
We can take care of that, no problem. I'll put it on my list. We shouldn't
let that hold up getting this in in any case.

gdpelican:
I've done a little styling on this, @Connoropolous; nothing fancy; pushed up into an `oauth-clone` branch on the loomio repo; I believe you can cherry pick it from there (commit hash is 412420ab3c7cc3a8fbffaa3aef2ac8a0de7a36fd)

Connoropolous:
Sweet James! Yes, I should be able to

>

-- 
Connor Turland
~to craft a sustainable and thriving culture within and without myself~
co-creator of Metamaps.cc

Connoropolous:
hey folks, I am in New Zealand now :) limited net access, but enough here and there to keep poking at this. ultimately the sweetest thing would be for me to be able to meet up with some of you when I'm in Wellington, that's going to be between the 20-28 of January that I'm going to be there. 

@gdpelican for today, I've just cherry picked your commit and it's in this pull request now :)

robguthrie:
Sound great. Keen to meet you in person then Connor!

app/controllers/api/restful_controller.rb
```diff
@@ -18,7 +20,8 @@ def destroy_action
   private
 
   def current_user
-    super || LoggedOutUser.new
+    token_user = User.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
```
Connoropolous: @gdpelican I'm still not quite clear on whether this PR should include more assurance that other API routes will work with this, or whether we should just go ahead with this, and then follow on with deeper implementations into the rest of the API routes. Do you have a sense?

gdpelican: Not quite sure I follow; this current_user method is used throughout the API (and I believe it's the only way we reference the current user anywhere in the API), so this should cover all of our API actions.

I would say caching the doorkeeper user so we don't have to hit the DB every time we reference current_user might be a good call though;

```
def current_user
  super || token_user || LoggedOutUser.new
end

private

def token_user
  @token_user ||= User.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
end
```

app/controllers/api/restful_controller.rb
```diff
@@ -1,6 +1,8 @@
 class API::RestfulController < ActionController::Base
   snorlax_used_rest!
 
+  before_filter :doorkeeper_authorize!, only: [:create_action, :update_action, :destroy_action]
```
gdpelican: I'm skeptical of this; `create_action`, `update_action`, and `destroy_action` aren't actually controller actions, they're called from the `create`, `update`, and `destroy` actions (and looking at this now they should probably be private to help make that more obvious).

We're using this gem: https://github.com/loomio/snorlax to abstract away heaps of the actual controller implementation, so there's even more opinionated magic than regular rails land (wee!)

In short, the controller actions look like this:
```
def create
  build_resource
  create_action
  respond_with_resource
end

def update
  load_resource
  update_action
  respond_with_resource
end
```
(etc)

with create_action, update_action being overriddes for this particular app.

app/controllers/api/restful_controller.rb
gdpelican: Not quite sure I follow; I think this method is the only thing we use anywhere in the API to refer to the current user, so modifying this method should cover the whole of the API.

Might be nice to have some testing around this?

gdpelican: Also, I reckon caching the doorkeeper user so we don't have to hit the database every time we reference the current user would be nice.

```
def current_user
  super || token_user || LoggedOutUser.new
end

private

def token_user
  @token_user ||= User.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
end
```

app/controllers/oauth_authorized_applications_controller.rb
```diff
@@ -0,0 +1,14 @@
+class OauthAuthorizedApplicationsController < Doorkeeper::AuthorizedApplicationsController
+
+  helper_method :dashboard_or_root_path
+
+  protected
+
+  def dashboard_or_root_path
```
gdpelican: Hm, smells like a concern to me? We already have something called dashboard helper, but doing something like `include RootPathHelper` in each of these controllers (and `ApplicationController`) might be nice.

app/controllers/api/restful_controller.rb
Connoropolous: thanks @gdpelican. I will investigate caching the doorkeeper and writing some basic testing

app/controllers/oauth_authorized_applications_controller.rb
Connoropolous: yes, this was totally a hack just to get it working. there's duplicate code in these three files. I am still unclear what the solution is after your comment, but I'll have a go at it soon.

Connoropolous:
@gdpelican do you think we could possibly tackle a couple of the last little things as a pair coding thing?

gdpelican:
Sure thing, a pairing day would be great.

gdpelican:
Hey @Connoropolous, I'm thinking about slapping this on my to-do list for the next sprint, either through pairing with you, or by pushing it over the edge from here. How would that sit with you?

Thanks for getting the ball rolling on this; it's really really important work.

Connoropolous:
So cool! I think just run with it. I am planning on coming back to Welly on
Feb 29 for 11 days so if it isn't done by then, we can pair!

gdpelican:
Closing in favor of #2980

PR #2804, opened 2015-12-12
Restructure CommentService events

gdpelican:
Today's :airplane: treat!

I did this as a thought experiment more than anything else, to decouple the concerns of the Service layer from the various events; this implements changes to the NewComment and UserMentioned events.

Something like this is an important piece towards a flexible, extensible plugin architecture, because it means people can hook into the things going on in the app without having to muck about in the things that are calling them.

Would be cool to hear your thoughts on honing this approach some @robguthrie 

(Hi from Amsterdam!)

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -26,11 +26,13 @@ class CommentsController < BaseController
 
   def like
     CommentService.like(comment: @comment, actor: current_user)
+    @comment.reload
     render template: "comments/comment_likes"
   end
 
   def unlike
     CommentService.unlike(comment: @comment, actor: current_user)
+    @comment.reload
     render template: "comments/comment_likes"
   end
 
--- a/app/extras/event_bus.rb
+++ b/app/extras/event_bus.rb
@@ -1,0 +1,29 @@
+class EventBus
+  include Singleton
+
+  def broadcast(event, *params)
+    listeners[event].each { |listener| listener.call(*params) }
+  end
+
+  def listen(*events, &block)
+    events.each { |event| listeners[event].add(block) }
+  end
+
+  def deafen(*events, &block)
+    events.each { |event| listeners[event].delete(block) }
+  end
+
+  def clear
+    @listeners = nil
+  end
+
+  # allows us to call EventBus.listen, instead of EventBus.instance.listen
+  [:broadcast, :listen, :deafen, :clear].each { |m| define_singleton_method m, ->(*args, &block) { instance.send(m, *args, &block) } }
+
+  private
+
+  def listeners
+    @listeners ||= Hash.new { |hash, key| hash[key] = Set.new }
+  end
+
+end
--- a/app/models/discussion_reader.rb
+++ b/app/models/discussion_reader.rb
@@ -18,17 +18,27 @@ class DiscussionReader < ActiveRecord::Base
     end
   end
 
+  def self.for_comment(comment:)
+    self.for(user: comment.author, discussion: comment.discussion)
+  end
+
+  def self.for_comment_vote(comment_vote:)
+    self.for(user: comment_vote.user, discussion: comment_vote.comment.discussion)
+  end
+
   def author_thread_item!(time)
     set_volume_as_required!
     participate!
     viewed! time
   end
+  EventBus.listen('comment_create') { |comment| for_comment(comment: comment).author_thread_item!(comment.created_at) }
 
   def set_volume_as_required!
     if user.email_on_participation?
       set_volume! :loud unless volume_is_loud?
     end
   end
+  EventBus.listen('comment_like') { |comment_vote| for_comment_vote(comment_vote: comment_vote).set_volume_as_required! }
 
   def participate!
     update_attribute :participating, true
--- a/app/models/events/comment_replied_to.rb
+++ b/app/models/events/comment_replied_to.rb
@@ -1,16 +1,9 @@
 class Events::CommentRepliedTo < Event
   def self.publish!(comment)
-    return unless comment.parent.present?
-    event = create!(kind: 'comment_replied_to',
-                    eventable: comment,
-                    created_at: comment.created_at)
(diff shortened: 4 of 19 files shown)


robguthrie:
Wow. This is really great. Nice one

robguthrie:
I'm so keen to move to this model asap. Let me know if you want me to do anything to help.

gdpelican:
Whoa, cool, glad to hear you're on board, I feel the same way.

I'll put some more grease into it later this week; I think doing new
comment first is really great because it's one of the hardest ones.

config/initializers/event_bus_mailers.rb
```diff
@@ -0,0 +1,9 @@
+
+EventBus.listen('new_comment_event') do |event|
```
robguthrie: is this supposed to be new_comment (without _event)?

app/extras/event_bus.rb
```diff
@@ -0,0 +1,29 @@
...
+  def deafen(event, &block)
+    listeners[event].delete(block)
+  end
+
+  def clear
+    @listeners = nil
+  end
+
+  # allows us to call EventBus.listen, instead of EventBus.instance.listen
+  [:broadcast, :listen, :deafen, :clear].each { |m| define_singleton_method m, ->(*args, &block) { instance.send(m, *args, &block) } }
```
robguthrie: Is this better than class methods and class variables?

config/initializers/event_bus_mailers.rb
gdpelican: No, there's two different events - one that's emitted after the comment is
created, one that's created after the new_comment event is created. We need
the event in this case because the email uses @event to generate the utm
hash for the links at the bottom (and maybe some other stuff)

robguthrie: oh i see now. never mind

app/extras/event_bus.rb
gdpelican: It's probably not better, per se. I have a probably unreasonable aversion to using class variables (@@something) for anything.

gdpelican:
Alright, this branch currently has an EventBus refactor for the whole CommentService now.

This feels like a really nice start to me. I'd be keen to call this one done, get it passing and smoked and merged in, and then iterating to the other services, so it's a more incremental architecture change.

app/controllers/comments_controller.rb
```diff
@@ -26,11 +26,13 @@ def show
 
   def like
     CommentService.like(comment: @comment, actor: current_user)
+    @comment.reload
```
gdpelican: Need this to pick up the changes to `liker_ids_and_names`, a thing we don't use on the angular side. I will look forward to not needing it anymore.

app/models/discussion_reader.rb
```diff
@@ -18,17 +18,27 @@ def self.for(user: , discussion: )
     end
   end
 
+  def self.for_comment(comment:)
```
gdpelican: I'm iterating this in the next pass to accept a vanilla model which has an author and a discussion

```
def self.for_model(model)
  self.for(user: model.author, discussion: model.discussion)
end
```

app/models/events/comment_replied_to.rb
```diff
@@ -1,16 +1,9 @@
...
-                    eventable: comment,
-                    created_at: comment.created_at)
-
-    if comment.parent.author != comment.author
-      ThreadMailer.delay.comment_replied_to(comment.parent.author, event) if comment.parent.author.email_when_mentioned?
-      event.notify! comment.parent.author
-    end
-
-    event
+    return unless comment.parent.present? && comment.parent.author != comment.author
```
gdpelican: This is a slight change in behaviour, but it's better I think. This means we don't even create an event for comment_replied_to when the replier is the same as the reply author.

Before we were creating the event, but not notifying anyone about it.

app/models/events/new_comment.rb
```diff
@@ -1,22 +1,10 @@
...
-            without(comment.parent_author).find_each do |mentioned_user|
-      Events::UserMentioned.publish!(comment, mentioned_user)
-    end
 
-    BaseMailer.send_bulk_mail(to: UsersToEmailQuery.new_comment(comment)) do |user|
-      ThreadMailer.delay.new_comment(user, event)
-    end
-
-    event
+  def self.publish!(comment)
```
gdpelican: This is getting pretty darn close to being able to generalize the `publish!` method and more it up into event.

Something like

```
# event.rb
def self.publish!(eventable)
  create(kind: kind, eventable: eventable, created_at: eventable.created_at).tap do |event|
    EventBus.broadcast("#{kind}_event", event
  end
end
```

app/models/search_vector.rb
```diff
@@ -49,6 +49,7 @@ def index!(discussion_ids)
     end
     handle_asynchronously :index!
   end
+  EventBus.listen('comment_create', 'comment_update') { |comment| index! comment.discussion_id }
```
gdpelican: Updated the event bus to allow it to accept several event names at the same time.

app/services/comment_service.rb
```diff
@@ -3,24 +3,18 @@ def self.unlike(comment:, actor:)
     return false unless comment.likers.include? actor
```
gdpelican: There's a pattern emerging in the services as well, although I'm not ready to generalize it quite yet.

config/initializers/event_bus_mailers.rb
gdpelican: I've changed this slightly to make the naming conventions a little more predictable.

For service calls, "#{model_class}_#{action_name}" ("comment_create")
For event calls, "#{event_kind}_event" ("new_comment_event")

I'm going to start emitting broadcasts for most service calls, as well as for event creation. This'll let us hook into most anything that's happening in the app pretty easily.

gdpelican:
Made more better in #2846

PR #2805, opened 2015-12-13
tool to rename, move translations

robguthrie:
`rake loomio:pluck_translations config/locales/en.yml email config/locales/emails.en.yml email`

Will move all translations with they key `email` in en.yml es.yml ... etc to the key email in emails.en.yml emails.es.yml etc

IE: We can move translations around without discarding all the work of our translators. I built this so we can break up our main .yml files into way smaller files.

### DIFF
--- a/lib/pluck_translations.rb
+++ b/lib/pluck_translations.rb
@@ -1,0 +1,52 @@
+module LoomioI18n
+  # next write spec that confirms some file is translated correctly.
+  # boose:
+  #   some_key: 'ok'
+  # subject:
+  #   thing: 'ues'
+
+  # gives identical destination file
+  # test inserts data
+  # test does not destroy existing keys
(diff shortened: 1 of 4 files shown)


robguthrie:
@mixmix have a wee look

gdpelican:
Noice!

robguthrie:
I think we want to:
Identify strings from en.yml that still apply to loomio 1.0 and pull them into new files. Emails, signup and login pages come to mind. 

I don't think we want more than 4 or so projects in transifex.. so this is not about breaking up translations by angular component.. that still needs discussion.

robguthrie:
so in building the new translation file I realise I need to support dot notation for keys. Then this will be ready to go.

ie: be able to move en.yml example_thread_content to server.yml example_content.welcome_thread

PR #2836, opened 2015-12-17
Remove link for recipient of comment reply

HSalmon:
@robguthrie, I think this should do the trick

### DIFF
--- a/lineman/app/components/thread_page/activity_card/new_comment.haml
+++ b/lineman/app/components/thread_page/activity_card/new_comment.haml
@@ -17,9 +17,7 @@
       %h3.new-comment__in-reply-to{ng-if: 'comment.parentId', id: 'event-{{event.id}}'}
         %strong
           %a{lmo-href-for: 'comment.author()'}> {{ comment.authorName() }}
-        %span{translate: 'new_comment_item.in_reply_to'}
-        %strong
-          %a{lmo-href-for: 'comment.parentAuthor()'}> {{ comment.parentAuthorName }}
+        %span{translate: 'new_comment_item.in_reply_to', translate-value-recipient: '{{comment.parentAuthorName}}'}
       %h3.new-comment__author-name{ng-if: '!comment.parentId'}
         %a{lmo-href-for: 'comment.author()'}> {{ comment.authorName() }}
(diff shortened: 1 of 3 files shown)


HSalmon:
screen shot 2015-12-18 at 11 30 12 am

robguthrie:
great

lineman/app/components/thread_page/activity_card/new_comment.haml
```diff
@@ -17,9 +17,7 @@
       %h3.new-comment__in-reply-to{ng-if: 'comment.parentId', id: 'event-{{event.id}}'}
         %strong
           %a{lmo-href-for: 'comment.author()'}> {{ comment.authorName() }}
-        %span{translate: 'new_comment_item.in_reply_to'}
-        %strong
-          %a{lmo-href-for: 'comment.parentAuthor()'}> {{ comment.parentAuthorName }}
+        %span{translate: 'new_comment_item.in_reply_to', translate-value-recipient: '{{comment.parentAuthorName}}'}
```
gdpelican: Why are we losing the link here? All that needs to be done to fix this is either add a check to `parentAuthor()` to not error when `parent()` is blank, or change the lmo-href-for to use `comment.parent().author()` instead of `comment.parentAuthor()`

robguthrie: I don't understand what you're saying. 

We don't serialize the parent author with the comment, so there are cases when the parent author is not loaded in the client so we don't know their user id

gdpelican: Oh, I see. specifically the case where we're jumping directly to a comment
and not back-loading it's parent.

Should we be serializing out the parentAuthorUsername alongside the
parentAuthorName then?

robguthrie: hrrmm.. yes I guess so. would be nice to have the whole user object though.
I wonder if we can do that with AMS yet.

gdpelican: What's the reasoning for not serializing out the parent with the comment? I
seem to remember we used to do that... was it too slow?

robguthrie: Active model serializer wouldn't do it.

Issue #2816, opened 2015-12-14
allow me to expand "new topic" textarea vertically


whit537:
I want to write more than will fit in the textarea as initially displayed, but I can't change the size of the textarea vertically using the browser's controls.

![screen shot 2015-12-14 at 12 13 38 am](https://cloud.githubusercontent.com/assets/134455/11773428/ac4f6262-a1f7-11e5-8826-1ba996f224fe.png)

whit537:
Ftr the textarea *does* expand when I fill it with content:

![screen shot 2015-12-14 at 12 15 35 am](https://cloud.githubusercontent.com/assets/134455/11773434/d840602e-a1f7-11e5-8c15-aea934888542.png)

gdpelican:
Yeah we're using a library to allow text areas to expand as content is entered into it. Likely we just want to stop the user from being able to expand the text areas manually at all.

PR #2845, opened 2015-12-18
Disallow manual text area resize

gdpelican:
@mhjb @rdbartlett Am I crazy for doing this?

For #2816

### DIFF
--- a/lineman/app/css/1_utilities.scss
+++ b/lineman/app/css/1_utilities.scss
@@ -22,6 +22,7 @@
 }
 
 .lmo-textarea {
+  resize: none;
   min-height: 100px;
 }
 


mhjb:
Mind explaining your rationale?

gdpelican:
Sure. The auto-expanding text area package means you can't resize vertically (it will only expand to the height of the text inside it). Messing with the horizontal size doesn't add much value, and more often than not looks like crap:

screen shot 2015-12-18 at 8 13 33 pm
screen shot 2015-12-18 at 8 13 51 pm

Thus... I'm not sure I can think of a good case for letting users resize their text areas.

robguthrie:
arr thanks. I agree remove it if we have auto resize

rdbartlett:
Sweet as

PR #2852, opened 2015-12-23
Mark all discussions in a group as read (take 2; don't merge)

gdpelican:


robguthrie:
I hate clicking mark as read, it feels gross for 2 reasons:

a) it is a total lie - i've not read it... i just don't want to know about the thread right now.
mute feels really harsh and I don't want to silence people. Honestly I just want to curate my dashboard so that next time I visit i'll see what's important to me, as well as anything new that I might be interested in.

b) as a system designer it's a terrible pollution of our database. It prohibits us from knowing if someone has read the content or not. That's vital information to us as user experience designers. We need to know the difference between those people who clicked mark as read and thought: "I'll come back to this later", versus those that thought "I don't want to know about this" versus those that have _actually read the discussion_. 

All of those user cases look the same to us in the database right now. that is really worrying to me, we should not be mashing those things together. 

Looking at this code I don't feel like it's the right thing to be doing.. especially the bit where we do group.discussions.select { mark as read! } that looks like it's going to have really long run times in the most popular groups. 

I think we should drop this feature. I'm really sorry but I feel strongly that we should not put effort into this direction any more.

This is not directed at anyone in particular, and I'm really sorry if this feels bad because you've been working on this. I don't want to do that at all.

I'm working on the user story for something to take the place of mark as read. I want to know if people would be ok with putting this on ice and focusing energy on helping design something better.

gdpelican:
Sounds goood to me.

PR #2862, opened 2016-01-07
Locales translation update

natiloomio:


robguthrie:
wowowowow! Thank you @natiloomio !!!

PR #2863, opened 2016-01-10
Plugin architecture

gdpelican:
Alright, ready for this to have some eyes on it. Keen to hear some ideas for plugins that folks have so I can be sure that I'm accommodating at least most of them with this.

Current capabilities include:

- Enabling / disabling plugins via ENV
- Adding new ruby classes
- Extending existing ruby classes
- Creating angular components
- Attaching directives to outlets in the angular interface
- Listening for arbitrary events (events are emitted whenever an Event is created, and every time a service method is called)
- Adding translations
- Adding new tables to the schema (I don't think I want to support modifying existing tables)

Big TODO is defining custom routes within the plugin, and how to reach them via the angular components in a way that isn't terrible. (edit: I was able to figure out how to do this with the existing system, wee!)

NB that this particular PR moves the webhook / slack functionality into a plugin.

I'd encourage interested parties to have a peek at [the README](https://github.com/loomio/loomio/blob/plugins-effyeah/plugins/README.md) to get an idea of how creating new plugins would work.


rdbartlett:
Wowowowowowow

Plugin idea: polling. Anyone can create up to 1 poll in a thread; anyone can add any number of options to the poll; anyone can vote positively for as many options as they like.

gdpelican:
PS I've also included a 'position specific' plugin which allows users to specify certain what certain positions on a proposal means, and has some really great examples of how to use the system.

screen shot 2016-01-11 at 7 53 46 am
screen shot 2016-01-11 at 7 54 51 am
screen shot 2016-01-11 at 7 56 15 am

gdpelican:
There's now a Loomio reactions plugin that works pretty well. That makes me very happy. :joy_cat:

toupeira:
This is very awesome, thanks for the hard work!

Some plugins I would love to see:
- different "comment items" on the left, like inline images/videos/audio, linking other discussions/proposals, generally linking all sorts of external sources
- different "action items" on the right as alternatives to proposals, like multiple-choice polls, date pickers, SWOT canvas
- trigger actions on events, like auto-accepting a proposal when it reaches a certain percentage of agreement. this should probably be also an item that can be added to individual discussions, and have a little box to communicate that it's there
- add more granular permissions, or implement different authorization models (also see discussion at https://www.loomio.org/d/Rg4rW70Y/permission-model-for-use-in-a-company)

The existing architecture already looks like it already provides for most (all?) of this and I'll definitely give it a try when it's merged. It will probably make sense to add high-level APIs for common plugin behaviour later, in order to avoid breakage with Loomio upgrades.

gdpelican:
Some plugins in the wild for reference:

https://github.com/loomio/loomio_webhooks
https://github.com/loomio/loomio_reactions
https://github.com/gdpelican/position_specific
https://github.com/gdpelican/peking_doge

angular/build/config/vendor.yml
```diff
@@ -0,0 +1,43 @@
+---
```
robguthrie: I am raising an eyebrow at the directory. I would have assumed that everything in build/ could be `rm -rf`'ed

robguthrie: Looks to me like build is really a config directory.

app/models/motion.rb
```diff
@@ -19,6 +19,7 @@ class Motion < ActiveRecord::Base
   has_many :did_not_voters, through: :did_not_votes, source: :user
   has_many :events,        -> { includes(:eventable) }, as: :eventable, dependent: :destroy
   has_many :motion_readers, dependent: :destroy
+  has_many :specifics, as: :specifiable
```
robguthrie: does emoji use this relation?

db/schema.rb
```diff
@@ -659,6 +659,15 @@
   add_index "organisation_visits", ["organisation_id"], name: "index_organisation_visits_on_organisation_id", using: :btree
   add_index "organisation_visits", ["visit_id", "organisation_id"], name: "index_organisation_visits_on_visit_id_and_organisation_id", unique: true, using: :btree
 
+  create_table "specifics", force: :cascade do |t|
```
robguthrie: I'm kind of wondering if specifics is necessary. It seems like something that people will try to use at first, then run into trouble because it's too limited. Also I fear that it's a source of pain when many plugins are using them and they clash/overwrite each other.

I wonder about not including specifics and just encouraging plugin authors to write their own models from the get go

app/models/motion.rb
gdpelican: No, specifics were put in to make reactions nice and easy.

I reckon there's a bunch of use cases that this covers though, and don't really want to encourage people to put tables into their instance because they'll have no easy way to remove them.

angular/build/config/vendor.yml
gdpelican: Open to moving the file, but we need it because we need a place to specify what files are included in vendor.js

There's also a .gitignored file in there called `plugins.yml` which is autogenerated and does the same thing for plugins (ie, include components from the plugins into app.js).

Once gulp compile is run, everything in /build could be rm -rfed and everything would work fine

db/schema.rb
gdpelican: Happy to take em out for this PR, but we'll need em for reactions, and ideally we'd move webhooks to working this way as well. I kinda want to discourage schema changes in plugins if we can help it.

robguthrie: I think it would be good to talk through why we want discourage schema
changes with plugins. Also I want to discuss downsides of a shared model..
maybe there are none but It kind of worries me. I'll see you tomorrow!

angular/build/config/vendor.yml
robguthrie: oh so this file is generated, not hand written?

gdpelican: Vendor is hand written; has to be, because we have no way of automatically
knowing what files from within the node_modules to include

gdpelican: plugins.yml (the gitignored file in the same folder) is auto generated
based on where you do

```
plugin.use_component
# or
plugin.use_asset
```

in any plugin files

robguthrie: So if we rm -rf angular/build then we lose stuff?  I'm suggesting that there is a convention around build being a directory of output and my expectations were that all of it's content would be generated.

Maybe we can put the config outside of build?

db/schema.rb
gdpelican: After having a ponder, I think I've landed on 'why take away options?' Tough to future imagine what the best solution will be, but it doesn't have to be one or the other, and I think there will definitely be cases where a shared model will be more effective than a database table, and vice versa.

Like, even the fact that we're respectfully clashing a bit over these two things is telling me 'aw, some authors will want to touch the schema, while others will prefer a more JSON-y approach, and that's cool.'

gdpelican:
I've taken out specifics and renamed the angular/build folder back to angular/tasks

I'll be good to merge this later today unless there are further thoughts @robguthrie @HSalmon

rdbartlett:
:tada:

PR #2865, opened 2016-01-11
translate subject at same time as email body

robguthrie:


robguthrie:
@gdpelican I'd appreciate a sign off if you have time

app/mailers/group_mailer.rb
```diff
@@ -12,24 +12,8 @@ def membership_request(admin, membership_request)
     @group = membership_request.group
     send_single_mail  to: admin.name_and_email,
                       reply_to: "#{@membership_request.name} <#{@membership_request.email}>",
-                      subject: t("email.membership_request.subject", who: @membership_request.name, which_group: @group.full_name),
+                      subject_key: "email.membership_request.subject",
+                      subject_params: {who: @membership_request.name, which_group: @group.full_name},
                       locale: locale_fallback(admin.locale)
   end
-
-  def group_email(group, sender, subject, message, recipient)
```
gdpelican: Looks like our only group email at the moment is in groups_controller#126, email_members.

Do we use this action or can we get rid of that too?

gdpelican:
Looks like a really nice improvement! :+1:

robguthrie:
Just need to check that invitation message still works as expected

robguthrie:
tested, and works well on clone

PR #2868, opened 2016-01-12
Emojis!

gdpelican:
http://g.recordit.co/fszjXFFSWr.gif


gdpelican:
Love to smoke test this tomorrow as well @HSalmon

gdpelican:
The more I look at this the less I see an easy win for email emojis with this as it currently works, as there's no ruby library that does what we need for emojione, and so frying this on the fly as we send the email out is tricky.

I'd like to get this in and circle back to it in the near future (if people complain about it, they probably will), as follows:

- User enters a post: `I love @saintnick! :santa:`
- We save the raw post to the database, but also have a library to do all of the post-processing we want on the server side (markdown, mentions, emoji, etc.)
 
  comment.raw: `I love @saintnick! :santa:` 
  comment.html: `I love <a href=/u/saintnick>@saintnick</a>! :santa:`

- This allows us to 

a) display the html in the user interface without doing on-the-fly markdown / mention / image rendering
b) still display the 'raw' comment for the user to edit (which will then get re-rendered)
c) opens a path for a live-updating composer... I imagine we'd store a rendered html version with the draft the user is creating, which means they can see what the html version would look like as they type = big win.

gdpelican:
edit: Holy crap, heh. I just looked up how Discourse does this (it's very similar to what's described above), except their server side renderer is running javascript code via `therubyracer`. Super crazy times!

```
  ctx = V8::Context.new(timeout: 5000)
  # emojis
  emoji = ERB.new(File.read("#{app_root}/app/assets/javascripts/discourse/lib/emoji/emoji.js.erb"))  
  ctx.eval(emoji.result)
```

robguthrie:
Looks great James.

I don't understand the reasoning for saving the rendered html, but it does not look like you're going to follow that up at this stage right?

gdpelican:
Yeah, not ready to go that far just yet.

I just pushed an experimental ExecJS change which might be kinda cool:
https://github.com/loomio/loomio/pull/2868/files#diff-24

It's just crazy enough to work?

robguthrie:
yeh. I reckon that's a great solution

gdpelican:
Okay, this is looking snappy. This is a big PR, but a lot of it is because we're adding emojis to so many places :blush: 

Would be cool to hear feedback on the ExecJS solution, as well as any caution around the changes to emails. Outline of those changes:

- We have a 'JSRuntime' class, which loads specified files from our vendor libraries (this happens in a 'gulp compile' task, and is a separate file from 'vendor' and 'app'. Idea is that we won't have our node_modules file around in prod, and we don't want to have to hold the entirety of vendor.js is memory just to perform emojione operations)
- There's an 'Emojifier' class, which executes `emojione.imageFromShortname` on email text going out.
- I've changed several of the emails so that they render_rich_text, and include a stylesheet which downsizes the email images (otherwise the images come through at 64x64px, when they should be 25x25). Roadie seems to handle this alright.

Issue #2692, opened 2015-11-05
Incorrect Privacy Notice on Start Thread Modal


alannallama:
When I go to raise a new thread in a private subgroup, it's telling me that it will be public. This is incorrect and misleading, and put me off submitting a thread with confidential information.

In the screenshot below, it says the thread will be public. But the group settings clearly show that it's private.

![screen shot 2015-11-05 at 11 48 30 am](https://cloud.githubusercontent.com/assets/5498878/10979808/712ebd7a-83b3-11e5-98d6-56bd149fb295.png)

![screen shot 2015-11-05 at 11 51 44 am](https://cloud.githubusercontent.com/assets/5498878/10979844/ad965dd6-83b3-11e5-9778-3adfa49eede6.png)

alannallama:
The above problem came up when I clicked "start thread" from the yellow "plus" button. When I went to the subgroup itself and clicked "start a new thread" it appeared to have the correct privacy message.

![screen shot 2015-11-05 at 11 56 34 am](https://cloud.githubusercontent.com/assets/5498878/10979942/4dc9a902-83b4-11e5-93b1-090b8630d91e.png)

robguthrie:
This is now fix. Thanks for the report @alannallama

PR #2870, opened 2016-01-12
Update privacy message on group change for thread form

gdpelican:
For #2692


gdpelican:
Looks like a really nice improvement! :+1:

PR #2876, opened 2016-01-15
Set title with unread notification count

gdpelican:
http://g.recordit.co/uuQjXI8hlD.gif

### DIFF
--- a/lineman/app/app.coffee
+++ b/lineman/app/app.coffee
@@ -52,7 +52,7 @@ angular.module('loomioApp', ['ngNewRouter',
     $compileProvider.debugInfoEnabled(false);
 
 # Finally the Application controller lives here.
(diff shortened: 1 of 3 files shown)


robguthrie:
James I'm not sure about this. 
I think I need a while longer to think about it.. but it feels weird to be changing the title of the page like this.

gdpelican:
My thoughts on why it's a good idea:


- My defacto browsing method (which I think is pretty common) is 'keep a tab open for each thing that I'm mildly interested in, check them when I see numbers'. I don't keep Loomio open in my everyday browsing because it doesn't give me this feedback.
- Almost every platform with notifications I know uses the title in this way (Facebook, Trello, Discourse, Twitter, any email client out there [I know at least yahoo and gmail]. Slack goes so far as to change the favicon based on tab activity)
- This is a really low-cost, really low-risk feature (it's definitely not going to break anything, and if it's only notifications I doubt it'll drive anyone away) which will make Loomio a lot stickier for certain users.

robguthrie:
James I hope you don't mind waiting a little bit longer on this one. Sorry to be drag!!!

robguthrie:
Talked with James and we've agreed to close this for now.

PR #2878, opened 2016-01-19
Instalike!

gdpelican:
Instantly update UI to show liked-ness:
http://g.recordit.co/e1VtMi370A.gif

If the request fails, revert to previous state:
http://g.recordit.co/PKUbx28BiF.gif

(Error message would be good, but happy with this for now.)


robguthrie:
marv! (ellous)

PR #2879, opened 2016-01-19
Move Intercom widget to avoid covering FAB

mhjb:
Can this be tested on an installation that has Intercom?

### Before

image

### After (I hope)

image

### DIFF
--- a/lineman/app/css/1_utilities.scss
+++ b/lineman/app/css/1_utilities.scss
@@ -262,3 +262,7 @@ textarea.form-control {
     opacity: 0;
   }
 }
+
+div#intercom-container div#intercom-launcher {
+    right: 85px;
+}


gdpelican:
Seems ok to me; any other objections?

mhjb:
did you have a chance to test it?

gdpelican:
Yep works fine for me.

Could be a less specific selector though, since ids are unique

```
#intercom-launcher {
  right: 85px;
}
```

mhjb:
I thought I needed a more specific selector than Intercom's own css uses

rdbartlett:
I don't understand why you'd not use an ID for this.

mhjb:
@rdbartlett both my version and @gdpelican use an ID — what are you saying?

rdbartlett:
I was confused. Ignore please :)

gdpelican:
Ah right. Well in any case this works for me and I'm merging it.

PR #2886, opened 2016-01-20
Ensure group exists for validating discussion creation

gdpelican:
http://errbit.loomio.org/apps/550f39ab5b229ba9b8000002/problems/569f198de305f76f5f000114


robguthrie:
oh, right.

thanks james

PR #2889, opened 2016-01-21
Send Spanish-speaking users to translated help manual

HSalmon:


lineman/app/components/group_page/group_help_card/group_help_card.coffee
```diff
@@ -6,5 +6,11 @@ angular.module('loomioApp').directive 'groupHelpCard', ->
   controller: ($scope, CurrentUser, AppConfig) ->
     $scope.showVideo = AppConfig.loadVideos
 
+    $scope.helpLink = ->
+      if _.contains(['es', 'an', 'ca', 'gl'], CurrentUser.locale)
```
gdpelican: This seems a little fragile and is a lot of logic to be duplicated. I'd love to see a UserHelpService or something like that, which handled this.

```
  $scope.helpLink = ->
    UserHelpService.helpLink()
```

Then, the help service would store a hash which would take a 

So in the service we'd have something like

```
helpLocale: ->
  switch CurrentUser.locale
  when 'es', 'an', 'ca', 'gl' then 'es'
  else                                      'en'
helpLink: ->
  "https://loomio.gitbooks.io/manual/content/#{@helpLocale()}/index.html"
```

gdpelican: I'm also uncomfortable with having hard coded loomio.org urls in here, but not enough to block this going in for now.

PR #2894, opened 2016-01-26
Fix word-break for CJK users

seungha-kim:
`word-break: keep-all;` is same as `word-break: normal;`, except for CJK words.
It will provide better ux for CJK users.

https://developer.mozilla.org/ko/docs/Web/CSS/word-break


seungha-kim:
Build failed: https://travis-ci.org/loomio/loomio/builds/104896594

PR #2910, opened 2016-02-03
Add companion PUT route to api PATCH actions

gdpelican:
Allow PUT action to be used in API where we're using PATCH

To tiptoe around this issue:
https://github.com/sandstorm-io/sandstorm/issues/1218

In reference to the Loomio / Sandstorm discussion happening here:
https://www.loomio.org/d/yw0D8Sc6/help-porting-loomio-to-sandstorm

NB: I've also updated the angular record store in 2.3.4 to add a `putMember` method, which will perform a `PUT` rather than `PATCH` request to the server (although I'd still like Loomio to be doing PATCH requests at the moment, since that's the railsy standard)
https://github.com/loomio/angular_record_store/pull/7

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -39,6 +39,7 @@ Loomio::Application.routes.draw do
     resources :groups, only: [:show, :create, :update] do
       get :subgroups, on: :member
       patch :archive, on: :member
+      put :archive, on: :member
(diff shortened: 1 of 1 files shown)


rdbartlett:
you are so cool

PR #2912, opened 2016-02-03
Add maximum height for attached images to avoid them dominating threads

mhjb:
Before | After
--- | ---
screen shot 2016-02-04 at 10 38 00 am | screen shot 2016-02-04 at 10 37 47 am

This has Product Owner approval :)


gdpelican:
Thanks!

PR #2915, opened 2016-02-04
Add revision history for discussion title and context

HSalmon:
screen shot 2016-02-04 at 3 42 05 pm

### DIFF
--- a/angular/core/models/version_records_interface.coffee
+++ b/angular/core/models/version_records_interface.coffee
@@ -1,3 +1,15 @@
 angular.module('loomioApp').factory 'VersionRecordsInterface', (BaseRecordsInterface, VersionModel) ->
   class VersionRecordsInterface extends BaseRecordsInterface
     model: VersionModel
+
+    fetchByDiscussion: (discussionKey, options = {}) ->
+      @fetch
+        params:
+          model: 'discussion'
+          discussion_id: discussionKey
+
+    fetchByComment: (commentId, options = {}) ->
+      @fetch
+        params:
+          model: 'comment'
+          comment_id: commentId
--- a/app/controllers/api/versions_controller.rb
+++ b/app/controllers/api/versions_controller.rb
@@ -1,0 +1,18 @@
+class API::VersionsController < API::RestfulController
+
+  private
+
+  def resource_class
+    PaperTrail::Version
+  end
+
+  def accessible_records
+    records = load_and_authorize(params[:model]).versions.order(created_at: :desc)
+    records = records.where(event: 'update') if params[:model] == 'discussion'
+    records
+  end
+
+  def default_page_size
+    100
+  end
+end
--- a/app/serializers/comment_serializer.rb
+++ b/app/serializers/comment_serializer.rb
@@ -1,6 +1,6 @@
 class CommentSerializer < ActiveModel::Serializer
   embed :ids, include: true
-  attributes :id, :body, :mentioned_usernames, :created_at, :updated_at, :parent_id, :parent_author_name
+  attributes :id, :body, :mentioned_usernames, :created_at, :updated_at, :parent_id, :parent_author_name, :versions_count
 
   has_one :author, serializer: UserSerializer, root: 'users'
   has_one :discussion, serializer: DiscussionSerializer
@@ -10,5 +10,4 @@ class CommentSerializer < ActiveModel::Serializer
   def parent_author_name
     object.parent.author_name if object.parent
   end
-
 end
--- a/app/serializers/discussion_serializer.rb
+++ b/app/serializers/discussion_serializer.rb
@@ -8,7 +8,8 @@ class DiscussionSerializer < Simple::DiscussionSerializer
              :updated_at,
              :archived_at,
              :comments_count,
-             :private
+             :private,
+             :versions_count
 
   attributes_from_reader :discussion_reader_id,
                          :discussion_reader_volume,
--- a/app/serializers/version_serializer.rb
+++ b/app/serializers/version_serializer.rb
@@ -1,16 +1,28 @@
 class VersionSerializer < ActiveModel::Serializer
   embed :ids, include: true
   attributes :id,
-             :changes
+             :changes,
+             :whodunnit,
+             :previous_id,
+             :created_at
 
   has_one :discussion
+  has_one :comment
   has_one :proposal
 
   def changes
     object.object_changes
   end
 
+  def whodunnit
+    object.whodunnit.to_i
+  end
+
   def discussion
+    object.item
+  end
+
+  def comment
     object.item
   end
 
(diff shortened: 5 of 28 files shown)


app/models/version.rb
```diff
@@ -0,0 +1,2 @@
+class Version < ActiveRecord::Base
```
gdpelican: Do we need this one? I think `PaperTrail::Version` has this covered.

app/serializers/discussion_serializer.rb
```diff
@@ -38,6 +39,10 @@ def reader
     @reader ||= scope[:reader_cache].get_for(object) if scope[:reader_cache]
   end
 
+  def edited
+    object.versions.where(event: 'update').count > 1
```
gdpelican: Ooh, we want to store this on the discussion once it's updated I reckon; this query is too expensive to call with every discussion serialized.

I think the easiest way to do this might be to put a counter cache on the number of versions, and check that

```
#discussion.rb
define_counter_cache :versions_count { |discussion| discussion.versions.where(event: :update).count }
```

```
# discussion_serializer.rb
def edited
  object.versions_count > 1
end
```

Something like that?

lineman/app/components/thread_page/revision_history_modal/revision_history_modal.haml
```diff
@@ -0,0 +1,18 @@
+.revision-history-modal
```
gdpelican: This will involve a little work to get into the new build system that's in master now; namely:

- Will need to copy component files from lineman/app to angular/core
- Need to update the haml, as the parser we're using has changed slightly. Biggest change is that all of the single quotes have to become double quotes for the parser to handle them correctly.

lineman/app/components/thread_page/thread_page.haml
```diff
@@ -37,6 +37,10 @@
           %timeago.nowrap{timestamp: '::threadPage.discussion.createdAt' }>
           %translate_button.lmo-card-minor-action{model: 'threadPage.discussion', showdot: 'true'}>
           %span{aria-hidden: 'true'}·
+          %span{ng-if: 'threadPage.threadEdited()'}
```
gdpelican: Could be `threadPage.discussion.edited` probly

lineman/app/models/version_model.coffee
```diff
@@ -13,3 +15,6 @@ angular.module('loomioApp').factory 'VersionModel', (BaseModel) ->
 
     attributeEdited: (name) ->
        _.include(_.keys(@changes), name)
+
+    authorOrEditor: ->
```
gdpelican: Could this be `authorOrEditorName`?

lineman/app/components/thread_page/revision_history_modal/revision_history_modal.coffee
```diff
@@ -0,0 +1,20 @@
...
+    
+    Records.versions.fetchByDiscussion($scope.discussion.key)
+
+    $scope.threadTitle = (version) ->
+      $scope.discussion.attributeForVersion('title', version)
+
+    $scope.threadDescription = (version) ->
+      $scope.discussion.attributeForVersion('description', version)
+
+    $scope.versionIsCurrent = (version) ->
```
gdpelican: This'll actually be off if we have a tonne of revisions, since we'll only load a single 'page' of them and end up with records that aren't loaded in the client but are on the server.

I'm tempted to raise the default page size of revisions to 100 or something and call it a good problem to have it we have it.

To raise the default page size:

```
# api_versions_controller
default_page_size
  100
end
```

app/controllers/api/versions_controller.rb
```diff
@@ -0,0 +1,8 @@
+class API::VersionsController < API::RestfulController
+
+  private
+
+  def accessible_records
+    load_and_authorize(params[:model]).versions.where(event: 'update')
```
gdpelican: Let's order this by `(created_at: :desc)` so we can ensure we always get the latest versions?

gdpelican:
Super fantastic work, @HSalmon. I've left some comments for the next iteration, and let's sit together sometime about transferring the lineman components over to the new build system.

app/serializers/discussion_serializer.rb
robguthrie: nice suggestion

app/models/version.rb
HSalmon: If I remove it I get a LoadError: "Unable to autoload constant Version". Thoughts?

app/serializers/discussion_serializer.rb
HSalmon: @gdpelican, having a bit of trouble with this, is it possible I need to implement something like the process outlined here: https://github.com/airblade/paper_trail#associations

lineman/app/components/thread_page/thread_page.haml
HSalmon: :+1:

lineman/app/models/version_model.coffee
HSalmon: :+1:

app/serializers/version_serializer.rb
```diff
@@ -1,27 +1,47 @@
...
   embed :ids, include: true
   attributes :id,
-             :changes
+             :changes,
+             :whodunnit,
+             :previous_id,
+             :created_at
 
   has_one :discussion
+  has_one :comment
```
gdpelican: Erg, we should really think about making this a polymorphic association. Not in this PR, but it's definitely technical debt we should pull out at some point.

```
# version.rb
belongs_to :versionable, polymorphic: true
```
```
#discussion.rb, comment.rb, proposal.rb
has_many :version, as: :versionable
```

gdpelican:
This diff looks fine to me. I reckon you could smack an e2e or two on here and we'd be good to merge.

HSalmon:
screen shot 2016-02-11 at 12 22 51 pm
screen shot 2016-02-11 at 12 44 59 pm

angular/core/models/version_records_interface.coffee
```diff
@@ -1,3 +1,15 @@
...
   class VersionRecordsInterface extends BaseRecordsInterface
     model: VersionModel
+
+    fetchByDiscussion: (discussionKey, options = {}) ->
+      @fetch
+        params:
+          model: 'discussion'
+          discussion_id: discussionKey
+
+    fetchByComment: (comment, options = {}) ->
```
gdpelican: Why pass the model on one and the key on the other? I think the convention is to pass id/key, rather than the whole object, but I'd have a poke around and go with one.

app/serializers/comment_serializer.rb
```diff
@@ -1,6 +1,6 @@
 class CommentSerializer < ActiveModel::Serializer
   embed :ids, include: true
-  attributes :id, :body, :mentioned_usernames, :created_at, :updated_at, :parent_id, :parent_author_name
+  attributes :id, :body, :mentioned_usernames, :created_at, :updated_at, :parent_id, :parent_author_name, :edited
```
gdpelican: In a complete change of opinion, I think we should serialize out the versions count, rather that just a boolean, which will enable this functionality down the road:

screen shot 2016-02-11 at 1 26 23 pm

(ie... it's been edited, but how many times!?)

gdpelican:
We'll need to formulate a plan about how to update the version counts for existing records... there's a _lot_ of them (~1mil unique edits between comments and discussions). Might take a little SQL-fu.

PR #2918, opened 2016-02-09
first batch of translations extracted into server.[locale].yml

robguthrie:
not quiite ready.


robguthrie:
Right.. I think I need to also remove the keys from en/locale.yml in this same commit

PR #2919, opened 2016-02-09
Extract translations

robguthrie:
This moves blocks of translation keys from en.yml to server.en.yml (and likewise for other locales)

Unfortunately the differences in YAML parsing cause this to be a bigger diff than necessary. I've tried some different YAML parsers and some other hacks but I cannot avoid this.

It's ready to go I reckon


robguthrie:
In this PR we move 

footer
start_group_form
pricing_page
password_reset
introduction_thread
how_it_works_thread
first_comment
first_proposal
contact_page
email
webhooks

into server.locale.yml from en.locale.yml this ends up removing them from en.yml.

The size of the PR is massive because of changes to the way the yaml is formatted when reparsing.

Look at the last commit to see more clearly what is happening.

@gdpelican I wonder if you could spare a couple of

gdpelican:
I had a peek, looks goood to me!

PR #2920, opened 2016-02-09
Pull chargify params out into service; use unique reference

gdpelican:
Currently testing on clone.

### DIFF
--- a/angular/core/services/chargify_service.coffee
+++ b/angular/core/services/chargify_service.coffee
@@ -1,0 +1,14 @@
+angular.module('loomioApp').factory 'ChargifyService', (CurrentUser) ->
+  new class ChargifyService
+
+    encodedParams: (group) ->
+      params =
+        first_name:   CurrentUser.firstName()
+        last_name:    CurrentUser.lastName()
+        email:        CurrentUser.email
+        organization: group.name
+        reference:    "#{group.key}|#{moment().unix()}"
+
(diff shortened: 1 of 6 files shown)


angular/core/services/chargify_service.coffee
```diff
@@ -0,0 +1,15 @@
+angular.module('loomioApp').factory 'ChargifyService', (CurrentUser) ->
+  new class ChargifyService
+
+    encodedParams: (group) ->
+      params =
+        first_name:   CurrentUser.firstName()
+        last_name:    CurrentUser.lastName()
+        email:        CurrentUser.email
+        organization: group.name
+        key:          group.key
```
gdpelican: NB: We'll want to change the callback url from 'http://loomio.org/g/{reference}' to 'http://loomio.org/g/{key}' on the public chargify page in prod; I've already done this for the test chargify account.

gdpelican:
This is ready for a look see and smoke, @robguthrie It's live on loomiotest.org at the moment.

robguthrie:
Saw an email comment about urls that used http rather than https. Just wanted to remind you that they should probably be httpS urls.

Have just successfully selected gift subscription and made donation. It was actually a really great experience.. (smoke test ok)

What is the term for developers getting so down on their own product??

gdpelican:
I put an s in our codebase, which made our product better, yay!

I dunno what that term is, 'dogfood anxiety'?

PR #2939, opened 2016-02-13
Add emojis to discussion / motion titles

gdpelican:


gdpelican:
I am a little torn about this one. Rendering images in titles means we need to render html in titles. This means we allow things like `<div>`, `<li>`, `<h2>`, whatever to be rendered in the title. A quick look into 'allow img html to be rendered but not other html tags' resulted in head pain (and we'd still allow people to put whatever image they like in there...)

It's sanitized html, so it's not dangerous (ie, no script tags, or styles, or errant angular directives, or anything that could be harmful), but it does still feel like a loophole that I'm not really sure I want to introduce.

screen shot 2016-02-13 at 5 14 09 pm
screen shot 2016-02-13 at 5 14 17 pm

screen shot 2016-02-13 at 5 16 19 pm

rdbartlett:
Haha, yeah... that seems a bit bad eh. Seems like arbitrary images in titles could be a major UI pain.

mhjb:
Just did a quick search — there's not a single like '%<div%' discussion title in the db presently. Are we expecting a rash of them for some reason?

gdpelican:
No one's had a reason to up to now, since we don't render html. For example, that doge appears that way (at that size) in the inbox, dashboard, thread page, group page... would do in the explore page, sidebar, everywhere. And for proposal names as well.

It's a catastrophically bad experience for a bunch of people in a lot of ways just because one person decides to be cute / experimental / trolly; I'm hesitant about introducing that possibility. We can do stuff to mitigate it (lessen the size of images, strip out certain tags, etc.), but those are the kind of thing that add complexity fast and those solutions are like plugging holes in a sinking boat, folks will always be able to find a way around them if they want to.

The other option to explore here is to take a look at Markdown titles (which is how we render the emojis in other places), but that's it's own (somewhat smaller?) can of worms.

Would be good to hear @robguthrie 's take as well.

robguthrie:
Sorry just caught up. I don't think it's worth doing right now. People can
paste unicode emojis if they want to go this far.. but it seems like a lot
more rendering overhead and also potentially really DOS'y to enable this.
I'd vote no thanks for this.

Issue #2829, opened 2015-12-17
Explain why someone should use the shareable group link


rdbartlett:
I'm keen to a) give the shareable group link a name, so we can refer to it easily in the help docs and b) promote its use (because I think it is better than sending unique invitations). 

@natiloomio and I have just tried some new language. What do you think of this?

![invite](https://cloud.githubusercontent.com/assets/970124/11858586/c4d17d2a-a4c7-11e5-8a8b-4357c77db445.png)

Also note the Group select is hidden when we already know what group they're on.

rdbartlett:
Next iteration:

![invite](https://cloud.githubusercontent.com/assets/970124/11883693/840f4b48-a579-11e5-9915-3ba9322d4040.png)

Next, I want to design the 'turn off link sharing' feature.

rdbartlett:
(Oops deleted previous comment w wrong image)

Next iteration:

![invite](https://cloud.githubusercontent.com/assets/970124/11883723/a3c6aa8a-a579-11e5-925f-bc8d88dc9269.png)

robguthrie:
I love it

rdbartlett:
The button says 'done' until you enter an email address, when it changes to say 'send'.

Think it is a good enough next iteration?

robguthrie:
i do

mhjb:
I think it is an improvement and a good enough iteration.

But:

The paragraph of text is quite long. I wondered about breaking it into two paras, so there's more of an association of the first sentence with the Shareable link, and the second with Email addresses. Perhaps adding a divider between the two sentences.

It's not super clear that the custom message only applies to the Email addresses option. 

Any interest in a tiny bit of hall user-testing?

HSalmon:
I wonder if it's worth saying a little more about what a shareable link is, e.g:

"You can copy the link and send it to people you wish to join your group. Clicking the link will automatically provide them access."

Just thinking about Teddy's confusion around what a shareable link is...

robguthrie:
that's great text Hannah

rdbartlett:
Thanks everyone. Here's the next iteration:

![invite](https://cloud.githubusercontent.com/assets/970124/12991409/fe16b914-d173-11e5-9fcb-513c24852315.png)

I hallway tested it with Teddy and Russell. I feel confident that this is an improvement.

HSalmon:
Cool! I could even have a stab at building it today if it's approved

rdbartlett:
Hallway test with Gina:

[...]

PR #2944, opened 2016-02-14
Update invitations form

HSalmon:
PR for [this issue](https://github.com/loomio/loomio/issues/2829#issuecomment-183081763)


HSalmon:
screen shot 2016-02-15 at 11 12 11 am

HSalmon:
@robguthrie, @gdpelican, @rdbartlett - does someone wanna take a look at this?

angular/core/components/invitation_form/invitation_form.coffee
```diff
@@ -26,12 +28,25 @@ angular.module('loomioApp').factory 'InvitationForm', ->
...
 
+    $scope.invitation = ->
+      group.shareableInvitation()
+
+    $scope.copied = ->
+      FlashService.success('common.copied')
+
+    $scope.invitationLink = ->
+      return unless group.shareableInvitation()
+      [AppConfig.baseUrl, 'invitations/', group.shareableInvitation().token].join('')
```
gdpelican: I'm a little nervous about hard coding the 'invitations' path into this. Can we instead have the shareable invitation know what url it's supposed to be spitting out?

```
group.shareableInvitation().url
```

```
# invitation_serializer
def url
  invitations_url(group, token) # => or something that spits out http://loomio.org/g/key/invitations/token
end
```
or something like this?

gdpelican:
I know we stumbled across some thorny experiences around email validation, but I think this iteration can ignore them for now (since they're pre-existing), and we'll take care of that in a second pass.

angular/core/components/invitation_form/invitation_form.coffee
gdpelican: This would also eliminate the AppConfig dependency here.

gdpelican:
I'm not a fan of 'done' here; I think 'send' is still right, after all, we only want people clicking that button to send invites; they can click 'cancel' or close the modal if they've copied the link.

I also think that disabling that send/done button unless there are email addresses in the email input is still the right behaviour here. I think there's some missing design (market research?) here about how to handle validation. My guess:

- We track the number of valid and invalid email addresses entered in the box (string delimited by ',' or ';')
- We validate client side that there's at least 1 valid email address, otherwise you can't hit send.
- If there's at least one invalid email address, a caution message appears saying that there are some invalid emails.
- Clicking 'send' while there are both valid and invalid email addresses sends the valid ones while discarding the invalid ones.
- If at least one email gets sent, the modal closes and we toast how many were sent (3 invitations sent.)
- If no emails get sent for whatever reason, a validation error shows up in the modal 'We were unable to send any of those invites. Please try again', and it does not close.

robguthrie:
I reckon port the email regex I wrote from Ruby to JS. It's just looks for something@something.something and works pretty well in practice. Then you don't need to specify delimeter to detect valid addresses, it'll just pull out all valid addresses for you.

HSalmon:
@gdpelican, @robguthrie, any objections to moving this forward?

robguthrie:
nope!

angular/core/components/invitation_form/invitation_form.coffee
```diff
@@ -21,17 +23,33 @@ angular.module('loomioApp').factory 'InvitationForm', ->
...
+      else
+        'invitation_form.done'
+
+    $scope.invalidEmail = ->
+      $scope.invitees().length == 0 and $scope.form.emails.length > 0
 
     $scope.canSubmit = ->
       $scope.invitees().length > 0 and $scope.form.group()
 
+    $scope.invitation = ->
```
gdpelican: Do we use this method?

gdpelican:
Have you tested changing the group dropdown value and making sure it picks up the new group correctly, @HSalmon ?

angular/core/components/invitation_form/invitation_form.coffee
```diff
@@ -21,17 +23,33 @@ angular.module('loomioApp').factory 'InvitationForm', ->
...
+      group.shareableInvitation()
+
+    $scope.copied = ->
+      FlashService.success('common.copied')
+
+    $scope.invitationLink = ->
+      return unless group.shareableInvitation()
```
gdpelican: I think we need to be updating this based on $scope.form.group(), since `group` is just the group passed into the form initially

angular/core/components/invitation_form/invitation_form.coffee
```diff
@@ -1,6 +1,8 @@
 angular.module('loomioApp').factory 'InvitationForm', ->
   templateUrl: 'generated/components/invitation_form/invitation_form.html'
-  controller: ($scope, group, Records, CurrentUser, AbilityService, FormService, FlashService, RestfulClient, ModalService, TeamLinkModal, AddMembersModal) ->
+  controller: ($scope, group, Records, CurrentUser, AbilityService, FormService, FlashService, RestfulClient, ModalService, AddMembersModal) ->
+
+    Records.invitations.fetchShareableInvitationByGroupId(group.id)
```
gdpelican: And we'll also need to do this when the group changes, right?

HSalmon:
@gdpelican, I think this is ready to go!

angular/core/components/invitation_form/invitation_form.coffee
```diff
@@ -1,12 +1,15 @@
...
   templateUrl: 'generated/components/invitation_form/invitation_form.html'
-  controller: ($scope, group, Records, CurrentUser, AbilityService, FormService, FlashService, RestfulClient, ModalService, TeamLinkModal, AddMembersModal) ->
+  controller: ($scope, group, Records, CurrentUser, AbilityService, FormService, FlashService, RestfulClient, ModalService, AddMembersModal) ->
 
     $scope.availableGroups = ->
       _.filter CurrentUser.groups(), (group) ->
         AbilityService.canAddMembers(group)
 
     $scope.form = Records.invitationForms.build(groupId: group.id or (_.first($scope.availableGroups()) or {}).id)
+    $scope.getShareableInvitation = ->
```
gdpelican: let's call this 'fetchShareableInvitation' ('fetch' being the codeword for 'make a web request', and get being 'pull from the local record store')

angular/core/components/invitation_form/invitation_form.coffee
```diff
@@ -21,17 +24,30 @@ angular.module('loomioApp').factory 'InvitationForm', ->
...
+      $scope.invitees().length == 0 and $scope.form.emails.length > 0
 
     $scope.canSubmit = ->
       $scope.invitees().length > 0 and $scope.form.group()
 
+    $scope.copied = ->
+      FlashService.success('common.copied')
+
+    $scope.invitationLink = ->
+      return unless $scope.form.group().shareableInvitation()
```
gdpelican: I think we have to make sure that $scope.form.group() is present here as well... or otherwise at least check out what the edge case of 'I'm logged in but can't actually invite someone to any of my groups' looks like

gdpelican:
Looks great Hannah, one small quibble and should be good to go.

PR #2945, opened 2016-02-14
Render Markdown in print version

toupeira:
I couldn't find a spec for this so I hope you'll forgive me if I don't add one :)

Also Emojis don't work yet but it looks like [you're already aware of that](https://github.com/loomio/loomio/commit/dac61c8e58dee0454e6b2f9f06e4b0c9a3a6ea67).

### DIFF
--- a/app/views/discussions/print.html.haml
+++ b/app/views/discussions/print.html.haml
@@ -9,7 +9,7 @@
         %h3= @group.name
   .row.context
     .col-md-9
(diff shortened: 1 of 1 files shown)


gdpelican:
Yep, emoji email rendering is being taken care of in this branch:
https://github.com/loomio/loomio/pull/2942

Thanks for the contribution!

Issue #2938, opened 2016-02-12
Subgroups show up in Intercom


mhjb:

toupeira:
I just wanted to report a bug that sounds related (subgroups showing up when searching for them on /explore), it looks like `Group#group_privacy=` isn't called when needed:

```ruby
group = Group.new(parent: Group.first, group_privacy: 'closed')
group.visible_to # -> "public"
group.valid? # -> false
group.visible_to # -> "public"
```

And yeah this behaviour is part of my confusion about how permissions work, so I'm glad it's just a bug :grin:

I noticed a similar bug as well when changing the privacy from "Closed" to "Secret" and having `parent_members_can_see_discussions` checked, it would only let me save the form if I uncheck it first before switching to "Secret". All these permissions should probably be set and/or verified in a `before_validation` hook.

toupeira:
Found another bug: when creating a sub-group the "Open" setting is visible, but when editing the sub-group it's not. I guess it should be always unavailable for sub-groups where the parent isn't Open?

gdpelican:
The visible 'open' setting is a known bug related to this one: #2888 I think @robguthrie may have his eyes on this one here: #2936

toupeira:
@mhjb I'm sorry for hijacking your issue, didn't see all the other related issues and will search first in the future ;-)

The UI fixes from #2899 seems to have resolved most of the bugs I encountered and I added #2952 for the checkbox bug I described above.

rdbartlett:
@robguthrie has got a fix for this issue coming. Thanks @toupeira for your new issue. Let's chat soon!

PR #2953, opened 2016-02-15
dont send subgroups to intercom

robguthrie:


robguthrie:
@gdpelican can you just check this JS looks like it does what I think it does and merge.

robguthrie:
fixes #2938

Issue #2837, opened 2015-12-17
Memberships page is only showing first 50 pending invitations


rdbartlett:
Currently in the Enspiral group I can see 50 pending invitations in angular and 52 in rails. In angular, the most recent 2 invitations are missing. I'm assuming that is a pagination issue. As a quick fix can we just up it to 500, and reverse it so I see the most recent invitations?

rdbartlett:
Or load the next 50 when you scroll to the bottom of the page?

rdbartlett:
@HSalmon wanna bump the max on this to 1000?

PR #2957, opened 2016-02-16
Allow display of up to 1000 pending invitations

HSalmon:
To fix: https://github.com/loomio/loomio/issues/2837


HSalmon:
Is this implementation correct @robguthrie?

gdpelican:
Yep that's right.

HSalmon:
Rad, cheers @gdpelican

PR #2967, opened 2016-02-17
Fix vertical mobile card headers; broadcast placeholders further

### LINKED ISSUE #2779: Card headings funky on at least iPhone 4, some Android

mhjb:
Resolves #2779 

Before | After
----- | -----
 | 

This is an intermittent bug. Worth testing on clone to be sure I've squashed it and there are no side effects.

### DIFF
--- a/angular/core/components/group_page/discussions_card/discussions_card.coffee
+++ b/angular/core/components/group_page/discussions_card/discussions_card.coffee
@@ -29,7 +29,7 @@ angular.module('loomioApp').directive 'discussionsCard', ->
                         discussion: -> Records.discussions.build(groupId: $scope.group.id)
 
     $scope.showThreadsPlaceholder = ->
-      AbilityService.canAdministerGroup($scope.group) and $scope.group.discussions().length < 3
+      $scope.group.discussions().length < 4
 
     $scope.whyImEmpty = ->
       if !AbilityService.canViewGroup($scope.group)
--- a/angular/core/components/group_page/discussions_card/discussions_card.scss
+++ b/angular/core/components/group_page/discussions_card/discussions_card.scss
@@ -3,7 +3,12 @@
 }
 
 .discussions-card__header{
-  margin: $cardPaddingSize;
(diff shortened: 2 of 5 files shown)


gdpelican:
This fix breaks looks like it breaks some stuff :scream_cat:, first thing I noticed was the thread page:

screen shot 2016-02-17 at 7 03 28 pm

but lmo-card-heading gets used in a lot of places, so wouldn't surprise me if there were other places where we need to account for this change somehow

mhjb:
Thanks James — I'll put some more effort in :)

mhjb:
* get gulp to watch core/css/*.scss too
* showed group description placeholder and thread placeholder to non-coordinators
* showed group description placeholder when there are <4 (rather than <3) discussions — so there's room for the two auto-generated ones plus one more
* rewrote group description placeholder to make sense to non-coordinators
* used flex-box for discussions-card__header & children (this seems to work but looks a teensy bit goofy and may want some more effort)

mhjb:
Here's where I've gotten it to. Margins less beautiful. But I'm wondering whether the flexboxy effort of perfecting them is worth doing this side of Material. 

iphone 4 | chrome on osx
---- | -----
 | image

rdbartlett:
This iteration is a major improvement.

robguthrie:
looks fine to me!

angular/tasks/paths.coffee
```diff
@@ -4,7 +4,7 @@ include = (key, file) -> _.map(config.vendor[key], (file) -> [config.vendor.path
 
 module.exports = {
   css: {
-    core:           _.flatten([include('css'), 'core/css/main.scss', 'core/components/**/*.scss']),
+    core:           _.flatten([include('css'), 'core/css/*.scss', 'core/components/**/*.scss']),
```
gdpelican: Hrm. This one's a bit tricky. We don't want to actually build these extra files, because they all get imported into main.scss (in a certain order), so we double up our css length this way. However, it would be nice if we watched for changes in them so that they get picked up.

Would you be willing to take this out of the PR (feel free to use it locally), and I'll take a look at this sometime next week?

PR #2970, opened 2016-02-17
clean up eventbus stuff, syndicate mark as read

robguthrie:
not quite ready but it works.

### DIFF
--- a/app/models/discussion_reader.rb
+++ b/app/models/discussion_reader.rb
@@ -97,12 +97,7 @@ class DiscussionReader < ActiveRecord::Base
       reset_counts!
     end
 
-    publish!
-  end
-
-  def publish!
-    publish_data = Simple::DiscussionSerializer.new(discussion, scope: { reader_cache: reader_cache })
-    MessageChannelService.publish(publish_data.as_json, to: user)
+    EventBus.broadcast('discussion_reader_viewed!', discussion, user)
   end
 
   def reset_comment_counts
@@ -164,9 +159,5 @@ class DiscussionReader < ActiveRecord::Base
   private
   def membership
     discussion.group.membership_for(user)
-  end
-
-  def reader_cache
-    @reader_cache ||= DiscussionReaderCache.new(user: self.user, discussions: Array(self.discussion))
   end
 end
--- a/config/initializers/event_bus.rb
+++ b/config/initializers/event_bus.rb
@@ -21,6 +21,7 @@ EventBus.configure do |config|
   end
 
   # send individual emails after thread events
+  # Single mail kinds is only Comment replied to and User mentioned.
   Event::SINGLE_MAIL_KINDS.each do |kind|
     config.listen("#{kind}_event") do |event, user|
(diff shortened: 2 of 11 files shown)


app/models/discussion_reader.rb
```diff
@@ -97,12 +97,7 @@ def viewed!(age_of_last_read_item = nil)
       reset_counts!
     end
 
-    publish!
-  end
-
-  def publish!
-    publish_data = Simple::DiscussionSerializer.new(discussion, scope: { reader_cache: reader_cache })
-    MessageChannelService.publish(publish_data.as_json, to: user)
+    EventBus.broadcast('discussion_reader_viewed', discussion, self, user)
```
gdpelican: So far the EventBus hasn't needed to dip down to the model level (it's called entirely from services). Is there a reason we can't call this from DiscussionService#mark_as_read?

robguthrie: I was trying to keep the broadcast super close to where the work is done.

Right now there is problem that when you create a discussion, other tabs don't see that it is read.

I don't think we're going to call DiscussionService#mark_as_read from DiscussionService#create are we?

Seems like the tidiest thing to do is fire an event whenever a discussion is viewed (ie: discussion_reader.viewed!).

There is heaps more I want to talk to you about today now I've had experience with this.. Would be great to pair again.

gdpelican: Nah, but we can listen for both 'discussion_create' and
'discussion_mark_as_read' in the EventBus listener

```
EventBus.listen('discussion_create', 'discussion_mark_as_read') do
|discussion, actor|
  reader = get_the_reader(discussion, actor)
  serialize_the_thing(reader)
end
```

robguthrie:
Those two have different signatures when calling EventBus.broadcast.

DiscussionService.mark_as_read broadcasts: discussion, params, actor
DiscussionService.create broadcasts: discussion, actor


Hrrmm...

robguthrie:
Also looking at stuff like Events::NewComment.. I can see that we do broadcast from models already..

gdpelican:
Yarp, I tripped over that before, you can see that the discussion_update actually does 'discussion, actor, params', so that I can pair it with 'discussion_create' in the event bus

Which kinda makes me think the convention might want to be |discussion, actor, params|, since sometimes params exists and sometimes it doesn't.

robguthrie:
Yea... I'm tipping more towards not using position based args and passing a hash so you can just get args by name..

config/initializers/event_bus.rb
```diff
@@ -54,6 +72,14 @@
     DiscussionReader.for_model(model, actor).set_volume_as_required!
   end
 
+  config.listen('discussion_reader_viewed!') do |discussion, discussion_reader, actor|
```
gdpelican: Do you think we could take the `discussion_reader` param out of this, since we don't use it, and just do `|discussion, actor|`?

robguthrie:
whoop hwoop!

PR #2972, opened 2016-02-17
rip out some of the stats.. probably a little premature

robguthrie:
@mhjb and @rdbartlett I think this is a little early, but I guess I'm giving it a poke.. Can we remove the admin/stats and aaarrrr table stuff from the app now that we use metabase?

Matthew, do you use (or think you might in the future) the group_measurements table?


rdbartlett:
I think it is probably fine to remove. Some of the definitions might be useful - but we can always look back to this commit if we want to reference them again in future.

mhjb:
Probably fine.

Can you tell me a bit about the group_measurements table?

robguthrie:
@mhjb group measurements table records counts of things each day. (think discussions count per day per group) so that you can look back and see change over time which might otherwise be lost information.

mhjb:
Might be useful for the retention chart, which I've been asked to add to
the AAAR dashboard.

robguthrie:
cool we can keep em around and rip out the other stuff

gdpelican:
How do we feel about this at the moment @robguthrie @rdbartlett @mhjb ? I think I want to tear this out since it's reeeally slow and I don't think we're actively using it.

rdbartlett:
I'm happy for it go as soon as Matt gives it the all clear.

mhjb:
all clear!

app/models/cohort.rb
```diff
@@ -1,41 +1,7 @@
...
-               .where("properties ->> 'url'= 'https://www.loomio.org/'")
-               .where(time: start_on..(end_on + 1.day))
-  end
-
   def organisations
     groups.where('parent_id IS NULL')
   end
-
-  def activated_organisations
-    # organisation_visits_count
```
gdpelican: @HSalmon would you mind taking this one? #3045 

I think you just have to put this method back in.

PR #2973, opened 2016-02-18
Notify inviter when invitation is accepted

HSalmon:
screen shot 2016-02-18 at 1 04 06 pm

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -1,4 +1,20 @@
 class Notification < ActiveRecord::Base
+
+  BETA_NOTIFICATIONS = %w[comment_liked
+                          comment_replied_to
+                          membership_request_approved
+                          membership_requested
+                          motion_closed
+                          motion_closed_by_user
+                          motion_closing_soon
+                          motion_outcome_created
+                          new_comment
+                          new_discussion
+                          new_motion
+                          new_vote
+                          user_added_to_group
+                          user_mentioned]
+
   belongs_to :user
   belongs_to :event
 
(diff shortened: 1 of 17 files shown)


rdbartlett:
:heart_eyes_cat:

robguthrie:
in the string "accepted your invitation to join Loomio" is Loomio the group
name? or is it just the string Loomio? I think it should be the group name.

mdshearer:
Yee har!! And yeah, group name would be swell.

getresults:
Having clickable URLs to both the person's loomio profile and the specific loomio group would be awesome if you haven't included this already. :)

HSalmon:
User name links to user profile, group name links to group:

screen shot 2016-02-18 at 2 07 44 pm

HSalmon:
@gdpelican, @robguthrie, would love yr eyes when you've got a moment or two?

spec/models/events/invitation_accepted_spec.rb
```diff
@@ -1,16 +0,0 @@
-require 'rails_helper'
```
gdpelican: Did these tests fail?

app/models/notification.rb
```diff
@@ -10,9 +26,9 @@ class Notification < ActiveRecord::Base
   delegate :eventable, :to => :event
 
   scope :unviewed, -> { where(viewed: false) }
+  scope :beta_notifications, -> { joins(:event).where(:events => {:kind => BETA_NOTIFICATIONS}) }
```
gdpelican: Brilliant! I think we want to stay away from the hash rocket though (as a matter of style)

```
joins(:event).where(events: { kind: BETA_NOTIFICATIONS })
```

gdpelican:
This looks pretty fantastic to me. An e2e would be fantastic. Here's my thought:

- Send invitation
- visit `development/accept_last_invitation`, which accepts the invite on behalf of the invitee, like this:
```
def accept_last_invitation
  InvitationService.redeem(Invitation.last, Invitation.last.invitee)
  redirect_to(test_group)
end
```
- Click notifications dropdown, I should have a notification about it
- Click the notification, I should be on the memberships page
- On that page, I should see the invited person's name or handle or something

HSalmon:
I'm having a lot of trouble implementing your suggestions for an e2e @gdpelican, would you be up for pairing on it?

PR #2976, opened 2016-02-18
Static landing page (WIP)

gdpelican:
I wonder if this works!

The hope to is get our time-to-first-landing down some, as well as provide an entry point for a potential mobile application.

I think this is also a slightly nicer way of managing our third party scripts.

cc @robguthrie @gardner


app/views/angular/boot.js.erb
```diff
@@ -0,0 +1,21 @@
+var includeJs = function(url, callback) {
```
gdpelican: All I have at this point is vanilla javascript and darn does it feel good. #stoneage

app/views/angular/boot.js.erb
```diff
@@ -0,0 +1,21 @@
...
+}
+
+var includeCss = function(url) {
+  var elem   = document.createElement('link')
+  elem.href  = url
+  elem.media = 'screen'
+  elem.rel   = 'stylesheet'
+  document.getElementById('boot').appendChild(elem)
+}
+
```
gdpelican: I like having these all bunched together like this into a single boot file; makes it really obvious what scripts show up and why

app/views/angular/boot/_piwik.js
```diff
@@ -0,0 +1,13 @@
...
+(function() {
+  var u="//piwik.loomio.org/piwik/";
+  _paq.push(['setTrackerUrl', u+'piwik.php']);
+  _paq.push(['setSiteId', 1]);
+  var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+  g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
+})();
+
+// ???
+// <noscript><p><img src="//piwik.loomio.org/piwik/piwik.php?idsite=1" style="border:0;" alt="" /></p></noscript>
```
gdpelican: Oh right, have to do something about this. What does this do again? Seems mighty fishy to me.

app/views/angular/index.html
```diff
@@ -0,0 +1,20 @@
+<html ng-app="loomioApp" ng-controller="ApplicationController as app">
```
gdpelican: Beautiful, glorious, regular html, just like mom used to make.

(PS this is the mobile app entry point.)

app/views/angular/boot/_piwik.js
gdpelican: Oh right tracking pixel.

gardner:
This is rad! I'll try to wrap it up in cordova now.

gdpelican:
I think I'm going to try to push this over the line, mostly because I like the idea of grouping our third party dependency scripts together in a really obvious way.

gardner:
Push this pull request over the line?

gdpelican:
Yep, I like the idea of a static page to increase our first-byte time, and then we can do another spike into a cordova thing at some point if we want to.

robguthrie:
Actually.. I'm a little concerned about going "All static" I think we need
to consider a few more things. A dynamic page is only slow if it's doing
lots of work. There are some cases where we do want to some light work and
put it into the boot page, and the time will still be acceptable. My first
thought is facebook/slack previews, do we need to insert meta tags into
this response so they work for discussion, group pages?

gdpelican:
That'd be a thing worth testing.

PR #2980, opened 2016-02-20
Add Loomio OAuth signin

gdpelican:
In favor of #2800 

I set up a really simple sample app here: https://github.com/loomio/loomio-oauth-sample


Connoropolous:
Niiiice

gdpelican:
Upon further thought, I think I'm going to pull this out into a plugin. ¯\_(ツ)_/¯

rdbartlett:
`james:plugin::rich:blog`

gdpelican:
Yeah. The more plugins we have out there, the easier it is for people to write them, and the more functional they get.

(For example, writing this as a plugin will mean we have to support adding gemfile dependencies (ie doorkeeper) within plugins. That's a useful piece of work.)

gdpelican:
On second thought, this is too much of a PITA to pull out into plugin land at the moment, but I'm glad I gave it a whirl because it gave me a few TODOs in the plugin arch which would be useful to add at some point.

- Plugin gem dependencies
- Adding indices to custom tables
- Adding more arbitrary code to `routes.rb` (not just individual routes)
- Adding code which runs with the app initializers

Connoropolous:
@gdpelican good call on adding it in to the angular app. Any sense how much more work its going to take, and when it might be making it to the live website? do you loomians always work in PRs and do code reviews, or something else?

gdpelican:
It's about another day of work (rather than day of time) away from CR-ready.

The doorkeeper UI out of the box is pretty horrid, so I'm keen to get
something nicer together.

I'll want to wait until @robguthrie and I have a time to sit down together
and go through it, since it's of course a security concern.

Suffice it to say it's a goal for the 2-week sprint which started today.
:balloon:

Connoropolous:
Awesome James :) 

Given one scenario in which Loomio happens to be the platform where all Enspiralites are guaranteed to have an account, one really exciting bonus of this work would be the ability to sign in to other platforms/tools used by the Enspiral network with Loomio. Signing in with Loomio, and having the right API endpoints, can automatically enable that other tool to validate that Loomio user X is a member of Loomio group XYZ(Enspiral) and be permitted usage of it, vastly simplifying user accounts for Enspiral members. Towards this end, a great follow-up step will be (and I may look into doing this) developing a 'omniauth-loomio' gem, alike to https://rubygems.org/gems/omniauth-twitter/versions/1.2.1 which eases client app development. Also be worthwhile writing an oauth-loomio node package (I've been experimenting with something similar) for node apps.

gdpelican:
Documenting how an OAuth flow works is here. But gifs can help!

**Step 1**: Create an application (as an application owner)
tqt8u7bx1r



**Step 2**: Obtain an access token (as a person using a third party app)
(I'm using the sample app [here](https://github.com/loomio/loomio-oauth-sample) for this)
lqkd8ru8kc



**Step 3**: Use the access token to query the Loomio database (as a person using a third party app)
q2rfurb9i2

Connoropolous:
just ran through the whole thing and love it :) using the client app and all. super easy. you're a star @gdpelican 

this will make it easier for me to test out the API as well, and potentially help on the documentation front.

gdpelican:
**Step 4**: If you think an app is misbehaving, or you don't use it anymore, you can visit /apps/authorized` to revoke access
d1qxwx0eit

Connoropolous:
so @gdpelican are apps registered on a user by user basis? no one can remove any one elses apps, corrent?

gdpelican:
Yep, there are two screens:

`/apps/registered` are apps that you've created yourself. You can remove, edit the name, change redirect uris, and fetch the app keys (uid and secret)

`/apps/authorized` is actually referring to access tokens (IE, these are the apps that exist which you have issued an access token to. They can do stuff as you.) The thing that you're revoking is actually your access token for that app; you're not touching the registered application at all (and you have no access to other people's registered apps, or other people's access tokens)

Connoropolous:
thanks :) final question. will the access token provided via a registered
app still work if that app gets deactivated by the user (rather than the
token revoked)?

On Tue, Feb 23, 2016 at 10:18 AM, James Kiesel <notifications@github.com>
wrote:

> Yep, there are two screens:
>
> /apps/registered are apps that you've created yourself. You can remove,
> edit the name, change redirect uris, and fetch the app keys (uid and secret)
>
> /apps/authorized is actually referring to access tokens (IE, these are
> the apps that exist which you have issued an access token to. They can do
> stuff as you.) The thing that you're revoking is actually your access token
> for that app; you're not touching the registered application at all (and
> you have no access to other people's registered apps, or other people's
> access tokens)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/loomio/loomio/pull/2980#issuecomment-187387484>.
>



-- 
Connor Turland
~to craft a sustainable and thriving culture within and without myself~
co-creator of Metamaps.cc

gdpelican:
I don't think so, but I haven't tried it yet.

robguthrie:
just wow

config/initializers/doorkeeper.rb
```diff
@@ -0,0 +1,109 @@
...
+  #   # Example implementation:
+  #   Admin.find_by_id(session[:admin_id]) || redirect_to(new_admin_session_url)
+  # end
+
+  # Authorization Code expiration time (default 10 minutes).
+  # authorization_code_expires_in 10.minutes
+
+  # Access token expiration time (default 2 hours).
+  # If you want to disable expiration, set this to nil.
+  access_token_expires_in nil
```
robguthrie: lets start with the default

config/locales/doorkeeper.en.yml
```diff
@@ -0,0 +1,54 @@
+en:
```
robguthrie: add doorkeeper-i18n gem?

robguthrie: and remove this file?

config/routes.rb
```diff
@@ -464,4 +472,9 @@
   get '/community'  => redirect('https://www.loomio.org/g/WmPCB3IR/loomio-community')
   get '/timeline'   => redirect('http://www.tiki-toki.com/timeline/entry/313361/Loomio')
   get '/robots'     => 'robots#show'
+
+  get 'apps/registered' => 'dashboard#show'
```
robguthrie: easy breezy stub action?

rdbartlett:
Beautiful work James, thanks for the gifs

Connoropolous:
+5000

Issue #2971, opened 2016-02-17
vote icons rendering too big in notification emails since we introduded emojis


rdbartlett:
![image](https://cloud.githubusercontent.com/assets/970124/13123068/2c9536ee-d61f-11e5-8bd7-67d74b5025bd.png)

PR #2994, opened 2016-02-22
Remove problematic width for email images

HSalmon:
Addresses https://github.com/loomio/loomio/issues/2971

May need to test on clone?


gdpelican:
Were you going to test on clone @HSalmon ? I may be able to do a quick smoke tomorrow if not.

HSalmon:
That would be super this all good @gdpelican!

gdpelican:
Quick smoke on clone looks fine to me.

PR #2997, opened 2016-02-23
Fix subscription webhook call

gdpelican:
#2996 @mhjb


mhjb:
I tested it by setting up a new subscription, and it worked properly.

I also tested cancelling the subscription — which also worked properly: Chargify & the Loomio database were correctly updated and the group was turned into a Gift plan group. I even recorded a really boring screencast about it which I can share if you're interested.

Issue #2987, opened 2016-02-22
Set proposal modal closes on background click


gdpelican:
http://g.recordit.co/a7oVFedwg9.gif

PR #3004, opened 2016-02-24
Update proposal outcome form coffee to prevent clicking out of modal

HSalmon:
Addresses https://github.com/loomio/loomio/issues/2987


gdpelican:
Would you be willing to extend this to the extend proposal and close proposal modal as well (which I reckon have the same issue as the outcome one?)

Basically I want to get rid of the ProposalFormService, which doesn't need to exist anymore.

gdpelican:
Looks like a good one!

angular/core/components/thread_page/proposal_outcome_panel/proposal_outcome_panel.coffee
```diff
@@ -3,13 +3,13 @@ angular.module('loomioApp').directive 'proposalOutcomePanel', ->
...
   replace: true
-  controller: ($scope, AbilityService, ProposalFormService) ->
+  controller: ($scope, AbilityService, ModalService, ProposalOutcomeForm) ->
 
     $scope.canCreateOutcome = ->
       AbilityService.canCreateOutcomeFor($scope.proposal)
 
     $scope.openProposalOutcomeForm = ->
-      ProposalFormService.openSetOutcomeModal($scope.proposal)
+      ModalService.open ProposalOutcomeForm, proposal: => $scope.proposal
```
gdpelican: As an FYI, you don't need the `=>` here, because `$scope` is accessible here, whereas `@ / this` is not.

```
proposal: => @proposal # works
proposal: -> $scope.proposal # works
proposal: => $scope.proposal # works
proposal: -> @proposal # doesn't work, because '@ / this' is different when the method is called
```

PR #3008, opened 2016-02-24
Add activity item for when thread privacy is changed

HSalmon:
[Trello card](https://trello.com/c/LhOfstmI/707-activity-item-for-changing-thread-privacy)

screen shot 2016-02-24 at 5 50 37 pm

### DIFF
--- a/angular/core/components/thread_page/activity_card/discussion_edited_item_controller.coffee
+++ b/angular/core/components/thread_page/activity_card/discussion_edited_item_controller.coffee
@@ -7,6 +7,17 @@ angular.module('loomioApp').controller 'DiscussionEditedItemController', ($scope
       version.changes.title[1]
     else
       ''
+  $scope.onlyPrivacyEdited = ->
+    version.attributeEdited('private') and
+    !version.attributeEdited('title') or
+    !version.attributeEdited('description')
+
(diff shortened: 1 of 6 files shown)


HSalmon:
There are probably a couple of things in here that could be more concise - would love some eyes when someone's got a mo'!

robguthrie:
Nice! I don't think you need to say what the privacy was. There is only
ever one other option. Can I suggest "Actor changed the thread privacy to
public"

angular/core/components/thread_page/activity_card/discussion_edited_item_controller.coffee
```diff
@@ -7,6 +7,17 @@ angular.module('loomioApp').controller 'DiscussionEditedItemController', ($scope
       version.changes.title[1]
     else
       ''
+  $scope.onlyPrivacyEdited = ->
+    version.attributeEdited('private') and
```
gdpelican: Isn't this just `_.keys(version.changes) == ['private']`?

HSalmon:
screen shot 2016-02-24 at 7 42 34 pm

robguthrie:
sorry to be picky but can you drop the colon?

PR #3010, opened 2016-02-24
Logged out angular

gdpelican:
Time to spruce this up in anticipation of beta going away.


angular/core/services/exception_handler.coffee
```diff
@@ -1,11 +1,11 @@
...
-    if AppConfig.reportErrors
-      Airbrake.push
-        error:
-          message : exception.toString()
-          stack : exception.stack
-        params:
-          user_id: AppConfig.currentUserId
-
-    console.error "LoomioApp exception:", exception, cause
+# angular.module('loomioApp').factory '$exceptionHandler', (AppConfig) ->
```
gdpelican: Have to remember to uncommit this. Actually I might make this so it console logs the exception stack, which would be much helpful.

robguthrie: maybe this is the time when we start to report client exceptions to
airbrake? (maybe as a seperate account?)

gdpelican:
A bunch of things ended up going into this branch; here is a list of them:

- Add pre and next links for crawlers
  - On logged out view
  - For thread page and group page
- Allows thread and group page to respond to ?from parameter to load certain chunks of comments
- For thread page, from refers to sequenceId in discussion events
- For group page, from refers to public discussions in group discussions
  - (This has added a counter cache for public discussions on a group, which is updated when a discussion is created/updated, and when the group privacy is changed.)
- Add and style log in link to logged out angular view
- Removes star from logged out view
- Ensures that locale does not error when current user doesn’t exist
- Handles email preferences same as beta. Visiting /email_preferences
  - Logged in: get edit email settings page
  - Logged out, w/ unsubscribe token: get beta update email settings page
  - Logged out, no unsubscribe token: redirect to login page
- Add rss link to angular interface (thread / group page)
- Add ENV for google site verification to allow fetch as google
- Remove certain beta specs which were dumb / failing
- Add specs for prev / next links appearing on group / thread page

robguthrie:
And plenty more to come I guess too. Thanks for the work and also keeping
track of it so well James.

Issue #2094, opened 2015-04-08
Inviting someone already in a group re-notifies them


alannallama:
If you invite someone to a group who is already a member of that group, the app will notify the user that they have been added. Also the inviter sees a "member added" message. The system should know if that use is already a member of the group and not notify them (which could confuse the user who will think they weren't already in the group previously), and should display "member already in group" to the inviter. 

I think this is a relatively common situation actually... for example I want to invite a whole list of event attendees to a Loomio group and a subset of them are already members. Short of manually cleaning the email list, if I just invite them all the ones who are already in there will get a confusing notification.

A good example would be how Google Groups does this... if you invite a list of 10 people and 3 are already in the group, you will see a message that says "7 people added, 3 already in group", and only the new members will get notified.

alannaloomio:
@HSalmon and I tested this by re-inviting me to the Loomio Community group and I did get a notification.

robguthrie:
Great bug report. Thank you!

zacharystenger:
I was looking into this issue and have an idea of how to fix it.  
In ``InvitationsController#create`` we want to check if any of the users (``@invite_people_form.members_to_add``) are already in the group before we call ``MembershipService#add_users_to_group`` and ``InvitationService#invite_to_group``.  To do this I was thinking of adding a new private method to the ``InvitiationsController`` that loops through the users checking if ``group.has_member?(user)`` which calls a new method in the ``InvitePeopleForm`` which would remove them from ``members_to_add`` and ``emails_to_invite`` (since these are used throughout ``InvitationsController``) and return all the users not re-invited (to be used to display a proper flash message).  
Feedback would be much appreciated before I starting writing tests & coding.  Thanks!

PR #3014, opened 2016-02-25
Stop notifications for existing group members who are re-invited to g…

HSalmon:
…roup

Addresses https://github.com/loomio/loomio/issues/2094

screen shot 2016-02-25 at 1 10 34 pm

### DIFF
--- a/angular/core/components/invitation_form/invitation_form.haml
+++ b/angular/core/components/invitation_form/invitation_form.haml
@@ -18,6 +18,7 @@
 
     %label{for: "email-addresses", translate: "invitation_form.email_addresses"}
     %textarea.invitation-form__email-addresses.form-control#email-addresses{ng-model: "form.emails", ng-required: true, msd-elastic: "true", rows: 1, ng-model-options: "{debouce: 600}", ng-change: "storeDraft()", placeholder: "{{ 'invitation_form.email_addresses_placeholder' | translate }}" }
+    %validation_errors{subject: "form", field: "email"}
     .invitation-form__no-invitations.lmo-validation-error{ng-show: "noInvitations", translate: "invitation_form.messages.no_invitations"}
     .invitation-form__max-invitations.lmo-validation-error{ng-show: "maxInvitations()", translate: "invitation_form.messages.max_invitations"}
     %p{ng-if: "!showCustomMessageField()"}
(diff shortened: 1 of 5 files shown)


app/controllers/api/invitations_controller.rb
```diff
@@ -5,7 +5,11 @@ def create
                                                      group: @group,
                                                      inviter: current_user,
                                                      message: invitation_form_params[:message])
-    respond_with_collection
+    if @invitations.empty? and email_addresses.any?
```
gdpelican: I reckon this should fire whether there are email_addresses or not

app/services/invitation_service.rb
```diff
@@ -20,6 +20,8 @@ def self.invite_to_group(recipient_emails: nil,
                            message: nil,
                            group: nil,
                            inviter: nil)
+    group_members = group.members.pluck(:email)
+    recipient_emails = recipient_emails - group_members
```
gdpelican: `(recipient_email - group_members).map do |email|`

angular/core/components/invitation_form/invitation_form.haml
```diff
@@ -18,6 +18,7 @@
 
     %label{for: "email-addresses", translate: "invitation_form.email_addresses"}
     %textarea.invitation-form__email-addresses.form-control#email-addresses{ng-model: "form.emails", ng-required: true, msd-elastic: "true", rows: 1, ng-model-options: "{debouce: 600}", ng-change: "storeDraft()", placeholder: "{{ 'invitation_form.email_addresses_placeholder' | translate }}" }
+    %validation_errors{subject: "form", field: "email"}
```
gdpelican: Let's match this to the field name on the form, which is 'emails' (same as the ng-model above)

(We'll need to change the 422 response to reflect this as well)

app/controllers/api/invitations_controller.rb
HSalmon: If it fires whether there are email addresses or not then we'd need to update the error message - what should it be? @robguthrie, what do you think?

app/services/invitation_service.rb
HSalmon: Thanks :)

angular/core/components/invitation_form/invitation_form.haml
HSalmon: Can do

app/controllers/api/invitations_controller.rb
gdpelican: As it stands it shouldn't get to the server unless there are email addresses, since we disable the form. I'd be happy with a generic error message like

'We weren't able to send any invitations to the emails provided. They have been added to the group already.'

robguthrie: Can I suggest the text:

'We weren't able to send invitations to those email addresses because those
people are in the group already.'

PR #3017, opened 2016-02-25
Remove mark as read link from Yesterday on Loomio

mhjb:
Friendly devs, could you either test this, or give me a hint about how I can test it? 

I can't see in the tech manual how to play with emails, nor a relevant route in /development/.

Fixes #3017

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -39,7 +39,7 @@ en:
       headlines: Headlines
       mark_as_read: Mark as read
       thanks_for_reading: Thanks for reading, have a great day!
-        heres_what_happened: Here’s what happened on Loomio yesterday.
+        heres_what_happened: "Here’s what happened on Loomio yesterday."
       marked_as_read_success: Everything from the email has been marked as read
     comment_replied_to:
       context: '%{author_name} replied to your comment in the discussion "%{discussion_title}"'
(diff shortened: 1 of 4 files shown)


config/locales/server.en.yml
```diff
@@ -39,9 +39,7 @@ en:
       headlines: Headlines
       mark_as_read: Mark as read
       thanks_for_reading: Thanks for reading, have a great day!
-      click_here_to_mark_all_as_read_html: |
-        Here's what happened on Loomio yesterday.
-        Once you've read it, click <a href="%{link}">mark as read</a> to remove it from your Loomio Inbox.
+        heres_what_happened: Here’s what happened on Loomio yesterday.
```
gdpelican: We'll need to wrap this in double quotes so it can be parsed correctly (because of the apostrophe in 'here's'

mhjb: I'm confused — other lines in that file have curly apostrophes without being wrapped in double quotes…

gdpelican: Hm, alright. Well the build is erroring because of a parse error in that
file, so there's something up with that change.

gdpelican:
Whoops, looks like I scanned the error too quickly; problem seemed to be an indentation one.

mhjb:
Thanks! I just attempted to pushed another commit to fix it but see you beat me to it :)

Issue #3016, opened 2016-02-25
Can't interpolate {{privacyKey()}} error


gdpelican:
Reproed on prod here: https://www.loomio.org/d/Z5wFLhOc/data-portability-and-loomio-member-size-limit-reasons

![screen shot 2016-02-25 at 1 43 11 pm](https://cloud.githubusercontent.com/assets/750477/13306151/098fba78-dbc6-11e5-8e0a-8219247c96fa.png)

PR #3018, opened 2016-02-25
Fix error

HSalmon:


HSalmon:
screen shot 2016-02-25 at 4 10 04 pm

HSalmon:
Addresses (hopefully): https://github.com/loomio/loomio/issues/3016

gdpelican:
Hey @HSalmon I want to hold off on this one just a little while longer; I don't quite yet feel confident in it.

Looking at this, I'm not quite sure why there needs to be two different null checks, since our element looks like this:

```
        %span
          %span{ng-if: "onlyPrivacyEdited()", translate: "{{privacyKey()}}"}
```
(PS nested spans seems a bit fishy there :P)

So wondering if we actually need to check to see if this should appear in both `privacyKey()` and `onlyPrivacyEdited()`?

HSalmon:
@gdpelican, double span thing is how we're doing the text for the other activity items. Can we pair on this at some point? Would like to get it merged because it's got the missing translation strings in it

PR #3022, opened 2016-02-26
Notify user when they have been made coordinator of a group

HSalmon:
[Trello card](https://trello.com/c/1nygVRew/1208-3-i-want-to-be-notified-when-i-am-made-a-coordinator-of-an-existing-group)

screen shot 2016-02-26 at 2 23 48 pm


gdpelican:
Perfect!

PR #3025, opened 2016-02-26
Tweaks for production deployment

toupeira:
A few tweaks to make S3 and SMTP optional, and avoid using `@loomio.org` when sending mails.


gdpelican:
Looks great, thanks @toupeira !

PR #3028, opened 2016-02-27
Fix some notifications not linking correctly

gdpelican:


angular/core/components/thread_page/thread_page_controller.coffee
```diff
@@ -32,9 +32,8 @@ angular.module('loomioApp').controller 'ThreadPageController', ($scope, $routePa
   @init = (discussion) =>
     if discussion and !@discussion?
       @discussion = discussion
-      if @discussion.hasActiveProposal() and $location.search().proposal == @discussion.activeProposal().key
-        @proposalToFocus = @discussion.activeProposal()
-
+      @activeProposalKey = $routeParams.proposal or $location.search().proposal
+      @activeCommentId   = parseInt($routeParams.comment or $location.search().comment)
```
gdpelican: This lets us take our comment id from the route params `d/abc/comment/123`, or `?comment=123` (so we can be backwards compatible)

angular/core/services/lmo_url_service.coffee
```diff
@@ -22,20 +19,23 @@ angular.module('loomioApp').factory 'LmoUrlService', (AppConfig) ->
...
+    proposal: (p, params = {}) ->
+      @route model: p.discussion(), action: "proposal/#{p.key}", params: params
+
+    comment: (c, params = {}) ->
+      @route model: c.discussion(), action: "comment/#{c.id}", params: params
+
     membership: (m, params = {}, options = {}) ->
       @route model: m.group(), action: 'memberships', params: params
 
+    membershipRequest: (mr, params = {}, options = {}) ->
```
gdpelican: Instead of lmo-hreffing `/d/abc?comment=123`, (which breaks because the 'new' angular router can't handle query params in that way, we're pointing to `/d/abc/comment/123`, meaning we can pick up the value from $routeParams, rather than $location.search()

gdpelican:
Would love a quick peek at this @robguthrie

robguthrie:
Oh wow. That's great.

Yeah. I like it. Well done.

Issue #3021, opened 2016-02-25
Change user mentioned email title to include discussion title


rdbartlett:
When someone mentions me in two different threads in one group, the subject line is the same for both messages. This means my email client stacks both messages into one email thread, so the reply-to function is clumsy and confusing. 

Suggested solution: 
replace the subject line "Robert Guthrie mentioned you in the Loomio Community - Contributors group on Loomio" with a line like "Robert Guthrie mentioned you while discussing 'Loomio in te reo'"

(I think this is only the case when "Activity from the threads I participate in" is not selected in my email settings.)

gdpelican:
@rdbartlett Do you think we need different email subjects based on whether you're following a discussion or not?

PR #3037, opened 2016-03-01
Update subject for user mentioned email

HSalmon:
Address https://github.com/loomio/loomio/issues/3021

Alternative subject will be sent to users if they're not following the discussion


gdpelican:
Looks good to meee! :monkey_face:

Issue #3002, opened 2016-02-23
Logged in angular user should redirect from https://www.loomio.org/start_group to https://www.loomio.org/dashboard?start_group=true


rdbartlett:

gdpelican:
Do you have a block on this one @HSalmon ?

PR #3039, opened 2016-03-01
Logged in angular users should be redirected when starting group

HSalmon:
Addresses https://github.com/loomio/loomio/issues/3002

### DIFF
--- a/app/controllers/start_group_controller.rb
+++ b/app/controllers/start_group_controller.rb
@@ -2,12 +2,12 @@ class StartGroupController < ApplicationController
   def new
     @errors = []
     @group = Group.new
-    if current_user.present? && current_user.angular_ui_enabled
-      redirect_to dashboard_path(start_group: true)
-    elsif current_user.present? && !current_user.angular_ui_enabled && ENV['NEW_GROUPS_ON_ANGULAR']
-      render :enable_angular
-    else
+    if !current_user_or_visitor.is_logged_in?
       render :new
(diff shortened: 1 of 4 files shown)


app/controllers/start_group_controller.rb
```diff
@@ -2,7 +2,9 @@ class StartGroupController < ApplicationController
   def new
     @errors = []
     @group = Group.new
-    if current_user.present? && ENV['NEW_GROUPS_ON_ANGULAR']
```
gdpelican: Hm, this feels complicated to me. What if we dropped the NEW_GROUPS_ON_ANGULAR setting (I don't think we need it), and used the `current_user_or_visitor` helper, which would make it so we don't have to make changes when we hop to logged out angular?

```
if current_user_or_visitor.angular_ui_enabled?
  render :enable_angular
else
  render :new
end
```

?

HSalmon: I think the enable_angular method shows the screen we want to bypass, so don't we just want redirect them to dashboard with the modal open if angular is enabled?

gdpelican: Ah, ok, I'm with you now. So there's actually 3 paths

- logged in, angular enabled = redirect to dashboard
- logged in, angular not enabled = render enable_angular
- logged out, render new

Is that right?

I think dropping the ENV is still alright, and would like to move to always using `current_user_or_visitor.is_logged_in?` rather than `current_user.present?` for these controllers

PR #3047, opened 2016-03-02
Update client.en.yml

rdbartlett:
Keen to update the text on the angular switch page before we encourage a bunch more people over from beta.


rdbartlett:
To go in tandem with this announcement to recently active english users not on angular

announce

robguthrie:
@rdbartlett do you want to remove the "switch back to old loomio" link from within the app?

rdbartlett:
Not yet but soon. 20% of MAUs are on old Loomio so I don't want to freak them out.

PR #3048, opened 2016-03-02
Allow closed proposals to load their own discussions / votes

gdpelican:
Need to test this more but looks good at first blush. A little unsure about the duplication in ClosedMotionSerializer

### DIFF
--- a/app/services/vote_cache.rb
+++ b/app/services/vote_cache.rb
@@ -12,7 +12,7 @@ class VoteCache
   end
 
   def get_for(motion)
-    @votes_by_motion_id[motion.id]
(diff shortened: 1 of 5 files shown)


app/services/vote_cache.rb
```diff
@@ -12,7 +12,7 @@ def initialize(user, votes)
   end
 
   def get_for(motion)
-    @votes_by_motion_id.fetch(motion.id) { Vote.new(motion: motion, user: user) }
+    @votes_by_motion_id[motion.id]
```
gdpelican: This might break voting horribly in beta, not sure. I'm thinking the test suite will catch it if it does.

PR #3064, opened 2016-03-07
Extract angular feedback card and routes

HSalmon:


gdpelican:
@HSalmon a really nice next step here would be to add a couple tests around the routes we've put in, like

```
get :about
expect(response).to be_success
expect(response).to include 'About us'

get :blog
expect(response).to redirect_to 'http://blog.loomio.org'
```

app/views/application/_footer.html.haml
```diff
@@ -1,5 +1,5 @@
 %footer.footer{role: 'contentinfo'}
-  - if hosted_by_loomio?
-    = render 'loomio_org_footer'
+  - if content_for? :footer
+    = yield :footer
   - else
-    = render 'independent_install_footer'
```
gdpelican: Gotta put the footer back in for the loomio.org instance

plugins/plugins.yml
```diff
@@ -15,3 +15,7 @@ plugins:
   loomio_webhooks:
     repo:    loomio/loomio_webhooks
     version: master
+
+  loomio_org:
+    repo:    loomio/loomio_org
```
gdpelican: I've renamed this to loomio_org_plugin FYI

PR #3065, opened 2016-03-07
replace the mark summary as read img

robguthrie:
this was taken out when it should not have been.


gdpelican:
You'll have to put the helper method back in as well I think, per here:
https://github.com/loomio/loomio/commit/cac48dfbf0f20572389684004d30757fde78d204

robguthrie:
Thanks gd

Issue #3062, opened 2016-03-06
Data export errors


mhjb:
Visiting /g/GROUPKEY/export.html as a user with rights to use the data export feature, I get "We're sorry, but something went wrong."

gdpelican:
@HSalmon we simply have to remove the `inviter_name` method from `invitation.rb`; I should have caught this in CR, sorry.

mhjb:
Mind letting me know when this is deployed so I can test & tell ze client?

PR #3071, opened 2016-03-08
Remove inviter_name method from invitation.rb

HSalmon:
Addresses https://github.com/loomio/loomio/issues/3062


gdpelican:
Can we confirm and/or write a test which makes sure that the groups#export method works after this change?

Issue #3024, opened 2016-02-26
Start thread privacy notice wrong


mhjb:
This is a secret subgroup.

![image](https://cloud.githubusercontent.com/assets/1380991/13341448/b37b5592-dc9d-11e5-92f4-d9c11fd4d005.png)

robguthrie:
@gdpelican could this be a drafts issue? I cannot reproduce it

mhjb:
It might have involved a draft, but if so it was an empty one.

gdpelican:
It's totally a draft thing. Have to take some some to reproduce to figure
how exactly what's going on though.

robguthrie:
I wonder if you can look into the produciton db and see

gdpelican:
I can't repro this to save my life, and see no evidence of it in the prod DB. Are you still seeing this in a particular place, @mhjb ?

rdbartlett:
(It's working properly for me on Contributors and Co-op Members subgroups)

mhjb:
Still erroring for me — I could demo if that would be any help.

I notice there are 267 rows in the drafts table with my user_id on … a clue?

PR #3073, opened 2016-03-08
Fix drafts bug

gdpelican:


gdpelican:
This is the easy fix, and really nice to have a spec around this.

Not convinced this is the most robust solution, but considering we've had very few complaints about drafts outside of this one, I'm happy to fix it and move on. Thoughts @robguthrie ?

gdpelican:
For #3024

PR #3077, opened 2016-03-08
Open proposal outcome form for motion closed notification

HSalmon:


angular/core/models/notification_model.coffee
```diff
@@ -16,5 +16,9 @@ angular.module('loomioApp').factory 'NotificationModel', (BaseModel) ->
     group: ->
       @event().group()
 
+    actionPath: ->
+      switch @kind()
+        when 'motion_closed', 'motion_closed_by_user' then 'outcome'
```
gdpelican: As these get more complicated, we'll want to make a `actionPathMap` in `notification_model.coffee`, so this becomes

```
actionPath: ->
  @constructor.actionPathMap[@kind()]
```

which will keep this a bit cleaner

angular/core/components/thread_page/thread_page_controller.coffee
```diff
@@ -19,6 +20,11 @@ angular.module('loomioApp').controller 'ThreadPageController', ($scope, $routePa
     @discussion.activeProposal().key == $location.search().proposal and
     AbilityService.canVoteOn(@discussion.activeProposal())
 
+  @openOutcomeModal = ->
+    $routeParams.outcome? and
+    @proposal and
```
gdpelican: Can we move this existence check into the AbilityService?

```
canSetOutcomeFor: (proposal) ->
  proposal? and proposal.isClosed() and....
```

PR #3084, opened 2016-03-09
Tighten api controllers (WIP)

gdpelican:
This is a little refactor I did a while ago to pull out some anomalous code in our controller layer.

Hooray for great test coverage!


gdpelican:
This will require some more work but I'll chip away at it bit by bit.

gdpelican:
Closing this as I don't think it's a great use of time at this particular moment.

Issue #3085, opened 2016-03-09
Cursor should look like a hand when I hover over the edit thread context link and the load earlier activity link


rdbartlett:

gdpelican:
Should be as simple as adding `cursor: pointer` on the anchor, although this has been a bit intermittent for me, so might be something a little hairier.

PR #3089, opened 2016-03-09
Make cursor a pointer on thread page links

HSalmon:
Addresses https://github.com/loomio/loomio/issues/3085


gdpelican:
It feels to me like this should be a helper method `lmo-pointer`?. I'm even wondering if we could get away with

```
a { cursor: pointer; }
```

gdpelican:
I'm going to merge this in any case, but seems like a bit of technical debt we could pick up at some point.

PR #3094, opened 2016-03-10
add support for inbound email post requests

robguthrie:
Stripping the griddler stuff from loomio-email-gateway app (that was supposed to be a microservice but wall really just a big overhead in terms of maintenance) and moving into loomio core.

This will make it way easier to setup reply by email when you're setting up your own loomio instance.

not quite ready.. I need to change the post stuff to a service call and write a test


robguthrie:
closing this because I'm merging it back into the docker pr

PR #3096, opened 2016-03-10
Github style markdown table styling

mhjb:
|Before|After|
|---|---|
|image|image|


gdpelican:
Whoa, great improvement!

Out of curiosity, do you know what happens if there's a table with heaps of
columns?

mhjb:
Tries to squish em all in, with some mysterious width calculation for each
row involving the number of characters in the most characterful cell in
that column.

PR #3100, opened 2016-03-11
Upload help text appears on group theme hover

HSalmon:


angular/core/components/group_page/group_theme/group_theme.coffee
```diff
@@ -34,3 +36,12 @@ angular.module('loomioApp').directive 'groupTheme', ->
 
     $scope.logoHoverOut = ->
       $scope.logoHover = false
+
+    $scope.triggerThemeHover = ->
+      $('.group-theme__trigger-upload').hover ->
```
gdpelican: we won't want to use jQuery anymore (I thought we took it out actually), instead we want

```
document.querySelector('selector')
```

angular/core/components/group_page/group_theme/group_theme.haml
```diff
@@ -1,6 +1,6 @@
 .group-theme
-  .group-theme__cover{ng-style: "coverStyle()"}
-    .group-theme__upload-photo{ng-mouseover: "themeHoverIn()", ng-mouseleave: "themeHoverOut()", ng-if: "canUploadPhotos()"}
+  .group-theme__cover.group-theme__trigger-upload{ng-style: "coverStyle()", ng-mouseover: "themeHoverIn()", ng-mouseleave: "themeHoverOut()", ng-if: "canUploadPhotos()"}
```
gdpelican: Also, I'm not that sure you need javascript for this. What if we did something like

```
.group-theme__cover           group-theme__upload-photo { display: none; }

.group-theme__cover:hover group-theme__upload-photo { display: block; }
```
?

HSalmon: Hey @gdpelican, this doesn't work because the `group-theme__header` element also needs to trigger the upload prompt and `group-theme__header` and `group-theme__upload-photo` aren't siblings or anything.

gdpelican: I don't really follow, but we can't be doing css manipulations in angular controllers unless we really need to. Let's close for now, as it's not that high a priority.

PR #3115, opened 2016-03-15
Use locally hosted emojis

gdpelican:
@HSalmon or @robguthrie would you be willing to smoke this on clone? (I'm pushing it up to clone now)


HSalmon:
Hey @gdpelican, tried to test this on clone but had more trouble deploying I'm sorry.

gdpelican:
The SSL cert on clone is bung at the moment (ping @robguthrie?), but the
deploy seems to have gone through fine. Can you do a quick smoke that the
emojis load up in all the places they should on the frontend?

gdpelican:
Going to merge this now as the functionality looks good on clone, even if clone's functionality isn't great at the moment.

PR #3120, opened 2016-03-16
Add warning bandaid to old help

mhjb:
image

### DIFF
--- a/app/views/help/help.html.haml
+++ b/app/views/help/help.html.haml
@@ -1,5 +1,6 @@
 .row
   .col-md-12#help-content
+    %p.alert-info.alert= t :"new_help.notice_html"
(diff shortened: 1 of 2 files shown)


gdpelican:
Hrm, if I'm logged in, and have the angular UI enabled, that 'switch to the new version' actually takes me to the page to deactivate angular, which confused me. :red_circle: 

What if we changed it to have two translation strings:

- This is the help manual for the old, 'beta' version of Loomio. [Get up to date help here]()
- You can also [switch to the new version here])
(^^ only display this if `current_user_or_visitor.angular_ui_enabled?` is false)

:recycle:

gdpelican:
Going to close this; decomming beta will make this obsolete anyway.

PR #3122, opened 2016-03-16
Don't allow visitors to mark as read or mute

gdpelican:
I think this just needs a quick once-over CR; works for me on `development/view_open_group_as_visitor`


gdpelican:
I'm just going to merge this.

robguthrie:
Great decision. Until we get automatic smoke testing environments running
(which I think we should do real soon) it feels quite heavy to smoke test
these kinds of things.

PR #3127, opened 2016-03-17
Squish last sequence unread bug yay!

gdpelican:


gdpelican:
Hey @HSalmon would you mind smoke testing this on clone?

I don't really need to see that the bug is fixed (it's tough to repro), I just want to feel sure that marking threads as read is still working as expected.

gdpelican:
I keep seeing an issue on here where we don't mark the last item as the thread as read, so this isn't ready for merge yet. Need to figure out what ails it.

PR #3128, opened 2016-03-17
Allowing printing angular thread page

gdpelican:
This lets angular provide a nice print page for the thread pages. Other pages are less pretty, but thread page is the primary use case for printing, so this is MVP for beta feature parity.

Would love a smoke test on clone; something @mhjb or @rdbartlett could take?

screen shot 2016-03-17 at 4 37 56 pm
screen shot 2016-03-17 at 4 38 03 pm
screen shot 2016-03-17 at 4 38 12 pm


gdpelican:
NB that a next iteration will be to take you to a new page with all the content loaded for printing.

In the meantime, I'm happy with the naive CTRL+P implementation, with a workaround of hitting 'load previous content' a few times before printing to allow the whole thing to be printed in one page.

robguthrie:
James are you able to add a card for the next stage so that it's able to be
prioritised?

gdpelican:
Yep sure.

PR #3130, opened 2016-03-17
Update email settings

HSalmon:
[Trello card](https://trello.com/c/QFHBZ6v0/873-13-update-email-settings-for-user-thread-and-group)
[Design doc](https://docs.google.com/document/d/1Q0nvkMaE_bBiuj3aZvlt-S4VGOsptyUkBxDPxoFBJiQ/edit#heading=h.bae5jxxh8nzq)


angular/core/components/change_volume_form/change_volume_form.coffee
```diff
@@ -0,0 +1,42 @@
...
+
+    $scope.title =
+      if setDefault
+        ''
+      else
+        model.title or model.groupName()
+
+
+    $scope.submit = ->
+      $scope.isDisabled = true
```
gdpelican: We'll want to move this into the FormService; I think something like this will work:

```
$scope.submit = FormService.submit $scope, $scope.model,
  submitFn: (model) ->
    model.saveVolume($scope.buh.volume, $scope.applyToAll, $scope.setDefault
  flashSuccess: $scope.someFunctionToGetTheFlashSuccess
```

angular/core/components/change_volume_form/change_volume_form.coffee
```diff
@@ -0,0 +1,42 @@
...
+      if setDefault
+        "change_volume_form.set_membership_default"
+      else
+        "change_volume_form.#{$scope.model.constructor.singular}"
+
+    $scope.title =
+      if setDefault
+        ''
+      else
+        model.title or model.groupName()
```
gdpelican: We want to use $scope.model (rather than model) throughout because that refers to the cloned version of the model

angular/core/components/email_settings_page/email_settings_page_controller.coffee
```diff
@@ -1,7 +1,20 @@
...
 
   @user = CurrentUser.clone()
 
+  @groupVolume = (group) ->
+    group.membershipFor(CurrentUser).volume()
+
+  @changeDefaultMembershipVolume = ->
+    ModalService.open ChangeVolumeForm,
+      model: => (_.first CurrentUser.groups()).membershipFor(CurrentUser),
+      setDefault: => true
```
gdpelican: The more I see this, the more I want to make this more polymorphic, so that we can pass the CurrentUser for defaultMembershipVolume, the Group for group volume, and a Discussion for discussion volume.

I think that'll eliminate the need for this setDefault boolean, and this pretty-awkward 'get the membership of the current user's first group'  call

:grey_question:

angular/core/models/membership_model.coffee
```diff
@@ -4,7 +4,7 @@ angular.module('loomioApp').factory 'MembershipModel', (BaseModel, AppConfig) ->
     @plural: 'memberships'
     @indices: ['id', 'userId', 'groupId']
     @searchableFields: ['userName', 'userUsername']
-    @serializableAttributes: AppConfig.permittedParams.membership
```
gdpelican: We need to move this into permittedParams.membership (in `permitted_params.rb` on the server), rather than replacing it on the client. :red_circle:

gdpelican: Also why `volume_value` rather than just `volume`?

angular/core/models/membership_model.coffee
```diff
@@ -20,9 +20,24 @@ angular.module('loomioApp').factory 'MembershipModel', (BaseModel, AppConfig) ->
...
-      @volume = volume
-      @save()
+    volume: ->
+      @volumeValue
+
+    saveVolume: (volume, applyToAll = false, setDefault = false) ->
+      @volumeValue = volume
+      @applyToAll = applyToAll
+      @setDefault = setDefault
+      @save().then =>
```
gdpelican: I'm not too sure about having this things as attributes on the model... I know what a discussion's title is, or a group's privacy, but I'm not sure what a membership's 'applyToAll' is... these are details about the particular post, and don't really have to do with the model itself.

Which makes me think that `save()` isn't the right way to go here, I think we want to move the out into a separate controller endpoint:

```
@remote.postMember 'update_volume', @id,
  volume: volume
  apply_to_all: applyToAll
  set_default: setDefault
```

```
# api/memberships_controller
def change_volume
  service.change_volume(membership: load_resource, params: params, actor: current_user)
  respond_with_resource
end
```

```
# membership_service
def self.change_volume(membership:, params:, actor:)
  actor.ability.authorize! :update, membership
  if (params.slice(:set_default) { update stuff }
  if (params.slice(:apply_to_all) { update stuff }
  EventBus.broadcast('membership_change_volume') { }
end
```

app/models/membership.rb
```diff
@@ -58,7 +60,15 @@ def user_name_or_email
     return user_name ? user_name : user_email
   end
 
+  def discussion_readers
```
gdpelican: This should definitely be a has_many relationship, although I haven't quite figured out how to get the right one.
It's something like

```
  has_many :discussion_readers, -> { group: self.group) }, through: :user
```

angular/core/models/membership_model.coffee
robguthrie: volume on discussion is a function (which returns discussion_reader_volume or membership_volume), volume on membership is a value. To give them the same interface we need volume on membership to be a function, so we moved the value to volume_value.

angular/core/models/membership_model.coffee
robguthrie: yep, I think that's a good improvement

angular/core/components/change_volume_form/change_volume_form.coffee
robguthrie: maybe we should write this so that access to the original version is not possible?

angular/core/components/change_volume_form/change_volume_form.coffee
robguthrie: I'd really like to sit down with you and discuss how we could do this. At the time form service just made things way more complex.

gdpelican: I don't follow; I just gave that above code a smoke test and it seems to work alright (at least all of the flash messages are showing up correctly.)

The idea is that `submitFn` is a function which takes the second FormService argument (in this case, $scope.model), as an argument.

(ie, we go: 
```
FormService.submit = (scope, model, options) ->
  options.submitFn(model)
```

So we want to write a function as submitFn which takes the model and does what we want, and then we get all of the magic of the form service like flash messaging, error handling, form disabling, and closing the modal automatically.

robguthrie: Cool. I think at the time we did not understand it.. This comment would make a great addition to the docs. For instance we did not understand that flash message could be a function.

gdpelican: Sweet; shall I put an entry into docs/en/somewhere titled 'How to use FormService'?

robguthrie: yea please!

robguthrie: I think there is going to be a docs/en/development/ folder (and a
docs/en/deployment one too)

robguthrie:
Smoke tested on loomioclone (feedback implemented) and all is well in angular land. Good to go I reckon!

Issue #3109, opened 2016-03-13
Proposal closed should be a salient activity item


rdbartlett:
"Proposals that have just ended should probably come up on your dashboard as unread if you haven't looked at them to see the result. I feel that "proposal ended" is a type of thread activity. Same with 'outcome published'."

PR #3138, opened 2016-03-21
Add motion_closed and motion_closed_by_user to salient items

HSalmon:
Addresses https://github.com/loomio/loomio/issues/3109

Does this need tests around it @gdpelican? Also, I noticed while I was testing it locally that it seemed like some of my own activity was showing up as unread. Will try and investigate further.

### DIFF
--- a/app/models/discussion.rb
+++ b/app/models/discussion.rb
@@ -1,6 +1,11 @@
 class Discussion < ActiveRecord::Base
   PER_PAGE = 50
-  SALIENT_ITEM_KINDS = %w[new_comment new_motion new_vote motion_outcome_created]
+  SALIENT_ITEM_KINDS = %w[new_comment
+                          new_motion
+                          new_vote
+                          motion_closed
+                          motion_closed_by_user
+                          motion_outcome_created]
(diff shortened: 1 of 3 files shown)


gdpelican:
Hrm, would need a smoke test. I'll put it on my list. Can you push up to clone tomorrow after smoke testing the unread bug?

gdpelican:
We need to add some events to the `author_thread_item!` call in the event bus, as stated here:
https://github.com/loomio/loomio/issues/3143

HSalmon:
@robguthrie, we've got email-related failures in here

robguthrie:
Looks like the same failures we fixed yesterday. Can you merge that work into this one?

HSalmon:
@robguthrie, @gdpelican, if someone would be up for reviewing this that would be great :)

gdpelican:
These events aren't being broadcast through the MessageChannelService, so we'll need to add that as well so they live update

repro of unwanted behaviour:
- User A closes a proposal while User B is viewing the inbox
- User B does not see that unread activity until reloading

gdpelican:
Cool, this is a pre-existing issue that we're going to handle separately :green_apple:

PR #3139, opened 2016-03-22
remove piwik and ensure ahoy tracks all url changes as views

robguthrie:
Wonder if we should remove Google analytics too.. Ahoy is way more useful for most stuff


gdpelican:
Would love to remove GA; I certainly don't use it anymore, and if you + @mhjb don't either, I reckon it can go.

robguthrie:
I think we should/could keep it on the marketing page.

mhjb:
I use GA all the time

gdpelican:
Looks good to me. :green_apple: Merge away!

PR #3151, opened 2016-03-23
remove admin dashboard because it's not used and it's slow

robguthrie:


gdpelican:
Let me smoke this before merge?

robguthrie:
certainly

gdpelican:
I just routed `/admin` to the groups page, and got rid of the dashboard entirely instead of having a blank landing screen. Should be good to merge if you're happy :green_apple:

PR #3154, opened 2016-03-24
Fix rspec deprecation warnings

gdpelican:


gdpelican:
This is a rabbit hole we don't need to spelunk at the moment. :)

PR #3155, opened 2016-03-24
Beta go byebye

gdpelican:


Gemfile
```diff
@@ -44,30 +45,28 @@ gem 'fog'
...
+# gem 'momentjs-rails'
 gem 'ruby-progressbar'
-gem 'responders', '~> 2.0'
+# gem 'responders', '~> 2.0'
 gem 'activerecord_any_of'
 gem 'rack-attack', '~> 4.2.0'
-gem 'ahoy_matey'
+gem 'ahoy_matey', '~> 1.1.1'
 gem 'ahoy_email'
-gem 'oj'
```
robguthrie: oj is supposedly a faster json implementation. I'm happy to take it out but it's not beta

robguthrie: and it may affect json serialization in angular

gdpelican: Oh cool, I'll throw it back in.

app/assets/stylesheets/unstructured/_pages.scss
```diff
@@ -121,6 +121,8 @@ a.btn-icon:hover { color: $dark-grey; }
 .pages.crowd,
 .pages.wallets,
 .woc.index,
+.start_group.new,
+body.passwords,
 .we_the_people.index {
```
robguthrie: this can go. But we can also remove it later

gdpelican: Yeah it's real bad, eh?

I was happy to get all of the non-angular pages into one application.html.haml template at least.

app/controllers/email_actions_controller.rb
```diff
@@ -19,19 +19,13 @@ def mark_discussion_as_read
...
   def mark_summary_email_as_read
-    @inbox = Inbox.new(user)
-    @inbox.load
-    time_start = Time.at(params[:time_start].to_i).utc
+    time_start  = Time.at(params[:time_start].to_i).utc
     time_finish = Time.at(params[:time_finish].to_i).utc
 
-    Queries::VisibleDiscussions.new(user: user,
-                                    groups: user.inbox_groups).
+    Queries::VisibleDiscussions.new(user: user).
```
robguthrie: Looks like a slight behaviour change here. Are you sure you want to do this?

gdpelican: Hm, so this came out of a desire to get rid of the concept of `inbox_groups`, which used the concept of `memberships.inbox_position`, which isn't used.

I'll take a look in the DB about the nature of the current state of inbox_position; maybe I can just back this out and leave the `inbox_groups` method on user for now.

app/controllers/invitations_controller.rb
```diff
@@ -1,114 +1,29 @@
...
-    render 'application/display_error',
-      locals: { message: t(:'invitation.invitation_cancelled') }
-  end
-
-  rescue_from Invitation::InvitationAlreadyUsed do
-    if current_user and @invitation.accepted?
-      redirect_to @invitation.invitable
+  rescue_from(ActiveRecord::RecordNotFound)    { respond_with_error :"invitation.invitation_not_found" }
+  rescue_from(Invitation::InvitationCancelled) { respond_with_error :"invitation.invitation_cancelled" }
+  rescue_from(Invitation::InvitationAlreadyUsed) do
```
robguthrie: Are there tests around invitiation already used and team link invititation?

app/controllers/redirect_controller.rb
```diff
@@ -1,45 +0,0 @@
-class RedirectController < ActionController::Base
```
robguthrie: I'm not sure we want to take this out actually. Infact this will break group subdomains wont it?

app/controllers/robots_controller.rb
```diff
@@ -2,10 +2,6 @@ class RobotsController < ActionController::Base
   respond_to :text
 
   def show
-    if ENV['ALLOW_ROBOTS'].to_i == 1
-      head :ok
-    else
-      render :private_robots
-    end
+    ENV['ALLOW_ROBOTS'].to_i == 1 ? head(:ok) : render(:private_robots)
```
robguthrie: I don't think ternary operators are more readable, sorry! (don't undo but just know it makes it a bit harder for me to read)

Also I think we can follow the convention of just checking if ENV is set or not.

app/controllers/start_group_controller.rb
```diff
@@ -1,49 +1,18 @@
 class StartGroupController < ApplicationController
+
   def new
     @errors = []
     @group = Group.new
     if !current_user_or_visitor.is_logged_in?
       render :new
-    elsif current_user_or_visitor.angular_ui_enabled?
+    else current_user_or_visitor.angular_ui_enabled?
```
robguthrie: this looks broken... Why is there code on the else line?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,7 +1,6 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   skip_before_action :verify_authenticity_token
   include OmniauthAuthenticationHelper
-  include InvitationsHelper
```
robguthrie: Why are you removing the invitations helper from here?

robguthrie: I think... we had it so that when you are signing up to join a group whe can say so on the form.

app/controllers/users/registrations_controller.rb
```diff
@@ -1,48 +1,28 @@
...
-    end
-
-    if omniauth_authenticated_and_waiting?
-      load_omniauth_authentication_from_session
-      @user.name = @omniauth_authentication.name
-      @user.email = @omniauth_authentication.email
-    end
+    @user = User.new(
+      name:  @omniauth_authentication&.name  || invitation_from_session&.recipient_name,
+      email: @omniauth_authentication&.email || invitation_from_session&.recipient_email
```
robguthrie: oh i see. you've got it covered.

app/extras/js_runtime.rb
```diff
@@ -40,7 +40,7 @@ def runtime
   end
 
   def execjs_asset
-    File.read [:public, client_asset_path(:"execjs.min.js")].join('/')
+    File.read [:public, client_asset_path(:"execjs.min.js")].join('/').squeeze('/')
```
robguthrie: cool method

robguthrie:
It's a whopper!! I've read through what github will let me read

app/controllers/redirect_controller.rb
gdpelican: Oh right, I can put back in.

app/controllers/invitations_controller.rb
gdpelican: No, I just have the happy paths; I'll write some for the unhappy paths. I've had a really great time writing some e2ez around this stuff

app/controllers/email_actions_controller.rb
robguthrie: As a matter of process I think you should avoid rewriting stuff you don't have to. Leaving those bits as small changes to make in the future seems less error prone when doing something this large.

app/controllers/invitations_controller.rb
robguthrie: Thanks!

app/controllers/robots_controller.rb
gdpelican: Moved it back to
```
if ENV['robots']
  thing
else
  other thing
```

app/controllers/start_group_controller.rb
gdpelican: Whoa, great catch thanks.

gdpelican: Updating to be a simple if/else; we shouldn't care about angular_ui_enabled? anymore.

robguthrie:
@gdpelican I've just realised you need to remove a lot of translations files as well.
all the *.yml and help.*.yml files should be removed. All server strings should be in server.*.yml now. 
If we find some are missing we'll need to move them over.

PR #3159, opened 2016-03-28
Switch in new icons, remove unused images

mhjb:
Swapped out a million PNGs for four SVGs. I feel confident to use SVGs for icons because of [caniuse.com](http://caniuse.com/#feat=svg) and because [Angular Material](https://material.angularjs.org/latest/api/directive/mdIcon) seem keen on them.

I made some judgments about colours and synchronised them across the app. 

I stripped out some unused images from public/img/, plus some unused css. 

screen shot 2016-03-29 at 12 31 20 pm
screen shot 2016-03-29 at 12 31 24 pm
screen shot 2016-03-29 at 12 31 51 pm

@HSalmon  do you have capacity to smoke-test this?


robguthrie:
FYI - we will need to test that emails that use these icons work.

mhjb:
I'll make some PNGs for emails

rdbartlett:
I would like to have some realtime with you on this Matt. Things I would like to play with:

* outer strokes on the small circle icons
* smaller internal knuckle/finger lines
* more colour distance between disagree and block
* less brightness in the agree

But this is already a huge improvement so I am not going to get in the way

mhjb:
Cool let's do it. 2.30 Thurs?

I also need some help with email icons — you might know what I don't.

rdbartlett:
2.30 is good. I don't think I will have your email icon answer alas.

mhjb:
Now email icons are working. 

Have left old icons in print page, as this will shortly be switched out for an angular version.

PR #3161, opened 2016-03-29
User admin panel: link group memberships to groups

mhjb:
image

Very minor — needs smoke test @gdpelican ?

### DIFF
--- a/app/admin/users.rb
+++ b/app/admin/users.rb
@@ -63,7 +63,7 @@ ActiveAdmin.register User do
         if can? :deactivate, user
           button_to 'Deactivate User', deactivate_admin_user_path(user), method: :put, data: {confirm: 'Are you sure you want to deactivate this user?'}
         else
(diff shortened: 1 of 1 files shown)


robguthrie:
I don't think this needs a smoke test.

app/admin/users.rb
```diff
@@ -86,7 +86,7 @@ def find_resource
       table_for user.memberships.each do |m|
         column :group_id
         column :group_name do |g|
-          Group.find(g.group_id).full_name
+          link_to Group.find(g.group_id).full_name, [:admin, Group.find(g.group_id)]
```
gdpelican: I really don't want to make 2 db calls inside a table here.
```
  group = g.group # I think g is a membership here?
  link_to group.full_name, [:admin, group]
```

mhjb:
Thanks for that @gdpelican — I couldn't quite figure it out. Amending commit now.

gdpelican:
Sweet, thanks! :green_apple:

PR #3162, opened 2016-03-29
changes to plugins system. ongoing

robguthrie:
not quite ready.. going to make sure it still works first.


.gitignore
```diff
@@ -25,13 +25,7 @@ public/client
...
 public/uploads
 
-plugins/*/
-plugins/Gemfile
-!plugins/*.rb
-!plugins/*.yml
-!plugins/*.md
-# NB: To reveal a plugin to hack on it locally, add a line here telling .gitignore to reveal it:
-# !plugins/my_awesome_plugin
+plugins/
```
robguthrie: This seems way simpler than before.

Ive taken the "this is how you unignore notice out because I think people should use their own repo from from the start" I'd recommend people `git init` and commit their plugins rather than expose them to the loomio repo.

lib/plugins/base.rb
```diff
@@ -1,4 +1,4 @@
-require "#{Rails.root}/plugins/repository"
+require "#{Rails.root}/lib/plugins/repository"
```
robguthrie: @gdpelican can you think of any other places I need to make this change?

.gitignore
gdpelican: At least for me I need to gitignore because otherwise it doesn't show up when I use Atom to navigate to files or search for them using cmd+P

lib/plugins/base.rb
gdpelican: I'd have a global repo search for `:plugins`, I see some in `plugins.rake`, and `plugins/base.rb`

gdpelican:
I'd like this to wait to go in until after the extract-loomio-org stuff please :)

Dockerfile
```diff
@@ -23,12 +23,12 @@ COPY config/database.docker.yml /loomio/config/database.yml
...
 WORKDIR /loomio/angular
 RUN npm install
+RUN gulp build
 
 WORKDIR /loomio
 RUN bundle install
 
-# set environment to production
-ENV RAILS_ENV development
+ENV RAILS_ENV production
```
robguthrie: dont commit this.. it's wrong

PR #3163, opened 2016-03-30
List coordinators first on memberships page

HSalmon:


HSalmon:
screen shot 2016-03-30 at 4 07 27 pm

gdpelican:
Brilliant! :green_apple:

PR #3164, opened 2016-03-30
Add thread moved activity item

HSalmon:


angular/core/components/thread_page/activity_card/discussion_moved.haml
```diff
@@ -0,0 +1,10 @@
+.thread-item{ng-controller: "DiscussionMovedItemController"}
+  .thread-item__header.media
```
robguthrie: Hrrmm.. are there examples of doing this without using .media? I think we want to avoid using .media if posssible.. but if there are no other examples then that's ok.

robguthrie:
Looks good thanks @HSalmon

angular/core/components/thread_page/activity_card/discussion_moved.haml
HSalmon: All the activity items use the same format

robguthrie: oh wow. ok. Thanks for that. Let me just say that I'd like to move away from bootstrap stuff (we don't use much now). This is fine if all the rest use the same format (should be pretty easy to remove/replace later), but new code should avoid bootstrap unless it provides incredible value.

PR #3165, opened 2016-03-30
Build angular client in Dockerfile

robguthrie:
@gdpelican nothing too crazy here. Can I merge please?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ public/client
 public/system
 public/uploads
 
+disabled_plugins/
 plugins/*/
 plugins/Gemfile
 !plugins/*.rb
--- a/lib/tasks/deploy.rake
+++ b/lib/tasks/deploy.rake
@@ -46,6 +46,12 @@ def heroku_migrate_and_restart(remote)
                 "ruby /usr/local/heroku/bin/heroku run rake loomio:notify_clients_of_update -a #{remote}"]
 end
 
+desc "Copy development client to a versioned client for production"
+task :version_client_assets do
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -25,6 +25,7 @@ public/client
 public/system
 public/uploads
 
+disabled_plugins/
```
robguthrie: this is just temporary I think. Just a place I can move a plugin so it's out of the way

lib/tasks/deploy.rake
```diff
@@ -46,6 +46,12 @@ def heroku_migrate_and_restart(remote)
                 "ruby /usr/local/heroku/bin/heroku run rake loomio:notify_clients_of_update -a #{remote}"]
 end
 
+desc "Copy development client to a versioned client for production"
+task :version_client_assets do
```
robguthrie: breaking up the big deploy and push method into smaller bits.

robguthrie:
actually, hold off and I'll update the readme now too.

robguthrie:
oh.. that's a different repo. 

Right.. requesting merge.

lib/tasks/deploy.rake
gdpelican: Will you then call this from within the deploy rake task?

gdpelican:
Looks alright to meeee! :green_apple:

lib/tasks/deploy.rake
robguthrie: I assume one would. I don't need a deploy rake task for the docker based system though.

I though that It would be better not to touch that stuff while you're working on it.

PR #3170, opened 2016-04-02
Fix type mismatch errors for did_not_votes and comments

gdpelican:


robguthrie:
Thanks :apple:

robguthrie:
guh.. I'm not sure how to do your emoji based thing.. but this is good.

PR #3171, opened 2016-04-02
FIX: Only load 3 closed proposals for group previous proposal card

gdpelican:


robguthrie:
What does this fix james?

gdpelican:
It's a perf improvement; right now we're loading all of the previous
proposals on the group page (50 proposals, 50 discussions), but only
displaying 3 of them, which is slowing down the time to bring up the group
page for larger groups like the Loomio community.

robguthrie:
oh ok, cool. so no visible behaviour change. Thanks √

Issue #3168, opened 2016-03-30
Open group I'm not a member of should show previous proposals panel


mhjb:
![image](https://cloud.githubusercontent.com/assets/1380991/14160673/51e14da2-f73a-11e5-962f-00596ac6c3e4.png)

gdpelican:
#3172

mhjb:
When I first arrived at https://www.loomio.org/g/NohQxyr9/nzgoal-software-extension-discussion-of-draft (not as a member), I didn't see the Previous Proposals panel. I joined the group, browsed around a bit, then left the group. Then I browsed some more, and happened to notice a [discussion](https://www.loomio.org/d/54Vxapcb/status-of-gpl-in-govt-software-projects) that had a previous proposal.  When I returned to the group page, as a not-member, I can now see the Previous Proposals panel.

robguthrie:
we've seen this inconsistent behaviour in the tests too.. There is a race
condition going on somewhere

gdpelican:
This was a bug where open groups were showing proposals, but closed groups with public content weren't. Now fixed #3241

mhjb:
Doesn't seem to be fixed — http://unctadyouthforum.loomio.org has some proposals but they didn't show up on group page for me until I visited the relevant threads.

PR #3172, opened 2016-04-02
Display previous proposals card for open groups when logged out

gdpelican:
#3168


robguthrie:
ignore hound. I was just trying it out. I hate that kinda shit.

gdpelican:
Yeah could be useful if we get the settings just right, but +1 to getting
rid of it.

PR #3176, opened 2016-04-04
Increase size of vote indicator in thread preview

mhjb:
Before
image

After
image


gdpelican:
Looks strictly better to me :+1:

PR #3179, opened 2016-04-05
Hook up new user help videos

HSalmon:


angular/core/components/group_page/group_help_card/group_help_card.coffee
```diff
@@ -3,11 +3,14 @@ angular.module('loomioApp').directive 'groupHelpCard', ->
...
   replace: true
-  controller: ($scope, CurrentUser, AppConfig, UserHelpService) ->
+  controller: ($scope, $sce, CurrentUser, AppConfig, UserHelpService) ->
     $scope.showVideo = AppConfig.loadVideos
 
     $scope.helpLink = ->
       UserHelpService.helpLink()
 
+    $scope.helpVideo = ->
+      $sce.trustAsResourceUrl(UserHelpService.helpVideo())
```
gdpelican: Wonder if UserHelpService could have a method `helpVideoUrl` which returned the appropriate url, so we didn't have to dup the $sce stuff?

HSalmon: Having trouble getting it to work that way...hmmm

gdpelican: Does it not like importing $sce into the service? I think I've had issues doing that with stuff like $location in the past. It's not a showstopper by any means.

HSalmon: It doesn't seem to be enjoying it, no.

gdpelican:
Perfect! :green_apple:

PR #3181, opened 2016-04-05
Make America (aka submitOnEnter) great again

gdpelican:


gdpelican:
Would love a smoke test for this; should work for creating a group, thread, proposal, vote, and comment.

The weird corner case is making sure that submitting a proposal/vote on the thread page doesn't accidentally submit a comment as well.

Can I get a smoke test from someone @robguthrie @HSalmon ?

gdpelican:
Still looking for a QA on this please.

mhjb:
I'm ready to MAGA!!!

mhjb:
Vote form seems to have nothing in focus when loaded (I'd expect Share your thoughts to have focus). And you can't Cmd+Enter until you give it focus.

mhjb:
Though maybe that's an accessibility thing — perhaps the modal title has focus. Anyway, one can't click vote button then Cmd+Enter to make it so.

* Start a Proposal Cmd+Enter works.
* Add a Comment doesn't. Well it didn't the first time I tried (after mucking about with voting & proposing), though when I refreshed the page it started working.
* Start a Group doesn't, even after refreshing.
* Weird behaviour: I type 'More' in the Comment field and quickly hit Cmd+Enter. A comment with just 'M' shows up. I do the same thing again — quickly type more + Cmd+Enter, and I get _Comment cannot be empty_.

image

gdpelican:
@mhjb Okay I've

- added submitOnEnter to the group form
- added an autofocus to the vote modal
- We've got a fun clash here between drafts and cmd+enter which looks like this:
The input field is debounced, meaning it takes ~300ms to update the model, because we want to see if you're going to type more before we save a draft of it (otherwise we end up with hundreds of calls to save the draft). However, if you cmd+enter in that 300ms window, you get a 'stale' version of the model (sans the last character or so you typed).

I've added a timeout which waits for 300ms after you've cmd/entered, which should ensure the draft clears... which I think isn't too annoying, but am open to lowering the debounce time as well.

Wasn't able to repro the comment one not working (other than in the way described above), so if there are specific repro steps for that I'd be interested.

gdpelican:
Cool, I'd love either a go-ahead or another smoke test on this @mhjb

mhjb:
* Start group works*
* Start thread works*
* Start proposal works*
* Vote & change your position work
* Comment works (even if you go really quick)

Asterisked items are prone to show 'Cannot be blank' if you hit Cmd+Enter really quick

robguthrie:
I'm happy with this when you are @gdpelican 🍏

PR #3182, opened 2016-04-06
Add switch over to angular banner for beta

HSalmon:
screen shot 2016-04-06 at 1 59 35 pm

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -185,4 +185,12 @@ module ApplicationHelper
       image_url("navbar-logo-beta.jpg")
     end
   end
+
+  def show_beta_banner?
+    if controller_name =='start_group' || controller_name == 'sessions'
+      false
+    else
+      true
+    end
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -187,10 +187,6 @@ module ApplicationHelper
   end
 
   def show_beta_banner?
-    if controller_name =='start_group' || controller_name == 'sessions'
-      false
-    else
-      true
-    end
+    ['dashboard', 'inbox', 'groups', 'discussions'].include? controller_name
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -187,6 +187,6 @@ module ApplicationHelper
   end
 
   def show_beta_banner?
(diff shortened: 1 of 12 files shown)


gdpelican:
@rdbartlett Where does the 'more info...' link go to? Is it the blog post @alannallama is working on?

gdpelican:
@HSalmon can you confirm that this won't show up on that sign up / sign in / start_group / marketing pages? (Which will exist even after beta goes away?)

rdbartlett:
Yep it will go to the blog post. We'll need to deploy this feature and the
blogpost at approximately the same time.

rdbartlett:
I'm expecting: beta in-app banner, angular welcome modal, and blog post all
to deploy concurrently.

HSalmon:
@gdpelican, I believe that last commit ensures that it won't show up sign in/up pages etc.

app/helpers/application_helper.rb
```diff
@@ -185,4 +185,12 @@ def logo_path
       image_url("navbar-logo-beta.jpg")
     end
   end
+
+  def show_beta_banner?
```
robguthrie: I think we should whitelist rather than blacklist. I don't feel like I know every controller in the system and don't feel super comfortable putting the banner on all the pages.

Can we just pick up DashboardController, InboxController, GroupsController and DiscussionsController?

Also whenever you have an if condition where the body of the condition is true or false, you can safely remove the condition and just evaluate the statement.

So in this case I think the method should be:
```
['dashboard_controller', 'groups_controller', 'discussions_controller'].include? controller_name
```

config/locales/server.en.yml
```diff
@@ -300,3 +300,8 @@ en:
         Images are the same, but prefixed with an exclamation mark.
 
         ![loomio logo](https://i.imgur.com/Fvtworb.png)
+
+  beta_banner:
+    new_loomio_html: |
+      <strong>The New Loomio is here!</strong> It's faster, prettier, and mobile-ready. More info...
```
gdpelican: How is our blog post coming along @rdbartlett ? We should get this thing, plus the angular modal thing built and merged.

rdbartlett: Blog post is ready when you are.

HSalmon:
screen shot 2016-04-12 at 12 16 43 pm

gdpelican:
Here we gooo!

PR #3188, opened 2016-04-08
increase the number of attempts allowed for a job to 10.

robguthrie:
The last attempt will be 2 hours after it was created.


robguthrie:
Actually this is dumb, I'm going to move it to ENV

robguthrie:
I want to merge this soon as a dependency on other server stuff.

gdpelican:
Seems fine to me :green_apple:

PR #3194, opened 2016-04-11
Build explore page

HSalmon:
screen shot 2016-04-08 at 4 31 46 pm


angular/core/components/explore_page/explore_page.haml
```diff
@@ -0,0 +1,18 @@
+.lmo-one-column-layout
+  %main.email-settings-page
```
gdpelican: %main.explore-page

HSalmon:
screen shot 2016-04-12 at 4 54 44 pm

HSalmon:
screen shot 2016-04-15 at 4 54 12 pm

HSalmon:
@gdpelican, @robguthrie - would love a code review early next week if possible!

Gemfile.lock
```diff
@@ -637,7 +637,7 @@ GEM
     simplecov-html (0.10.0)
     snorlax (0.1.3)
       cancancan (~> 1.10.1)
-      rails (~> 4.1)
+      rails (> 4.1)
```
gdpelican: Let's revert this change.

angular/core/components/explore_page/explore_page_controller.coffee
```diff
@@ -0,0 +1,43 @@
...
+  $timeout -> document.querySelector('#search-field').focus()
+
+  @groups = =>
+    Records.groups.find(@groupIds)
+
+  @search = =>
+    if @query
+      @groupIds = []
+    Records.groups.fetchExploreGroups(@query, {from: @groupIds.length, per: @perPage}).then (object) =>
+      @groupIds = @groupIds.concat(_.map object.groups, (group) -> group.id)
```
gdpelican: How about
```
@groupIds.concat _.pluck(object.groups, 'id')
```
?

angular/core/components/explore_page/explore_page.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.explore-page__group {
+  width: 365px;
+  @include cardNoPadding;
+}
+
+.explore-page__group-cover {
+  height: 100px;
+  background: url('http://placehold.it/640x160') no-repeat center center #f5f5f5;
```
gdpelican: Are we able to lose the placehold.it background? ❔

angular/core/components/explore_page/explore_page_controller.coffee
```diff
@@ -0,0 +1,43 @@
...
+    @groups().length > 0
+
+  @searchResultsMessage = ->
+    if @groups().length == 1
+      'explore_page.single_search_result'
+    else
+      'explore_page.multiple_search_results'
+
+  @noResultsFound = ->
+    !@searching && (@groups().length < @perPage || !@canLoadMoreGroups)
```
gdpelican: Hm, is the intended effect here that the message is displayed if we have less than one page of results? ❔

angular/core/components/navbar/navbar.scss
```diff
@@ -59,6 +59,7 @@
 
 .lmo-navbar__sign-in {
   @include lmoBtnFeatured;
+  margin: 7px 0;
```
gdpelican: Could you screenshot what this one does? ❔

app/controllers/api/groups_controller.rb
```diff
@@ -1,6 +1,12 @@
 class API::GroupsController < API::RestfulController
   load_and_authorize_resource only: :show, find_by: :key
   load_resource only: [:upload_photo, :use_gift_subscription], find_by: :key
+  skip_before_action :authenticate_user!, only: :index
```
gdpelican: Hrm, I'm surprised to see we need to do this, as we shouldn't be authing the user for group actions across the board.

What kind of error do you get if this line goes away?

app/extras/queries/explore_groups.rb
```diff
@@ -0,0 +1,17 @@
+class Queries::ExploreGroups < Delegator
+  def initialize
+    @relation = Group.where(is_visible_to_public: true).
+                            parents_only.
```
gdpelican: Cool to see the nice lean explore group criteria.

spec/extras/queries/explore_groups_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    it 'only shows parent groups' do
+      subgroup = FactoryGirl.create(:group, parent: group)
+      subgroup.update_attribute(:is_visible_to_public, true)
+      expect(Queries::ExploreGroups.new).to_not include subgroup
+    end
+  end
+
+  describe '#search_for' do
```
gdpelican: These are great-looking specs for this class @HSalmon 🍏

gdpelican:
@rdbartlett How do we link to the explore page?

gdpelican:
@HSalmon could you elaborate on what doesn't work for logged out users?

app/controllers/api/groups_controller.rb
HSalmon: This was left in from when Rob and I were trying to make it work for logged out users. It doesn't work. I'll remove. However, when you try and access the explore page as a logged out user you get redirected to sign in page. Would be keen to work out how to avoid that. I think the BaseController ```  before_filter :authenticate_user!``` is being applied.

gdpelican:
Some further feedback:

- I found myself wanting to click on the little people icon to go to the memberships page. Maybe just for open groups?
- It's also not incredibly obvious what the comment and people icons mean; could we put a title onto those buttons to that effect? ('25 people are in this group'; '15 discussions have happened in this group')
- It would be nice to include proposal count here as well imho; thoughts on that @mhjb ?
- The search results aren't ordering on the page, meaning I click 'show more' and end up with a different first group than I had before (= pretty confusing)
- I want to rip out the pg_search gem. It's doing advanced search stuff I don't think we need like lexeme matching, and we don't use it anywhere else. I'm happy to pair on this at some point.
- With a single column of groups, they end up left-aligned, which looks a little wonky:
screen shot 2016-04-18 at 12 00 48 pm
Maybe we could apply `flex-grow: 1` to them, so they take up the available space?

angular/core/components/explore_page/explore_page_controller.coffee
HSalmon: Yeah, the message says "Can't find the group you're looking for?...", and, according to the design, it is displayed when there are no results or when you've gotten to the end of the results.

app/controllers/api/groups_controller.rb
gdpelican: I was able to fix by moving `include AngularHelper` from BaseController to ApplicationController, then changing the route from `base#boot_angular_ui` to `application#boot_angular_ui`

HSalmon: That's what Rob and I did initially, but it broke all of the tests. If the tests are broken for you, perhaps we could pair on making it work?

robguthrie: Yea.. At some point Hannah and I did that. I think Hannah may have moved it
back trying to fix broken specs. Note to all that I think BaseController
should be renamed to LoggedInController

gdpelican: Are we thinking that this goes in before the decom beta stuff? This is fixed in that branch.

angular/core/components/navbar/navbar.scss
HSalmon: screen shot 2016-04-20 at 10 53 51 am
screen shot 2016-04-20 at 10 54 48 am

HSalmon:
@gdpelican, I think I've implemented your feedback. Would you mind checking and perhaps smoke-testing? Or perhaps @mhjb could smoke-test if he has capacity?

mhjb:
@HSalmon I think I need to push other things out of the way to make room for sales until we're closer to meeting our targets.

HSalmon:
That is a-ok @mhjb, totally understandable

mhjb:
Just a couple thoughts:

* My design had the people & discussions icons & counts really light grey; partly so they wouldn't feel clickable (just wanted to give a hint at the level of activity). 
* I like the idea of add proposal counts

gdpelican:
This is fantastic work @HSalmon, thank you!

rdbartlett:
This is such awesome work Hannah!!!

rdbartlett:
Belated reply to your question James: right now this is only linked from logged in beta and logged out front page. 

I expect soon we'll replace blank dashboard with 'start or join a group' though

gdpelican:
That design would be on the top of my design pipeline wishlist.

rdbartlett:
@mhjb can you point me to any relevant work that has happened in this area (if you've done any?)

mhjb:
This is all I've done on that front: https://www.loomio.org/d/en67cwpd/getting-into-the-product-with-no-group

rdbartlett:
Thanks Matt, I'm going to continue this conversation over there where it is in context.

Issue #3150, opened 2016-03-23
clicking change email preferences from email takes me to beta interface even when I'm logged in


robguthrie:
It would be better if, when I'm logged in I was taken to the normal email preferences page, as this is the best one to use.

PR #3198, opened 2016-04-12
Allow usage of angular email settings page from email unsubscribe link

gdpelican:
For #3150


gdpelican:
Does @mhjb or @HSalmon want to smoke test this?
I tested by:
- visiting `development/setup_group_with_expired_legacy_trial_admin`
- making jennifer grey a coordinator
- logging out and then logging in as jennifer
- visiting the Dirty Dancing Shoes group

spec/controllers/email_preferences_controller_spec.rb
```diff
@@ -2,14 +2,27 @@
 
 describe Users::EmailPreferencesController do
   describe "edit" do
-    let(:user) { create (:user) }
+    let(:user) { create :user, angular_ui_enabled: true }
 
-    it 'renders the beta interface when an unsubscribe token is present' do
+    it 'renders the beta interface for betea user when an unsubscribe token is present' do
```
robguthrie: I think you need to update the string. It says it's going to render beta interface... and I don't think you mean it

gdpelican: ??

The logic's been simplified to say 'if the current_user_or_visitor has angular enabled, show the angular version, otherwise show the beta version'

This test is 'I am a user with angular off, and an unsubscribe token. I should get the beta interface' I don't think we can show beta users the angular interface for just one page, because there's no easy path back to beta from there.

robguthrie: Can we ditch the beta page altogether? It's totally different to the angular one. I'd be keen to work out how we could make the angular page available to signed out users with unsubscribe tokens.

gdpelican: Won't matter in a couple of weeks because beta's going away.

robguthrie: I would love to have a 10 minute vebal run through what will be required. Not blocking at all - but it's something that could be done any time and really improve things

app/serializers/restricted/user_serializer.rb
```diff
@@ -0,0 +1,10 @@
+class Restricted::UserSerializer < ActiveModel::Serializer
+  embed :ids, include: true
+  attributes :id, :restricted, :username, :email, :email_when_proposal_closing_soon, :email_missed_yesterday,
```
gdpelican: I ended up not being able to care about extending modules like this to extract it out, as it ended up being more of a pain than it was worth.

app/helpers/current_user_helper.rb
```diff
@@ -1,9 +1,23 @@
 module CurrentUserHelper
```
gdpelican: There's some duplication here. Will revisit after beta's gone, I think most of it can go.

app/controllers/api/profile_controller.rb
```diff
@@ -45,7 +45,11 @@ def resource_class
   end
 
   def resource_serializer
-    CurrentUserSerializer
+    if current_user == restricted_user
```
gdpelican: Make sure we're not serializing out full users in response to restricted user actions

angular/core/services/current_user.coffee
```diff
@@ -1,18 +1,20 @@
...
-  if AppConfig.currentUserId?
+  # User has an email (ie, has a Loomio account)
+  if currentUser.id?
     AppConfig.membershipsLoaded = true
     $rootScope.$broadcast 'currentUserMembershipsLoaded'
 
-    Records.discussions.fetchInbox().then ->
-      AppConfig.inboxLoaded = true
-      $rootScope.$broadcast 'currentUserInboxLoaded'
+  # User is authenticated (ie, not signed in via an unsubscribe token)
```
gdpelican: Only request notifications and inbox for authenticated users

gdpelican:
Okay this one's good for a look @robguthrie . It needs a spec (or maybe three), but I don't have time right now.

PR #3200, opened 2016-04-12
Don't pass nil user to comment service in griddler

gdpelican:


gdpelican:
http://errbit.loomio.org/apps/56ef80443132390011000000/problems/57084ab8657272000e00001f

gdpelican:
It occurs to me that we want to handle this from within the service, actually. But way more work at this point to go through every action and make sure there's a user of some kind trying to do it.

robguthrie:
Acutally I don't think this is the right thing to do. It never makes sense
to create a comment as a logged out user (currently). And if we do this,
then the service returns false.. which is not really explicit behaviour.

Can we put a bang on the User.find_by so the response is a 404. I think
that makes more sense.

gdpelican:
Service should raise an auth on `user.ability.authorize!(:create, comment)`, which will raise 403 unauthorized. I think that's more appropriate for this than 404 not found?

robguthrie:
oh cool. Did not see that.

PR #3202, opened 2016-04-12
fix angular on docker image

robguthrie:
@gdpelican FYI I've moved the versioning of the client assets into the build method.


gdpelican:
Yeah! I've fixed this in the beta branch, but we need it now too. 🍏

PR #3205, opened 2016-04-13
If the bootstrap:create_user task fails to create the user it should say so and why.

cillian:
Before if your password was too short it would say the user was created even though it wasn't.


gdpelican:
Sweet, thanks @cillian

PR #3210, opened 2016-04-16
close motion before validation if necessary

### LINKED ISSUE #3203: Unable to set proposal outcome
I click 'set outcome' on a recently closed proposal and the modal pops up. I type in the message and click 'publish outcome'. The 'loading' circle spins for a moment, but nothing happens. I press 'publish outcome again' - same thing. I tried on Chrome and Firefox - same thing.

robguthrie:
fixes #3203

### DIFF
--- a/app/models/motion.rb
+++ b/app/models/motion.rb
@@ -45,6 +45,7 @@ class Motion < ActiveRecord::Base
 
   after_initialize :set_default_closing_at
 
+
   attr_accessor :create_discussion
 
   scope :voting,                   -> { where(closed_at: nil).order(closed_at: :asc) }
@@ -94,6 +95,18 @@ class Motion < ActiveRecord::Base
 
   def closed?
     closed_at.present?
+  end
+
+  def needs_to_be_closed?
+    (!closed? and closing_at < Time.now)
+  end
+
+  def close!
+    did_not_votes.delete_all
+    non_voters = group_members - voters
+    DidNotVote.create! non_voters.map { |user| {motion: self, user: user} }
+    update(closed_at: Time.now,
+           did_not_votes_count: did_not_votes.count)
   end
 
   def has_votes?
@@ -227,20 +240,6 @@ class Motion < ActiveRecord::Base
 
   def group_members_without_outcome_author
     group.members.without(outcome_author)
-  end
-
-  def store_users_that_didnt_vote
-    did_not_votes.delete_all
-    group.users.each do |user|
-      unless user_has_voted?(user)
-        did_not_vote = DidNotVote.new
-        did_not_vote.user = user
-        did_not_vote.motion = self
-        did_not_vote.save
-      end
-    end
-    update_attribute(:did_not_votes_count, did_not_votes.count)
-    reload
   end
 
   private
--- a/app/services/motion_service.rb
+++ b/app/services/motion_service.rb
@@ -10,6 +10,7 @@ class MotionService
   end
 
   def self.update(motion:, params:, actor:)
+    close(motion) if motion.needs_to_be_closed?
     actor.ability.authorize! :update, motion
 
     if params[:closing_at] && motion.closing_at.to_s == Time.zone.parse(params[:closing_at]).to_s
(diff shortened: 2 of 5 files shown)


app/models/motion.rb
```diff
@@ -96,6 +98,18 @@ def closed?
...
 
+  def close_if_lapsed!
+    close! if (!closed? and closing_at < Time.now)
+  end
+
+  def close!
+    did_not_votes.delete_all
+    non_voters = group_members - voters
+    DidNotVote.create! non_voters.map { |user| {motion: self, user: user} }
+    update(closed_at: Time.now,
```
gdpelican: Do we need Time.zone.now here?

app/models/motion.rb
```diff
@@ -44,6 +44,8 @@ class Motion < ActiveRecord::Base
   has_paper_trail only: [:name, :description, :closing_at, :outcome]
 
   after_initialize :set_default_closing_at
+  before_validation :close_if_lapsed!
```
gdpelican: I think this is totally okay to go in, but want to keep my ear out for a better way. Here's what makes me a little uncomfortable:
- I think it's a little odd to have such a slow event (load up a bunch of relations, create a bunch of models, then save = lots of stuff for the user to wait on) happen before validation.
- 'before validation' happens quite a lot, esp. in the test suite. Also, we'd get slightly weird behaviour if a motion is invalid, tries to save, but still closes. ('I'm updating the title to blank; whoops that didn't work. Wait why is it closed now, and I can't edit anymore?')

I also wonder if it'd be possible to bump the DidNotVotes creation (which is the slow bit) off to a side job after the motion has closed?

```
EventBus.listen('motion_closed', 'motion_closed_by_user') do |motion|
  motion.did_not_votes.delete_all
  motion.did_not_votes.create!((motion.group_members - motion.voters).map { |user| {motion: motion, user: user} }
end
```

💛

robguthrie:
Thanks @gdpelican. I'm not happy about it either.

Infact it's a buggy solution because when it closes itself it won't create events. So I don't think we should merge it yet.

The core of the problem is there is always going to be some amount of time where "lapsed" motions exist. But I think I got carried away deciding how bad the problem is - maybe we should just close_if_lapsed as a step in the updating methods of the motion service?

app/models/motion.rb
robguthrie: Nope: Time.now will return UTC time, which is fine for the db.

spec/services/motion_service_spec.rb
```diff
@@ -35,7 +35,7 @@
         MotionService.create(motion: motion, actor: user)
       end
 
-      it "syncs the discussion's search vector" do
+      it "syncs the discussion's search vector", focus: true do
```
gdpelican: Will want to drop the `focus` here before merge

app/models/motion.rb
```diff
@@ -96,6 +97,18 @@ def closed?
     closed_at.present?
   end
 
+  def needs_to_be_closed?
+    (!closed? and closing_at < Time.now)
```
gdpelican: Yeah so much better.

robguthrie:
@gdpelican I reckon this is good to merge. Don't want to background the did_not_votes because I'm not sure we know how that will behave in the system.

Good news is we're improving the perf of it by moving from many inserts to one bulk insert.

spec/services/motion_service_spec.rb
robguthrie: done ta

app/services/motion_service.rb
```diff
@@ -36,27 +37,20 @@ def self.reopen(motion, close_at)
...
-    motion.closed_at = Time.now
-    motion.save!
-
+    motion.close!
     EventBus.broadcast('motion_close_by_user', motion, user)
     Events::MotionClosedByUser.publish!(motion, user)
   end
 
   def self.create_outcome(motion:, params:, actor:)
+    close(motion) if motion.needs_to_be_closed?
```
gdpelican: I found myself thinking 'is it odd that an unauthorized user can close a lapsed motion by trying to create an outcome for it?'

But... meh. 💛

gdpelican:
This is great! 🍏

PR #3212, opened 2016-04-17
Ensure email params don't blow up as often

gdpelican:
http://errbit.loomio.org/apps/56ef80443132390011000000/problems/570eb4b665727200670003f5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ cache:
     - angular/node_modules
 
 env:
-  - DB=postgresql CXX=g++-4.8 CANONICAL_HOST=loomio.example.org
+  - DB=postgresql CXX=g++-4.8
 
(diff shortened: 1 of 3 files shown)


robguthrie:
thanks james 🍏

.travis.yml
```diff
@@ -9,7 +9,7 @@ cache:
     - angular/node_modules
 
 env:
-  - DB=postgresql CXX=g++-4.8
+  - DB=postgresql CXX=g++-4.8 CANONICAL_HOST=loomio.example.org
```
robguthrie: why do you need to do this?

gdpelican: I don't, heh. I modified the EmailParams class slightly to allow it to be tested a little more easily.

PR #3213, opened 2016-04-18
clean up smtp config

robguthrie:
pulled this out of the email preview branch. All environments set SMTP settings the same way.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -54,12 +54,6 @@ module Loomio
 
     # Version of your assets, change this if you want to expire all your assets
     config.assets.version = '1.4'
-
-    config.action_mailer.default_url_options = {
-      host:     ENV['CANONICAL_HOST'],
-      protocol: ENV['FORCE_SSL'] ? 'https' : 'http'
-    }
-
     config.roadie.url_options = nil
 
     # required for heroku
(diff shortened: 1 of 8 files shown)


config/environments/production.rb
```diff
@@ -29,13 +29,6 @@
   # Defaults to Rails.root.join("public/assets")
   # config.assets.manifest = YOUR_PATH
 
-  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  if ENV['FORCE_SSL']
```
gdpelican: Does force ssl default to true? ❔

gdpelican:
Does `development/last_email` (which the test suite relies on) still work with these changes?

Other than that looks good to me 🍏

config/application.rb
```diff
@@ -87,5 +81,29 @@ def self.fog_credentials
...
       }
     end
+
+    config.action_mailer.default_url_options = {
+      host:     ENV['CANONICAL_HOST'],
+      port:     ENV['CANONICAL_PORT'],
+      protocol: ENV['FORCE_SSL'] ? 'https' : 'http'
+    }
+
+    config.force_ssl = ENV.has_key?('FORCE_SSL')
```
robguthrie: @gdpelican doing force_ssl for all environments here. but no it does not default to true :)

PR #3215, opened 2016-04-18
clean up count columns on motions

robguthrie:
not ready but this is the gist


app/models/motion.rb
```diff
@@ -101,11 +100,24 @@ def needs_to_be_closed?
...
+    end
+  end
+
   def close!
     did_not_votes.delete_all
     non_voters = group_members - voters
     DidNotVote.create! non_voters.map { |user| {motion: self, user: user} }
     update(closed_at: Time.now,
+           members_count: group.members_count,
            did_not_votes_count: did_not_votes.count)
```
robguthrie: I'll drop did_not_votes_count

PR #3216, opened 2016-04-18
pre migrate motions to have voters_count and members_count

robguthrie:


robguthrie:
just going to copy production db to clone and run this mirgation there to understand how long it takes and if it has terrible locking properties

gdpelican:
🍏

Issue #3173, opened 2016-04-03
Vertical alignment of vote activity item


rdbartlett:
![image](https://cloud.githubusercontent.com/assets/970124/14234864/1ce877e6-fa42-11e5-953e-4d73c9021acf.png)

I would expect the icon, name, and statement to all be vertically aligned

rdbartlett:
This is an edge case: the width of the statement is small and it is multiple lines long. I wouldn't spend too much energy trying to fix it. Just wanted to register it here in case there is an easy fix

gdpelican:
My guess is `vertical-align: top` on the div holding 'Name Obscured disagreed:'

HSalmon:
This is one fix:
<img width="434" alt="screen shot 2016-04-20 at 2 50 28 pm" src="https://cloud.githubusercontent.com/assets/2882097/14662117/5778aa12-0707-11e6-9596-90736a4d0d0a.png">

But would it be better to strip out new lines?

rdbartlett:
I think that fix looks fine, I don't think we should strip new lines unless
we're sure we want to.

HSalmon:
Cool, PR over here: https://github.com/loomio/loomio/pull/3230

PR #3230, opened 2016-04-20
Align vote activity item components

HSalmon:
Address: https://github.com/loomio/loomio/issues/3173


gdpelican:
Sweet, thanks @HSalmon

PR #3232, opened 2016-04-20
support more smtp_configs and asset_host

robguthrie:


robguthrie:
images in emails works with this.

PR #3233, opened 2016-04-20
We still ref category_explore_path on the beta group page

gdpelican:


gdpelican:
Closed in favor of #3234

PR #3238, opened 2016-04-20
Handle Module::DelegationError errors from Comment via email

robguthrie:
We're seeing a few Module::DelegationErrors and I think this might be a good way to deal with them.

Check the discussion exists as part of permissions checking, so failure to reference existing discussion results in unauthorised.

http://errbit.loomio.org/apps/56f0588e31323900110002ad/problems/5717bc276572720248000034

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -215,7 +215,7 @@ class Ability
     end
 
     can [:create], Comment do |comment|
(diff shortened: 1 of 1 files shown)


gdpelican:
OMG it'll be so nice to get rid of those annoying errors. 🍏

Issue #3227, opened 2016-04-20
You can't tell from looking at the context panel when it's been updated


mhjb:
![image](https://cloud.githubusercontent.com/assets/1380991/14661513/4cfa423a-0702-11e6-92a1-7b5b1f6ebd8b.png)

In this long discussion I was looking for an up-to-date summary. It turned out that the context panel had recently been updated, but I couldn't tell that by looking at it — it just tells me Rob started it 7 months ago. I think it might be more useful to roughly follow the format of the edit revisions modal and say "Edited by Alanna Krause 2 hours ago"

HSalmon:
You can see it in the revision history modal (and as a thread activity item in the thread), but you think it would be better to have it as part of the context panel?

mhjb:
Yeah — I think an updated context panel is usually a really significant event in a discussion thread. And particularly in a massive one like this, any individual activity item gets lost / is hard to find.

HSalmon:
The way I'm currently thinking about this means that it’s hard to make the whole line of text a link, seeing as it is actually a button, and we need to use a directive to format the date properly. Do you have thoughts as to how to solve this @gdpelican, @robguthrie? Apologies if the solution is staring me in the face...

mhjb:
How about like this:

![image](https://cloud.githubusercontent.com/assets/1380991/14663854/4b866cc8-0719-11e6-81fe-d16ec7fccd2a.png)

mhjb:
Relevant conversation happening in another Enspiral thread:

![image](https://cloud.githubusercontent.com/assets/1380991/14735985/4ee92242-08c8-11e6-9d77-a41d9913939d.png)

@alannallama does your facilitators' paradise vision touch on this?

HSalmon:
@mhjb, do you mind if I close this for now? We'll be revisiting it soon. Perhaps you could put a card in inbox on the trello board?

PR #3244, opened 2016-04-21
Include updated at date for context panel

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3227

screen shot 2016-04-21 at 11 54 50 am
screen shot 2016-04-21 at 1 20 06 pm


HSalmon:
Might need a smoke test. @mhjb - maybe when you've got a spare moment? No urgency

gdpelican:
I'll grab this one.

gdpelican:
So I don't think we'll be guaranteed to have the latest version, since it's serialized out with the event. A test might be

- Make a thread
- Change the context
- Make a page's worth (30) comments
- The thing will (I think) be broken

We could look at serializing out the most recent version with the discussion page, or firing off another web request for it? I dunno I'm not too inspired by this change, especially given that it adds more network overhead /complexity to the most-visited page in the app. Thoughts @robguthrie ?

robguthrie:
Yea.. I think we need to discuss architecture for this one. A discussion will need to serialize out the last  version record by default if we do this. That's possibly a n+1 thing which I'm not sure we want to do right now.

mhjb:
Sing out if any input from me would be useful.

HSalmon:
Happy to leave this until we've had an architecture discusssion :)

robguthrie:
Let's close this for now.

Issue #3208, opened 2016-04-14
link to the wrong locale on the Help card articles 


natiloomio:
for .es and .fr link have been changed in transifex but it still links to the english version on the blog

HSalmon:
Sorry Nati, I think I misled you by accident. It looks like the links are hardcoded, I can update this today!

HSalmon:
I think we've fixed this now @natiloomio!

PR #3245, opened 2016-04-21
Link help card articles to correct locale on blog

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3208

### DIFF
--- a/angular/core/services/user_help_service.coffee
+++ b/angular/core/services/user_help_service.coffee
@@ -18,3 +18,15 @@ angular.module('loomioApp').factory 'UserHelpService', ($sce, CurrentUser) ->
 
     helpVideoUrl: ->
       $sce.trustAsResourceUrl(@helpVideo())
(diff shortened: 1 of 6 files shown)


angular/core/services/user_help_service.coffee
```diff
@@ -18,3 +18,15 @@ angular.module('loomioApp').factory 'UserHelpService', ($sce, CurrentUser) ->
 
     helpVideoUrl: ->
       $sce.trustAsResourceUrl(@helpVideo())
+
+    tenTipsArticleLink: ->
+      switch CurrentUser.locale
```
gdpelican: Not for this PR, but I'm seeing a pattern of hard coding urls in the client here, and think this wants to get pulled out into the AppConfig if it gets any more complex 💛

gdpelican:
Looks good to me! 🍏

Issue #3243, opened 2016-04-21
Previous proposals dates / order incorrect


mhjb:
![image](https://cloud.githubusercontent.com/assets/1380991/14694962/af1e9956-07c1-11e6-8c45-e7c57cc334cf.png)

E.g. the first proposal actually closed 17 days ago. When I clicked it, it resorted itself down the page:

![image](https://cloud.githubusercontent.com/assets/1380991/14694984/d0233396-07c1-11e6-861c-4300c1f60bd2.png)

PR #3246, opened 2016-04-21
Fix previous proposals closed at date

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3243


gdpelican:
Hey @HSalmon this bug was from an oversight in adding `closed_or_closing_at` to the MotionSerializer, but not the ClosedMotionSerializer; I've put that in here: #3248 ❤️

PR #3250, opened 2016-04-21
Update bootstrap task to work with new gulp build system

gdpelican:


gdpelican:
I just had a real nice time spinning up the Pirate party's version of Loomio with this. I reckon it's handy and I'd like to keep it around.

lib/tasks/bootstrap.rake
```diff
@@ -42,14 +34,10 @@ namespace :bootstrap do
 
   desc 'Create user (optional arguments email and password)'
   task :create_user, [:email, :password] => :environment do |t, args|
-    args.with_defaults(email: 'default@loomio.com', password: 'bootstrap_password')
+    args.with_defaults(email: 'default@loomio.org', password: 'bootstrap_password')
```
robguthrie: can we please use admin@example.org and generate a password which is puts'ed to the screen at the end?

robguthrie:
Also where we say "View this digital ocean guide for postgresql", seeing as we're assuming macosx, can we just put "Run brew install postgresql"? As that's all that needs to happen.. the permissions stuff in the digital ocean article is unnecessary

gdpelican:
I've just updated the script to be more automatic and assume OSX. I have a hunch that if someone tries to use this and cares about windows / Linux, they'll let us know / update it for us 🎌

PR #3251, opened 2016-04-21
Angular sign in

gdpelican:
Hey this totally works!


gdpelican:
But of course this has to be reviewed and go in after beta is decommissioned.

robguthrie:
did you get oauth support in there?

gdpelican:
Yep

gdpelican:
Looking ahead a bit to getting this one in:

The major change here is that, since the user can now change within a single angular session, it needs to be a method (`User.current()`), rather than a static variable (`CurrentUser`). Seem like ok naming @robguthrie?

robguthrie:
Hrrmm.. if you're happy with it. Seems a bit weird , but also practical, to
put current user into the User model like that.

gdpelican:
That's why the naming's weird; I renamed `CurrentUser` -> `User`, which is
different from `UserModel`. Maybe it should be called something else like
`Session.user()`?

robguthrie:
Session.user() seems good to me.

PR #3255, opened 2016-04-21
Add proposal writing tips to new proposal form; language tweaks

mhjb:
see https://www.loomio.org/d/jWEAarxg/solution-candidate-hints-for-proposals

looks like:

image

and 

image

@HSalmon @natiloomio care to smoke test?


HSalmon:
It would be good to make sure that this takes into account the user's locale (that blog post has been translated into both French and Spanish). We do this for the group help card: https://github.com/loomio/loomio/tree/master/angular/core/components/group_page/group_help_card.

mhjb:
Great point! Done now.

gdpelican:
🍏  This looks great.

PR #3263, opened 2016-04-25
Add quickstart development guide to README

gdpelican:
I realized we don't really have a good 'this is how to dev on loomio' article anymore; the existing one still has lineman in it.

Also, I've updated the rake bootstrap task and had a good time setting up my own instance the other day, so I'm going with it.


robguthrie:
Could you please take out the old one, and point people to this one from the main readme?

gdpelican:
Right now it's

main README.md -> quickstart.md -> existing dev handbook

There's still some useful stuff in the existing handbook, like installing postgres and rbenv and linux packages you need. Maybe I'll just go through and clean that up now; I had in mind calling that the 'Advanced Development Setup Guide' or something.

robguthrie:
I really want to clean up the docs and have everything be succinct and
clear. I usually find that when documentation is too long it's because it's
unnecessary, eg: there is a bit of unnecessary stuff in the postgres setup
stuff. I'd like to have just one document we can point people to when they
ask how to setup their dev env.

gdpelican:
Okay I summarized the old one at the bottom of the quickstart guide, and took the existing handbook out. (I like to think that folks who are going to develop on Loomio will be able to get ruby, node, and postgres installed on their own)

robguthrie:
Thanks for that.  My take is a bit different.. I think we can easily
explain how to setup those things on macosx and ubuntu. It takes people heaps
of time.

gdpelican:
I don't think that belongs in the quickstart guide, which should ideally be incredibly bite sized. If we want to put that in, I think it should go in a separate document.

PR #3270, opened 2016-04-27
Angular sign in

gdpelican:


robguthrie:
How's this look on mobile? Is it up somewhere I can see?

gdpelican:
It's deployed to loomioclone.org at the moment; there's still a race-condition-y bug I think related to cookies which results in failed logins sometimes. Will raise a proposal about smoke testing the thing tomorrow.

http://g.recordit.co/QQBdRLrnd2.gif

gdpelican:
The race condition here looks like this, here we see the cookies of a 'discussions/dashboard' calls after 2 different 'successful' login attempts.

screen shot 2016-05-05 at 3 42 12 pm
screen shot 2016-05-05 at 3 42 16 pm

You can see that the failed one isn't returning a 'signed_in' or 'XRSF_TOKEN' cookie, even though the login request to the server returned successfully, which results in an interface that appears logged in (with current user data imported), but unable to make queries to the API without getting `forbidden` responses.

As far as I can tell, this is intermittent behaviour which happens ~20% of the time on loomioclone. Still investigating...

robguthrie:
what if you turn off xsrf protection?

gdpelican:
Good thought, I'll give that a go.

app/helpers/protected_from_forgery.rb
```diff
@@ -1,18 +1,18 @@
...
-    base.after_filter :set_csrf_cookie_for_ng
+    base.after_filter :set_xsrf_token
   end
 
   protected
-  
+
   def verified_request?
-    super || form_authenticity_token == request.headers['X_XSRF_TOKEN']
+    super || form_authenticity_token == request.headers['X-XSRF-TOKEN']
```
gdpelican: Alright, I think this was our trouble here; the request stores the token in `X-XSRF-TOKEN`, rather than `X_XSRF_TOKEN`

I haven't been able to repro the above issue on clone at all after changing this; I was following along with this thingy: https://stormpath.com/blog/angular-xsrf

angular/core/components/group_page/group_theme/group_theme.coffee
```diff
@@ -12,7 +12,7 @@ angular.module('loomioApp').directive 'groupTheme', ->
       { 'background-image': "url(#{$scope.group.coverUrl()})", 'z-index': (-1 if $scope.compact) }
 
     $scope.isMember = ->
-      CurrentUser.membershipFor($scope.group)?
+      Session.current().membershipFor($scope.group)?
```
robguthrie: At first sight, I find it annoying that this is not `Session.user()` I thought that's what you were going to go with.

angular/core/services/session.coffee
```diff
@@ -0,0 +1,28 @@
+angular.module('loomioApp').factory 'Session', ($rootScope, Records, AppConfig) ->
+
+  login: (data) ->
+    return unless data.current_user and data.current_user.id
+    data.users = data.users || []
+    data.users.push(data.current_user)
+    Records.import(data)
```
robguthrie: Oh hi there.

app/helpers/angular_helper.rb
```diff
@@ -40,7 +40,13 @@ def app_config
         groupThreads:      ENV['GROUP_PAGE_SIZE'],
         threadItems:       ENV['THREAD_PAGE_SIZE'],
         exploreGroups:     ENV['EXPLORE_PAGE_SIZE'] || 10
-      }
+      },
+      oauthProviders: [
```
robguthrie: nice.

angular/core/components/group_page/group_theme/group_theme.coffee
gdpelican: Oh yeah I can swap that over, Session.user() is better.

angular/core/components/group_page/join_group_button/join_group_button.coffee
```diff
@@ -1,13 +1,13 @@
 angular.module('loomioApp').directive 'joinGroupButton', ->
-  scope: {group: '='}
+  scope: {group: '=', block: '=?'}
```
robguthrie: So what is block here?

gdpelican: I've repurposed the join group button to be used at the bottom of threads you can't comment on:

screen shot 2016-05-10 at 10 20 30 am

The only change needed there, styling wise, is to add `btn-block` to the button, which unfortunately can't be done from outside the directive without css hackery.

gdpelican: ...wait, let me look at that again, we might not be using that anymore.

gdpelican: Ack sorry, I'm jumping to conclusions again. We _do_ need this for when you're

- logged in
- viewing a public discussion in a group you're not a member of.

That's when you get this:
screen shot 2016-05-10 at 10 32 09 am

gdpelican:
I'm about ready to merge this. Any burning qualms @mhjb @robguthrie ?

mhjb:
Nope (though I haven't retested)

robguthrie:
na! all g

Issue #3169, opened 2016-03-30
Last proposal should be open when you arrive at a thread


mhjb:

gdpelican:
Hrm, seems like a regression at some point. I'll take this one at some point.

gdpelican:
Well, I haven't done this yet. Maybe you could grab it @HSalmon ?

HSalmon:
Sure

HSalmon:
I think maybe this is fixed. I can't reproduce. Has anyone experienced this recently? @mhjb?

mhjb:
Happening for me right now on
https://www.loomio.org/d/AyTkIcZC/who-does-loomio-want-as-enspiral-venture-stewards-?proposal=QPahfxm9

Ah a clue: that "?proposal=QPahfxm9" was trailing on my urls from some
earlier visit to perhaps a previous proposals page.

gdpelican:
When you specify the proposal in the url, it's opened. If you take the
proposal=key off, it should work alright.

gdpelican:
So I guess the fix is to clear the ?proposal=key after visiting a page.

HSalmon:
@mhjb, I don't understand, in that particular example the proposal _should_ be expanded

mhjb:
Yeah. If it's helpful, steps to reproduce are:

* visit a proposal via e.g. a closing soon email
* browse to a thread where the most recent proposal is closed (url will still include link to first proposal)

PR #3273, opened 2016-04-27
Remove query string after requested model id set

HSalmon:
Supposed to address issue discussed here: https://github.com/loomio/loomio/issues/3169

Not sure if this is the best implementation. @gdpelican, would be good to get your thoughts


angular/core/components/thread_page/thread_page_controller.coffee
```diff
@@ -3,6 +3,7 @@ angular.module('loomioApp').controller 'ThreadPageController', ($scope, $routePa
 
   @requestedProposalKey = $routeParams.proposal or $location.search().proposal
   @requestedCommentId   = parseInt($routeParams.comment or $location.search().comment)
+  $location.url($location.path())
```
gdpelican: Heh, weird. but looks good.

gdpelican:
🍏

Issue #3274, opened 2016-04-27
Coordinators should appear first on memberships page


HSalmon:
This was implemented but it seems to have been recently reverted

gdpelican:
Weird. Might have been a merge conflict or something that got overwritten;
should be able to orderBy:"-admin" to put it back in.

PR #3277, opened 2016-04-27
List coordinators first for admin and non-admin memberships panel

HSalmon:
Looks like they were only ordered with coordinators first on the admin_memberships_panel, not the memberships_panel. Have updated.

Addresses: https://github.com/loomio/loomio/issues/3274


gdpelican:
Looks good to me 🍏

PR #3282, opened 2016-04-28
remove comment_counts stuff

robguthrie:


robguthrie:
@gdpelican ready for smoke test.

gdpelican:
A quick search for `itemsCount`
screen shot 2016-04-28 at 11 04 47 pm

and `readItemsCount`
screen shot 2016-04-28 at 11 05 10 pm

reveals that they're only used in `unreadItemsCount`, which:
screen shot 2016-04-28 at 11 06 13 pm

isn't used on the client side at all.

Do you think we can get rid of these as well?

gdpelican:
This PR's a pretty solid improvement, so I have no problems with merging it 🍏 
I do think we can do more to simplify the concept of unread, though.

PR #3289, opened 2016-04-29
locales update

natiloomio:


config/locales/client.cs.yml
```diff
@@ -1,546 +0,0 @@
-cs:
```
gdpelican: Er, why are we removing the czech client translations?

gdpelican:
Can't merge this yet as the build's erroring because client.cs.yml is missing; any idea why that is @natiloomio @robguthrie ?

natiloomio:
No idea!

--

gdpelican:
Okay, well I'm going to close this and hope the next one goes better 🙈

PR #3290, opened 2016-04-30
one layout, stylesheet for all emails

robguthrie:


app/views/thread_mailer/new_discussion.text.haml
```diff
@@ -4,5 +4,3 @@
 = '-------------'
 \
 = @discussion.description
-\
-= render 'footer', link: discussion_url(@discussion, @utm_hash)
```
gdpelican: Why no more footer in the text email?

app/views/thread_mailer/new_vote.html.haml
```diff
@@ -1,7 +1,2 @@
-= render '/application/mailer_doctype'
-%html
-  %head
-    = stylesheet_link_tag 'email'
-  %body
-    = render "vote", vote: @vote
-    = render "footer", link: motion_url(@motion, @utm_hash)
+= render "vote", vote: @vote
+= render "footer", link: motion_url(@motion, @utm_hash)
```
gdpelican: I'm wondering if we could pull out the footer into the layout as well, and render it iff we provide a link in the action?

```
def new_motion
  ...
  @footer_link = motion_url(@motion, @utm_hash)
end
```
```
render 'footer', link: @footer_link if @footer_link
```
💛

app/views/application/_mailer_doctype.html.haml
```diff
@@ -1 +0,0 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
```
gdpelican: I want to note that we're moving from 4.0 transitional to an html 5 header here. The little research I did said 'Stick with the lower header unless you plan on using HTML5 elements. Also it doesn't matter that much because most email clients ignore it and do whatever they want.' It was also from a few years ago, so I imagine HTML is more the expected standard by now. I'm guessing it's fine.

http://blog.mailermailer.com/email-design/html5-doctype-in-email-too-soon
http://stackoverflow.com/questions/34319889/doctype-for-html-email

gdpelican:
Thanks @robguthrie , this'll make working with emails that much easier and less scary. 🍏

app/views/thread_mailer/new_discussion.text.haml
robguthrie: mistake. cheers for that.

app/views/thread_mailer/new_vote.html.haml
robguthrie: I'm not keen on doing any more work on this ATM, but that is a good idea.

app/views/application/_mailer_doctype.html.haml
robguthrie: Yea.. My reading told me that there is very little difference between 4.0t and 5, and it was quite old. I like the simplicity of the 5.

robguthrie:
The main benefit is we now have a super clear understanding of our email stylesheets.. which broke after the betabyebye. This fixes yesterday on loomio styles

Issue #3292, opened 2016-05-01
Disable thread form after submit


rdbartlett:
Somehow I started the same thread three times. I think I accidentally hit the keyboard shortcut to start the thread, then later I realised it had done the same thread 3x

![image](https://cloud.githubusercontent.com/assets/970124/14944250/85c1e01c-1041-11e6-8592-c35180858128.png)

PR #3294, opened 2016-05-01
Don't submit form again on cmd+enter if already submitted

gdpelican:
#3292


gdpelican:
Okay with this @robguthrie ?

robguthrie:
oooh great. yes. 🍏

PR #3301, opened 2016-05-02
Add reference to replied-to-comment in comment replies

HSalmon:
Thought this might be a good interim thing to implement while we consider more complicated options.

screen shot 2016-05-02 at 2 56 43 pm

[Trello card](https://trello.com/c/7whXZS1b/1326-add-reference-to-replied-to-comment-in-comment-replies)
[Loomio discussion](https://www.loomio.org/d/lCfoxy9v/ux-replies-to-comments)

### DIFF
--- a/angular/core/components/thread_page/activity_card/new_comment.haml
+++ b/angular/core/components/thread_page/activity_card/new_comment.haml
@@ -15,10 +15,12 @@
     .media-left
       %user_avatar{user: "comment.author()", size: "small"}
     .media-body
-      %h3.new-comment__in-reply-to{ng-if: "comment.parentId", id: "event-{{event.id}}"}
+      %h3.new-comment__replied-to{ng-if: "comment.parentId", id: "event-{{event.id}}"}
         %strong
           %a{lmo-href-for: "comment.author()"}> {{ comment.authorName() }}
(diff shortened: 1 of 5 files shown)


gdpelican:
Hm, this will run into a problem when you're replying to something out of the current event window.

screen shot 2016-05-03 at 11 25 19 am
(note how the link is undefined because comment.parent() isn't loaded)

We'll have to deal with this case somehow, as I don't think it'll be all that uncommon.

robguthrie:
I seem to remember it was an AMS bug keeping us from serializing the parent
by default. Can we revisit if serializing the parent comment is possible
now?

HSalmon:
I can take a look at that @robguthrie

gdpelican:
Even given that the parent comment is loaded, we get into weirdness with
trying to load and manage multiple windows of events here. I think we need
to ensure the page is reloaded here if the target comment is outside the
window.

robguthrie:
Yea.. I just don't think the effort is worth it right now. I think to do
this properly we need to be able to dynamically load a new window of
items.. and that is a bit bigger than the current scope. I'd rather not
spin cycles on this.

gdpelican:
Gonna close, and we'll revisit laters.

PR #3306, opened 2016-05-02
[WIP] first go at nicer invite  email

mhjb:
A little iteration to the invite email I mucked about with last night:

image

### DIFF
--- a/app/views/invite_people_mailer/to_join_group.html.haml
+++ b/app/views/invite_people_mailer/to_join_group.html.haml
@@ -1,11 +1,13 @@
 %p
+  = @invitation.inviter.name
   = t(:"invitation.link_help", group_name: @invitation.invitable_name)
-  %br
-  = link_to "#{invitation_url(@invitation)}", invitation_url(@invitation, @utm_hash)
+
 %p
-  = simple_format @message_body
+  = link_to image_tag('http://i.imgur.com/FWXRc8c.png'), invitation_url(@invitation, @utm_hash)
+
+%p
+  %em= simple_format @message_body
 %p
   = t(:"invitation.value_proposition")
 %p
   = t(:"invitation.thanks")
-= @invitation.inviter.name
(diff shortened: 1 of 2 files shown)


gdpelican:
Sweet!

PS, if you wanted to try out sprinkling some emojis here or there, I believe you could do so using the `render_rich_text` method,

```
render_rich_text(":heart: from team Loomio")
```

robguthrie:
Code wise I think the button should be done with CSS (we have automatic inlining so should not be a problem). If an image must be used, then it should have ALT text.

Design wise, I want to see something quite a lot more styled. A frame, with all the text styled. Something that feels like it comes from a million dollar company with a nice design asthetic.

robguthrie:
Overall, I think this should be designed in sketch first.

robguthrie:
Here is a site Rich found with awesome email examples:

http://reallygoodemails.com/

robguthrie:
The trello invitation email seems like good inspiration. A reusable template, a simple call to action

http://reallygoodemails.com/social/invited-you-to-join/

mhjb:
I was interested to experiment with what could be done without going all the way to (I understand time-consuming-to-implement) proper templating system. 

I too would ultimately like to see us get to properly designed lovely emails. And I understand that's what the spike in this sprint is about investigating. But depending on what comes out of that investigation it _might_ be worth going for a little iteration like this first.

robguthrie:
At this stage I think it's worth going all the way for a nice design. The email template stuff in this sprint is not a spike as I understand it.

mhjb:
Oh cool. Sorry — I was misremembering re. the spike.

robguthrie:
Ohhh great. Yea.. I'm super excited to see a simple design like the trello one. It is not super hard to implement a single nice template, and the costs of smaller changes from non full time devs are quite high, as both need careful attention to quality detail. (eg: the image issue, a11y etc)

robguthrie:
FYI - matthew and I just had a good in person chat about this.

PR #3314, opened 2016-05-03
Change from roadie to premailer

robguthrie:
Roadie has been failing to inline styles, it also had other issues, so I'm having a go with premailer.


robguthrie:
I've smoketested this on loomioclone. Permission to merge @gdpelican

gdpelican:
Down wih roadie! 🍏

Issue #3309, opened 2016-05-03
In reply to blank


rdbartlett:
![image](https://cloud.githubusercontent.com/assets/970124/14973408/7ab74022-113b-11e6-9b30-ecac304f36cf.png)

Discussion id: G0BegSO8

HSalmon:
I'm pretty sure that this is because the author of the comment you are replying to has an apostrophe in their name - testing this locally seems to prove that thinking. However, I'm not 100% sure what the best way to fix it is. @robguthrie? @gdpelican?

robguthrie:
I suspect it's something about how we use Angular Translate. Check this
out: https://angular-translate.github.io/docs/#/guide/19_security

HSalmon:
And it was! Thanks @robguthrie. PR over there ^^^

HSalmon:
Should be fixed!

PR #3315, opened 2016-05-03
Update translate-values attribute to fix reply to bug

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3309

Look! No errors:

screen shot 2016-05-03 at 4 04 03 pm


gdpelican:
We got this! 🍏

PR #3323, opened 2016-05-04
Allow add to homescreen on Android

gdpelican:
TODO: Add dummy service worker to appease the chrome gods. (update: now done)
https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android#span-idcriteria-faqfrequently-asked-questionsspan


gdpelican:
Hannah totally smoke tested this, and it was great.

angular/core/app.coffee
```diff
@@ -43,6 +43,9 @@ angular.module('loomioApp', ['ngNewRouter',
 angular.module('loomioApp').controller 'ApplicationController', ($scope, $filter, $rootScope, $router, KeyEventService, ScrollService, CurrentUser, BootService, AppConfig, ModalService, ChoosePlanModal, AbilityService) ->
   $scope.isLoggedIn = AbilityService.isLoggedIn
 
+  if document.location.protocol.match(/https/) && navigator.serviceWorker?
```
gdpelican: We're adding a service worker here, not because we need one, but because chrome requires it for your thing to be seen as 'an app'.

One the plus side, now I know a little bit more about workers and how we can leverage them towards nifty things like push notifications in the future.

PR #3328, opened 2016-05-05
Resend invitations

HSalmon:


app/models/invitation.rb
```diff
@@ -26,6 +26,8 @@ class InvitationAlreadyUsed < StandardError
   scope :pending, -> { not_cancelled.single_use.where(accepted_at: nil) }
   scope :shareable, -> { not_cancelled.where(single_use: false) }
   scope :single_use, -> { not_cancelled.where(single_use: true) }
+  scope :ignored, -> (send_count, age_in_hours) { pending.where('send_count = ? AND created_at < ?', send_count, age_in_hours.hours.ago) }
+  scope :too_old_to_send, -> { pending.where('created_at < ?', '2016-05-05 00:00:00 +1200') }
```
gdpelican: This doesn't appear to be used? ❔

app/mailers/invite_people_mailer.rb
```diff
@@ -11,14 +11,15 @@ def to_start_group(invitation:,
                      subject_params: {group_name: @invitation.invitable_name}
   end
 
-  def to_join_group(invitation: , locale: )
+  def to_join_group(invitation:, locale:, subject:)
```
gdpelican: I'd default this to

```
def to_join_group(invitation:, locale:, subject: "email.to_join_group.subject")
   ...
end
```

so that we don't have to change the existing calls at all, but can override where we need to.

app/mailers/invite_people_mailer.rb
```diff
@@ -11,14 +11,15 @@ def to_start_group(invitation:,
...
     @invitation = invitation
+    invitation.increment_send_count!
```
gdpelican: Should we wait to increment the send count until after the `send_single_mail` call?

config/locales/server.en.yml
```diff
@@ -28,6 +28,8 @@ en:
       link: Join the discussion
     to_join_group:
       subject: "%{member} has invited you to join %{group_name} on Loomio"
+    resend_to_join_group:
+      subject: "%{member} has invited you to join %{group_name} on Loomio"
```
gdpelican: Probably want to update this to something different 🔴

db/migrate/20160505005253_add_send_count_to_invitations.rb
```diff
@@ -0,0 +1,5 @@
+class AddSendCountToInvitations < ActiveRecord::Migration
+  def change
+    add_column :invitations, :send_count, :integer, null: true, default: 0
```
gdpelican: I'm pretty sure we want an index on this column; maybe a partial index as described in this article:
http://rny.io/rails/postgresql/2013/08/20/postgresql-indexing-in-rails.html

spec/services/invitation_service_spec.rb
```diff
@@ -44,4 +44,38 @@
       expect(Event.last.kind).to eq 'invitation_accepted'
     end
   end
+
+  describe '#resend_ignored' do
```
robguthrie: holy shit these specs are incredible. 100 points A+++

db/migrate/20160505005253_add_send_count_to_invitations.rb
robguthrie: great article!

Seems like the most important index will be `accepted_at`.

```
add_index :invitations, :accepted_at, where: 'accepted_at IS NULL'
add_index :invitations, :send_count, where: 'accepted_at IS NULL'
```

Also worth nothing that we could combine them (the indices into 1) but I don't think there is a lot of benefit to that.

app/services/invitation_service.rb
```diff
@@ -27,11 +27,20 @@ def self.invite_to_group(recipient_emails: nil,
...
       InvitePeopleMailer.delay.to_join_group(invitation: invitation,
-                                             locale: I18n.locale)
+                                             locale: I18n.locale,
+                                             subject: "email.to_join_group.subject")
       invitation
     end
   end
 
+  def self.resend(invitation)
+    return if invitation.accepted_at || invitation.cancelled_at || !invitation.send_count
```
robguthrie: I think this should be moved into the invitation model

```
return unless invitation.is_pending?
```

Also .. why the sent_count check?

app/models/invitation.rb
```diff
@@ -53,6 +55,10 @@ def to_join_group?
     intent == 'join_group'
   end
 
+  def increment_send_count!
```
robguthrie: Turns out rails already has this.

remove this method and try using

```
invitation.increment!(:send_count)
```

app/mailers/invite_people_mailer.rb
robguthrie: can we please call it subject_key?

app/models/invitation.rb
robguthrie: If legacy invitations have a 0 or negative send_count then we don't need this.

db/migrate/20160505005253_add_send_count_to_invitations.rb
gdpelican: `accepted_at` is probably a good index too; I was thinking of doing `WHERE send_count IS NOT NULL` to ignore all of the heaps of existing invitations we have.

robguthrie: whatddya reckon about send_count being null: false, default 0? We don't really need that null value for legacy invites, they can be 0's

gdpelican: Yeah I was kinda wondering that too.

PR #3329, opened 2016-05-05
Add location to users and groups

robguthrie:


app/jobs/locate_groups_job.rb
```diff
@@ -0,0 +1,27 @@
+class LocateGroupsJob < ActiveJob::Base
+  def perform(last_active)
+    group_ids = Discussion.select('DISTINCT(group_id) group_id, last_activity_at').
+                 where('last_activity_at > ?', last_active).
+                 order('last_activity_at desc').map(&:group_id)
+
+
+    Group.where(id: group_ids).each do |group|
```
gdpelican: Am I right in sniffing that this is the most common country/region/city amongst members of a particular group? ❔

db/schema.rb
```diff
@@ -807,6 +810,9 @@
     t.boolean  "email_on_participation",           default: true,       null: false
     t.integer  "default_membership_volume",        default: 3,          null: false
     t.boolean  "has_muted",                        default: false,      null: false
+    t.string   "country"
```
gdpelican: I also wonder if it might be cool to have users be able to specify (or at least view?) their location on their profile

app/jobs/locate_groups_job.rb
robguthrie: yes. for user it is the last location, for group it is the most common

robguthrie:
ok this works now. I've run it on clone for a set of recently active groups. I'll add a rake task in the morning.

db/schema.rb
robguthrie: Nice Idead. I'm excited about including location in /explore

robguthrie:
@gdpelican I've done a bit more work. Can you review again please?

app/jobs/send_missed_yesterday_email_job.rb
```diff
@@ -1,12 +1,10 @@
-class SendMissedYesterdayEmail
-  def self.to_subscribers!
+class SendMissedYesterdayEmailJob < ActiveJob::Base
+  def perform
     zones = User.pluck('DISTINCT time_zone').select do |zone|
-      Time.find_zone(zone) &&
-      DateTime.now.in_time_zone(zone).hour == 6
+      Time.find_zone(zone) && DateTime.now.in_time_zone(zone).hour == 6
```
gdpelican: Could we make this hour configurable, perhaps via ENV?

gdpelican:
Looks pretty good to me. 🍏

app/jobs/send_missed_yesterday_email_job.rb
robguthrie: Seems like the easiest win for a new contributor with that need.

robguthrie: but fair call. but not this time thanks :)

robguthrie:
testing on loomio-clone now

PR #3332, opened 2016-05-05
Ensure all img links are relative to root url

gdpelican:
Because there are some browsers out there not smart enough to account for "img/image.jpg" as a url.


robguthrie:
choice 💚

PR #3338, opened 2016-05-05
Add info about blockquotes to markdown page

HSalmon:


robguthrie:
Hannah, can you remove the "you need to put a line between lines or it will
be a big paragraph" thing.. it's not true for our markdown anymore.

HSalmon:
@robguthrie, I can't get the block quotes to render properly so James suggested that I move the whole thing over to angular, which sounds good to me. It would mean creating a markdown modal. Does that sound alright to you?

robguthrie:
I would suggest just refusing to include blockquotes, and calling the job
done. This is not a super great thing to spend time on. We know that it's
not going to help most users anyway, who need WYSIWYG or atleast a toolbar.

PR #3340, opened 2016-05-06
Refactor group welcome modal to use experiences

HSalmon:
[Trello card](https://trello.com/c/RG3nYvyr/1499-8-refactor-the-welcome-to-group-modal-to-use-experiences)


gdpelican:
We've got one error, which is that if you visit a group, go to another page, and return to the group (without loading the whole app), you shouldn't see the welcome modal.

HSalmon:
@robguthrie, this one could do with a CR

angular/core/components/group_page/group_page_controller.coffee
```diff
@@ -62,11 +62,11 @@ angular.module('loomioApp').controller 'GroupPageController', ($rootScope, $loca
     @group.noInvitationsSent() and
     !@group.trialIsOverdue() and
     !@subscriptionSuccess and
-    !GroupWelcomeModal.shownToGroup[@group.id]
+    !CurrentUser.experiences["seenWelcomeFor#{@group.key}"]
```
robguthrie: I think this should be CurrentUser.hasExperienced('welcomeFor#{group.key}'). So that we abstract the implementation.

robguthrie:
Looks great. Just blimmin' great. I would like it if you implemented the last suggestion.. a query method to tell if a user has had an experience on the client.. 🍏  either way though.

HSalmon:
Sweet, I'll implement that now. Thanks :)

HSalmon:
Is this all good @robguthrie?

robguthrie:
Oh! It's looking fine indeed. 

But there is another consideration: migration of existing users. We don't want everyone to have to go through this experience for every group when they're already familiar with them.

I kind of think this might be a regular pattern - these kinds of users don't need to get the experience.

I'm wondering what the best approach is: do we migrate every membership in the database to have the experience already, or do we add a condition that this is only for memberships that are newer than the date of the deploy?

gdpelican:
I'd lean towards the former; I'd rather not have a series of deployed at dates floating around in the codebase.

The other question is that raises for me is, "Do some experiences actually belong on the Membership, rather than the user?" ie, rather than having heaps of 'welcomedToGroupKey' experiences on the user, does it make sense to track experiences at a membership level as well?

This would cover things like
- Has this user commented in this group?
- Have they viewed the group / read its description? Or interacted only via email?

Whereas user experiences would be
- Has this user uploaded a profile picture?
- Contacted us for help before?
- Ever started a group?
etc.

robguthrie:
Yeah.. I think you're right GD. 

Membership experiences mean we don't need to encode groupId into the user experience, which I think might make things tricky later on.

HSalmon:
@gdpelican, @robguthrie, just added experiences to memberships, would appreciate feedback :)

app/controllers/api/memberships_controller.rb
```diff
@@ -67,6 +67,13 @@ def set_volume
     respond_with_resource
   end
 
+  def save_experience
+    raise ActionController::ParameterMissing.new(:experience) unless params[:experience]
+    load_resource
+    service.save_experience membership: @membership, actor: current_user, params: { experience: params[:experience] }
```
gdpelican: Fun tip; `load_resource` returns the resource it loaded, so this can be

```
service.save_experience(membership: load_resource, ...)
```

app/models/membership.rb
```diff
@@ -67,6 +67,11 @@ def discussion_readers
       where('discussion_readers.user_id = ?', user_id)
   end
 
+  def experienced!(key, toggle = true)
```
gdpelican: Let's pull this out into a module, `HasExperiences`

```
module HasExperiences
  def experienced!
  end
end
```

```
include HasExperiences
```

gdpelican: ^^ (I could see us down the road interested in experiences for other things, like 'has a group made a decision together?', or 'has a network shared a proposal between groups?')

robguthrie:
Looks superb!

HSalmon:
Ready to go even?

gdpelican:
How shall we cover existing experiences?

gdpelican:
Migration to set all memberships to have 'welcomeModal: true' ?

robguthrie:
yea

HSalmon:
Tests are failing, but pass for both James and I locally

HSalmon:
@gdpelican, @robguthrie, could one of you take a final look at this to see whether it's ready to go?

HSalmon:
I suppose it should be tested on clone first.

robguthrie:
Smoke test passed super great

PR #3341, opened 2016-05-06
Update to Rails5 (not for lols anymore!) [WIP - lol]

robguthrie:
not ready.. need to merge configs and test and complete the migration (ApplicationRecord etc)


robguthrie:
http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html

robguthrie:
nice to know it's possible if we just drop simple_form

gdpelican:
Is the sign_in / sign_up forms the only place we use simple_form? I guess `/start_group` too

robguthrie:
nope.. just contact us

gdpelican:
A non-stab in the dark list:

- user sign in
- user sign up
- start group
- forgot password
- contact form
- network request form (should be removed)
- new invitation form (should be removed)
- admin themes form (<- do we need this?)

robguthrie:
whoa.. I cant read grep. Thanks for checking. I'll be happy to migrate those

gdpelican:
Since rails 5 was [officially released](http://weblog.rubyonrails.org/2016/6/30/Rails-5-0-final/) this week, it's time to get on this upgrade path. Unfortunately I ran into a snag with activeadmin:

```
Bundler could not find compatible versions for gem "actionpack":
  In Gemfile:
    activeadmin (~> 1.0.0.pre4) was resolved to 1.0.0.pre4, which depends on
      inherited_resources (~> 1.6) was resolved to 1.6.0, which depends on
        actionpack (< 5, >= 3.2)

    rails (~> 5.0.0) was resolved to 5.0.0, which depends on
      actionpack (= 5.0.0)
``` 

Although they're working on it, so once they're ready for Rails 5, we should be too.

gdpelican:
I've been struggling with some rate limiting problems that have cropped up with the upgrade, looks like they're coming from here: https://github.com/ankane/ahoy/issues/222

gdpelican:
Putting this on hold until we can find a more compelling reason to upgrade.

PR #3342, opened 2016-05-06
Resend ignored invitations

HSalmon:


app/services/invitation_service.rb
```diff
@@ -32,6 +32,14 @@ def self.invite_to_group(recipient_emails: nil,
     end
   end
 
+  def self.resend(invitation)
```
robguthrie: I think this should be `resend(invitation: )` maybe? so it's in line with the other methods?

HSalmon:
@robguthrie, James updated this. Do you have anything else to add?

app/services/invitation_service.rb
HSalmon: It's currently inline with ```redeem```...but if you'd prefer I can implement that now

app/models/invitation.rb
```diff
@@ -26,6 +26,7 @@ class InvitationAlreadyUsed < StandardError
   scope :pending, -> { not_cancelled.single_use.where(accepted_at: nil) }
   scope :shareable, -> { not_cancelled.where(single_use: false) }
   scope :single_use, -> { not_cancelled.where(single_use: true) }
+  scope :ignored, -> (send_count, age_in_hours) { pending.where('send_count = ? AND created_at < ?', send_count, age_in_hours.hours.ago) }
```
gdpelican: I'd slightly prefer this to be

```
pending.where(send_count: send_count).where('created_at < ?', age_in_hours.hours.ago)
```

...I think in the interest of writing as little SQL as possible. 💛

app/services/invitation_service.rb
gdpelican: I like it just because
```
.each { |invitation| resend invitation }
```

reads better than

```
.each { |invitation| resend invitation: invitation }
```

app/models/invitation.rb
gdpelican: I'd also slightly prefer, instead of `age_in_hours`, `since`

```
-> (send_count, since) { pending.where(sc: sc).where('created_at < ?', since) }
```

which takes out the slight magical number-ness of
```
resend_ignored(send_count: 1, age_in_hours: 24)
resend_ignored(send_count: 2, age_in_hours: 72)
```

and instead reads

```
resend_ignored(send_count: 1, since: 1.day.ago)
resend_ignored(send_count: 2, since: 3.days.ago)
```
?
💛

robguthrie: yeah.. seems neater.

app/services/invitation_service.rb
robguthrie: colour me convinced

db/migrate/20160505005253_add_send_count_to_invitations.rb
```diff
@@ -0,0 +1,5 @@
+class AddSendCountToInvitations < ActiveRecord::Migration
+  def change
+    add_column :invitations, :send_count, :integer, null: true, default: 0
```
robguthrie: I think we decided null was going to be false, ae?

gdpelican: Nice catch +1

db/migrate/20160505040422_add_partial_index_to_invitations_send_count.rb
```diff
@@ -0,0 +1,6 @@
+class AddPartialIndexToInvitationsSendCount < ActiveRecord::Migration
+  def change
+    change_column :invitations, :send_count, :integer, null: false, default: 0
```
HSalmon: Null was set to false in this subsequent migration

HSalmon:
@gdpelican, @robguthrie, I believe I've implemented all feedback at this point

gdpelican:
This looks great. Are there any migration concerns, @robguthrie, or are we confident that `send_count` will prevent people from getting flooded with old invitations when we deploy?

robguthrie:
I've just added a ResendIgnoredInvitationsJob class which is called by the rake task. Let's test this on clone and see all those emails that might go out if there was a bug.

robguthrie:
@HSalmon hope you don't mind.. I did the rakey bit of this for you. Have a look.

Hannah are you able to deploy this to LoomioClone?

HSalmon:
Thanks @robguthrie, I'm just in the middle of something else atm, but I can deploy to clone a bit later on. If you're up for it now though...

robguthrie:
@mhjb I've just become concerned that this is too many attempts. IE: I think 3 sends is over the top. In the cases where the user does want to ignore the invitation, it seems to me like a third email would be pretty annoying, and might make me want to mark as spam so as to not get any more emails.

That would be very bad.

I think the metrics say that most emails are accepted on their first day. Can we try just having 1 reminder email for now, rather than 2?

mhjb:
My suggestion would be to leave it as designed and see what feedback we get…

robguthrie:
Sorry @mhjb, but I don't feel good about that. If we send three emails to someone who does not want them, then they'll give mark as spam, not reply to an email address they dont have and say could you stop sending me emails.

If people click mark as spam then soon _all_ our emails don't get through. We have a responsibility not to send unwanted email and I think 3 attempts, when the problem was never that big in the first place, is a feature too far.

mhjb:
I bow to your greater understanding of wider email delivery considerations.

robguthrie:
@HSalmon I've removed the day 3 resend.

HSalmon:
Thanks @robguthrie, should I deploy to clone? Or are you up to deploying to clone?

robguthrie:
I think you should do it. `rake deploy loomio-clone resend-invitations`

robguthrie:
tested on clone √

mhjb:
@HSalmon what does it mean when send_count is 0?

PR #3344, opened 2016-05-09
Don't display previous proposals card link unless there are more propo…

HSalmon:
…sals to show

Addresses: https://github.com/loomio/loomio/issues/3334

Not sure we want to be serializing more stuff out but yeah, first go

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -424,6 +424,10 @@ class Group < ActiveRecord::Base
     Discussion.published.where(group_id: org_group_ids).sum(:motions_count)
   end
 
+  def closed_motions_count
+    motions.closed.count
+  end
+
   def org_group_ids
     [parent_or_self.id, parent_or_self.subgroup_ids].flatten
   end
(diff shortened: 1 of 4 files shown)


app/models/group.rb
```diff
@@ -424,6 +424,10 @@ def organisation_motions_count
     Discussion.published.where(group_id: org_group_ids).sum(:motions_count)
   end
 
+  def closed_motions_count
```
gdpelican: Can't do this, unfortunately, it'll be too slow. If we want this feature, we'll have to make a counter cache column for it, `closed_motions_count`, on the group model.

HSalmon:
And what do you reckon? Do we want this feature?

HSalmon:
@robguthrie, what do you reckon? Do we want this feature right now?

robguthrie:
Yeah. Seems like a nice thing to have. A counter cache is pretty easy to add.
We use custom_counter_cache for this kind of thing. It's actually pretty great.

If you add the counter, I'll help you write a migration SQL statement to populate it for existing groups.

gdpelican:
:+1:

PR #3346, opened 2016-05-09
More robust group page threads

gdpelican:


robguthrie:
Interesting. @gdpelican while you're in there, could you please change the behaviour of the page so that it immediately displays what we have in the RecordStore, rahter than waiting for the fetch to complete each time? 

Could also include loading symbol, but right now if we did an offline loomio, this page would not work and the other ones would mostly.

gdpelican:
I think it does that, but I'll make sure. Seems like the right thing to do.

gdpelican:
This is ready for smoke test / CR.

robguthrie:
CR'd: This is a really nice improvement @gdpelican. Thank you.

PR #3347, opened 2016-05-09
destroy user from admin panel for special circumstances

robguthrie:


HSalmon:
I tested this locally and it worked as expected 🍏

PR #3348, opened 2016-05-09
Don't allow comment input while submitting

gdpelican:
Fixes this issue:
http://g.recordit.co/ph9ByELSXs.gif

Which was occasionally causing drafts to get saved after submit, meaning they showed up after the comment was submitted.

(NB the only change here is adding `ng-disabled: "isDisabled"` to the comment form text area.)


robguthrie:
🍏  nice work

PR #3350, opened 2016-05-10
set default job priority to 10, prioritise mailers, invitations first

robguthrie:


robguthrie:
Just noticed we have a lot of Jobs in the DJ queue. Geocoding visits etc.. We need to prioritise invitation emails above these.

gdpelican:
I've noticed that the slack webhooks are going through a bit slowly too, wonder if that's contributing? What other things are in the DJ queue?

robguthrie:
Seems like there has been a spike of jobs. I'd need to install a DJ web interface to get a good understanding.. metabase seems to be choking on DJ queries.

gdpelican:
Anyway this looks good to me 🍏

PR #3351, opened 2016-05-10
Use loomio hosted emoji urls in email

gdpelican:


app/extras/emojifier.rb
```diff
@@ -2,6 +2,8 @@ class Emojifier
   extend JSRuntime
 
   def self.emojify!(text)
-    eval "emojione.shortnameToImage(\"?\")", text
+    root_url = Rails.application.routes.url_helpers.root_url(ActionMailer::Base.default_url_options)
```
gdpelican: This is pretty horrid. I'd be open to hearing thoughts on a better way to do this.

robguthrie: asset_host! :)

PR #3355, opened 2016-05-10
fix inline css in emails

robguthrie:


Gemfile
```diff
@@ -74,7 +57,7 @@ gem 'simple-rss'
 gem 'snorlax', '~> 0.1.3'
 gem 'custom_counter_cache'
 gem 'safe_shell'
-gem 'premailer'
+gem 'premailer-rails'
```
robguthrie: problem 1: we were using the wrong gem.

robguthrie: problem 2, there was an email.png which was being referenced rather than email.css

config/application.rb
```diff
@@ -6,6 +6,17 @@
 
 require_relative '../lib/version'
 
+def lmo_asset_host
```
gdpelican: I have found myself wanting access to this in other parts of the application as well

(namely the emojifier, to ensure that we're displaying images from our server in emails)
I guess with this I could then use `Rails.application.config.action_mailer.asset_host`?

robguthrie: Yep!

robguthrie:
ok to merge @gdpelican ?

gdpelican:
Green bean!

PR #3356, opened 2016-05-11
Don't allow logged out users to create groups

gdpelican:
Is this the right thing for this @robguthrie? 
https://errbit.loomio.io/apps/57215524657272000e000001/problems/5731bf2765727200130011f5


robguthrie:
yea, I guess! √

PR #3366, opened 2016-05-12
Don't publish multiple motion_closing_soon events

gdpelican:


app/models/motion.rb
```diff
@@ -50,6 +50,16 @@ class Motion < ActiveRecord::Base
   scope :closing_in_24_hours,      -> { where('motions.closing_at > ? AND motions.closing_at <= ?', Time.now, 24.hours.from_now) }
   scope :chronologically, -> { order('created_at asc') }
 
+  scope :closing_soon_not_published, -> {
+     voting
+    .joins("LEFT OUTER JOIN events e ON e.eventable_id = motions.id AND e.eventable_type = 'Motion'")
```
gdpelican: The logic here is "give me all of the voting motions where a `motion_closing_soon` event has not been published in the last 2 days".

I'm hypothesizing that there will be use cases where people see 'Oh, hm, that proposal closes in 24 hours, but I think we need longer', and bump it back for a few days, in which case we want to publish another motion_closing_soon event when it's about to close again.

If, however, a motion_closing_soon event was just published say an hour ago, we don't want another one just because the job ran twice.

spec/jobs/proposals_closing_soon_job_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+describe ProposalsClosingSoonJob do
+  let!(:motion) { create :motion, closing_at: 24.hours.from_now }
+  let(:subject) { ProposalsClosingSoonJob.new.perform }
+
+  describe 'perform' do
+    it 'publishes a proposal closing soon event' do
```
gdpelican: I think the tests spell out the intended functionality nicely.

- publishes a proposal closing soon event successfully
- does not publish for closed motions
- does not publish for motions that aren't closing soon
- does not publish for motions that just had a motion_closing_soon published
- publishes if a motion_closing_soon event was already published, but more than 2 days ago.

gdpelican:
Could really use a QA on this please @HSalmon @robguthrie

HSalmon:
I tried testing this locally by publishing the event (```Events::MotionClosingSoon.publish!(motion)```) but wasn't able to generate a notification...

gdpelican:
That might be the intended effect, since this now looks at whether there's been a motion closing soon event in the past couple days before publishing another one.

I'm, at this point, more concerned with thoughts on _how_ it should work; I'm pretty confident that this particular functionality works as written.

app/models/motion.rb
HSalmon: I think I understand the logic, however I'm wondering there's any chance we could still duplicate the events, or miss the opportunity to publish one. Trying to get my head around the intended functionality. Can you think of any cases that would mean we published too many/too few events?

robguthrie: This looks great to me.
And I think this is the right approach..

spec/jobs/proposals_closing_soon_job_spec.rb
robguthrie: You've been writing some really great tests recently.

robguthrie:
I think this is high quality work 🍏 📗 🍵 ♻️

PR #3380, opened 2016-05-23
A mobile app for Loomio [WIP]

gdpelican:


jpdeplaix:
Any progress on this ?

gdpelican:
It's not on my immediate radar, but we will be putting some real effort into this, likely early next year.

jpdeplaix:
Hi, what is the status of this ? Of course I noticed it has been closed but is it still on your TODO list ?

gdpelican:
As mentioned, it's not on the immediate radar, but we will be discussing options for tackling this in a way that fits with our strategy soon.

PR #3381, opened 2016-05-23
Use locale for logged out user

gdpelican:


mhjb:
Great! I've let the client know

Issue #3327, opened 2016-05-05
I expect to get a proposal closed email when I've set the thread email settings to "important activity"


rdbartlett:

rdbartlett:
supersedes #3304

PR #3382, opened 2016-05-24
Update UsersByVolumeQuery to include normal and loud for motion closed

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3327


gdpelican:
Seems solid to me! 🍏

PR #3384, opened 2016-05-24
do not load loomio_org plugins in Dockerfile

dachary:
Signed-off-by: Loic Dachary <loic@dachary.org>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ ENV DEVISE_SECRET boopboop
 ENV SECRET_COOKIE_TOKEN beepbeep
 
 # build assets
-RUN bundle exec rake deploy:build
+RUN bundle exec rake deploy:build[docker]
 RUN bundle exec rake assets:precompile
 
 EXPOSE 3000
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -37,7 +37,7 @@ ENV DEVISE_SECRET boopboop
 ENV SECRET_COOKIE_TOKEN beepbeep
 
 # build assets
-RUN bundle exec rake deploy:build
+RUN bundle exec rake deploy:build docker
```
gdpelican: This is definitely the right approach, but I found I needed 

```
RUN bundle exec rake deploy:build[docker]
```

to make this go. Does passing `docker` in like that work for you?

plugins/docker.yml
```diff
@@ -0,0 +1,5 @@
+# Plugins loaded by https://github.com/loomio/loomio/blob/master/Dockerfile
+plugins:
```
gdpelican: This wants to mirror https://github.com/loomio/loomio/blob/master/config/plugins.yml
(we've dropped the top level 'plugins' line)

Dockerfile
dachary: I did not try, that was a quick draft after browsing the code, to check if I was going in the right direction :-)

plugins/docker.yml
dachary: ok

dachary:
amended to address comments & repushed

gdpelican:
Sweet. I've made a few further changes we need to the plugin system here: #3385 

Closing this one for now; we'll get that one merged in soon. Thanks for this!

PR #3385, opened 2016-05-24
Do not load loomio_org plugin for docker image

gdpelican:


robguthrie:
Thanks for sorting this out @gdpelican!

robguthrie:
💚

PR #3387, opened 2016-05-24
Add closed_motions_count to group

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3334

### DIFF
--- a/spec/models/discussion_spec.rb
+++ b/spec/models/discussion_spec.rb
@@ -1,7 +1,10 @@
 require 'rails_helper'
 
 describe Discussion do
-  let(:discussion) { create :discussion }
+  let(:user)       { create :user }
+  let(:group)      { create :group }
+  let(:discussion) { create :discussion, group: group }
+  let(:motion)     { create :motion, discussion: discussion }
 
   context "versioning" do
     before do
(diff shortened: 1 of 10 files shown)


spec/models/discussion_spec.rb
```diff
@@ -47,6 +47,31 @@
...
+    end
+
+    it "returns a count of closed motions" do
+      expect(@discussion.reload.closed_motions_count).to eq 1
+    end
+
+    it "updates correctly after motion is closed" do
+      expect {
+        @discussion.motions.create(attributes_for(:motion).merge({ author: @user })).close!
+      }.to change { @discussion.reload.motions_count }.by(1)
```
gdpelican: `@discussion.reload.closed_motions_count`?

spec/models/group_spec.rb
```diff
@@ -94,7 +94,36 @@
           @discussion.destroy
         }.to change { @group.reload.motions_count }.by(-1)
       end
+    end
+
+    describe "#closed_motions_count" do
+      before do
+        @user = create(:user)
```
gdpelican: Why not use the `let(:user) { create(:user) }` syntax we use most other places? ❔

spec/models/discussion_spec.rb
```diff
@@ -47,6 +47,31 @@
 
   end
 
+  describe "#closed_motions_count" do
+    before do
+      @user = create(:user)
```
gdpelican: (same here, why not `let(:user) { create(:user) }`?

gdpelican:
Cool, a couple nits, but looks good! 🍏

robguthrie:
nice work hannah

spec/models/group_spec.rb
HSalmon: I used this syntax because it's consistent with the other counter cache specs. Should I change it?

gdpelican:
Great work @HSalmon !

Issue #3386, opened 2016-05-24
Visiting unsubscribe link from email gives borked experience


mhjb:
![image](https://cloud.githubusercontent.com/assets/1380991/15516621/bc81d694-2247-11e6-91b7-5cb6d89a53eb.png)

I seem to be able to change my email settings without logging in. 

Once I do log in from that page, I still get the logged-out header:

![image](https://cloud.githubusercontent.com/assets/1380991/15516735/4dfc9de8-2248-11e6-97e9-3fc3fbd286d5.png)

gdpelican:
This is intended behaviour. You have the ability to change your email
settings if you have an unsubscribe token from your email, but you can't do
other things like comment or vote without logging in.

mhjb:
True!

The second bit is still a bug, I think.

mhjb:
Was that second bit fixed?

PR #3388, opened 2016-05-24
Don't prefer restricted user

gdpelican:
#3386


gdpelican:
The bad bahaviour at the moment is that if you're logged in AND have an unsubscribe token, we assume you're a restricted user, when you're actually not.

So, even if you have an unsubscribe token, you should still be treated as a full-on logged in user, if you're logged in, which primarily means seeing the logged in navbar.

HSalmon:
Just followed the same steps as the test, but works as expected. However if you visit the email settings page when logged out, it looks like this:

screen shot 2016-05-25 at 4 14 10 pm

I'm wondering if we should redirect logged out users who hit that route ❔

gdpelican:
Sweet, I've added that.

Issue #3373, opened 2016-05-18
Ask to join button disappears after you click ask to join


mhjb:
Weird thing. I'm a member of two groups, from /explore I click on an attractive looking group. The Ask to Join Group button flashes up briefly, but then disappears.

Maybe I already asked to join but forgot I did? In which case it would be nicer if the button was visible but disabled, and the text changed 'Membership request pending'.

gdpelican:
I'll grab this.

PR #3389, opened 2016-05-24
Disable ask to join if membership requested already

gdpelican:
#3373


gdpelican:
screen shot 2016-05-24 at 3 05 58 pm

HSalmon:
This works as expected for me! The one thing I was wondering about is whether there's a way to disable the button immediately after a visitor requests membership (not a non-member)? Maybe not necessary, and maybe not possible! Otherwise 🍏

gdpelican:
Yep, it's a good thought, but a bit more complicated because we don't track membership requests that way as far as I can tell.

Issue #2900, opened 2016-01-28
Group title link 'undefined' when viewing subgroup I'm not a member of


mhjb:
I'm logged in, and not a member of the (closed) subgroup or (closed) group. 

The underlined link is to /undefined:

![image](https://cloud.githubusercontent.com/assets/1380991/12661209/1a2473c8-c67d-11e5-8580-342f6e134867.png)

rdbartlett:
Seems related to #2911?

toupeira:
Seems to be an Angular issue, the link is fine if you navigate to the sub-group from the parent, but if you load/reload the sub-group directly you'll get `undefined`.

toupeira:
Slightly related: editing a discussion (opening the dialog through the arrow-menu) changes the URL to `/true`, which leads to a 404 on reload.

gdpelican:
Seems fixed!

mhjb:
Alas not fixed: https://www.loomio.org/g/52jcgA8f/le-r-veil-du-gabon-gouvernance-et-institutions

mhjb:
Sweet!

PR #3390, opened 2016-05-25
Serialize out parent groups for groups controller

gdpelican:
#2900

### DIFF
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -56,4 +56,8 @@ class API::GroupsController < API::RestfulController
     Queries::ExploreGroups.new
   end
 
+  def resource_options(scope:, serializer:, root:)
+    { resources: [resource, resource.parent].compact, scope: scope, serializer: serializer, root: root }
+  end
+
 end
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -56,6 +56,9 @@ class API::GroupsController < API::RestfulController
     Queries::ExploreGroups.new
   end
 
+  # NB: this override means that groups won't ever serialize out an event in response to an API call.
+  # at the time of this writing, we don't ever serialize events from groups_controller.
+  # if you're expecting an event from the groups controller and aren't getting one, this is probably why.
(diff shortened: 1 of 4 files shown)


gdpelican:
@robguthrie Would you mind taking a look at this?

Currently there's no way to tack on a different model to the serialized object's hash (so, adding a separate group to a group serializer doesn't work), I'm pretty sure due to a bug or design flaw in AMS (we're running an ancient version, after all)

This hacks the groups controller to serialize out the group as well as its parent all the time.

Possible shortcoming / weirdness is that we won't get the parent serialized out when requesting from a different controller (so, I request a group from `api/v1/groups/key` and get the parent, but I request a discussion from `api/v1/discussions/key` and I get the discussion, and its group, but not the group's parent) but I'm pretty sure that's alright, if a bit of a gotcha.

app/controllers/api/groups_controller.rb
```diff
@@ -56,4 +56,8 @@ def accessible_records
     Queries::ExploreGroups.new
   end
 
+  def resource_options(scope:, serializer:, root:)
```
robguthrie: Seems likely there are times where there will be an event to return from the groups controlller, right? 

Does this code mean we won't serialize out a event for groups controller?

robguthrie:
I think your code looks good. If the hack works.. then go for it... but I don't see what will happen in the case of event being serialized.

I just checked in briefly on AMS and it gives me the impression that it might be worth trying the 0.10 version. The docs make it look like it might have some really useful features and be prod ready.

app/controllers/api/groups_controller.rb
gdpelican: We don't currently serialize out any events for groups (it'd be the return values from methods in GroupService; so like `discussion_service.rb#create` returns an event because it calls `Events::NewDiscussion.publish!` as a return value, but `group_service.rb` doesn't ever do that.)

Doesn't mean we won't down the road though, which would be another gotcha.

robguthrie: cool. maybe leave a comment to highlight if/when someone is expecting same behaviour?

gdpelican: Done!

PR #3392, opened 2016-05-25
Add ms-flexbox vendor prefix for navbar

gdpelican:


robguthrie:
Oh.. this is why we have a mixin for flexbox.

Maybe use that?

gdpelican:
Oh, perfect, didn't know about that. Done.

PR #3393, opened 2016-05-26
Branded invitation email

HSalmon:
screen shot 2016-05-26 at 2 16 29 pm
screen shot 2016-05-26 at 2 17 14 pm
screen shot 2016-05-26 at 2 18 58 pm
screen shot 2016-05-26 at 2 19 21 pm


HSalmon:
Would love a review if someone's got capacity, @robguthrie, @gdpelican, @rdbartlett, @mhjb

mhjb:
I'd like to help, but I don't think I know how to set up Loomio locally to send real emails. Could it go on loomiotest perhaps?

app/mailers/invite_people_mailer.rb
```diff
@@ -13,7 +13,7 @@ def to_start_group(invitation:,
 
   def to_join_group(invitation:, locale:, subject_key: "email.to_join_group.subject")
     @invitation = invitation
-    @message_body = invitation.message
+    @message_body = invitation.message unless invitation.message.blank?
```
gdpelican: I don't get this; why have a check for blank here? Isn't it okay if it's blank since we do

```
- if @message_body
```

later?

gdpelican:
Did we decide against user avatars next to the inviter?

app/mailers/invite_people_mailer.rb
HSalmon: Ha, I got carried away. I'll fix it up

HSalmon:
Personally I feel like an avatar isn't necessary for this iteration

gdpelican:
I'm fairly certain we'll want to use mailer layouts here
(http://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-layouts)

rather than having a `mailer_header` partial.

I'm not 100% on what the best way to go on the layout stuff is, but a couple options would be:

- Have one layout for all mailers:
```
# layouts/base.html.haml
.header-stuff
= yield
.footer-stuff
```
```
# mailers/thread_mailer, mailers/user_mailer
layout :base
```

- Have a layout each for `thread_mailer`, `invite_people_mailer`, and `user_mailer`, with partials to cover what's common between them. (Maybe only the thread mailer needs a footer to ensure we're clipping signatures, but the others don't, or maybe the appearance needs to be slightly different for an invite email vs a reset password email.)
```
# layouts/thread_mailer.html.haml
render 'mailers/header'
= yield
render 'mailers/footer'
```
```
# layouts/user_mailer.html.haml
render 'mailers/header'
= yield
```

NB: I'm happy with it as it is, but I think layouts will be the way to go as we add more.

gdpelican:
This is a really great step up. 🍏

gdpelican:
PS A smoke test of this on loomiotest would be lovely. :D

PR #3399, opened 2016-05-28
Don't overflow images in proposal outcomes

gdpelican:


gdpelican:
Merging this one as it's a gimme :)

PR #3400, opened 2016-05-29
Polymentions

gdpelican:
This one blew out a bit, primarily because the event / notification system is not set up for polymorphism, because it primarily cares about 'what is the kind of this event?' when serializing.

(IE, now that 'user_mentioned' is polymorphic, I can't figure out what to serialize based on `event.kind`, but I can figure it out based on `event.eventable.class`)

With this I've changed it across the board to care about 'what class is the eventable for this event?' rather than 'what is the kind of this event', which I think is the right way to go here for a more flexible solution.

The other option here would be to add 'comment_user_mention', 'motion_user_mention', and 'discussion_user_mention' event kinds... my gut reaction is that that kinda sucks.


angular/core/components/notifications/notification_controller.coffee
```diff
@@ -1,19 +0,0 @@
-angular.module('loomioApp').controller 'NotificationController', ($scope, LmoUrlService, Records) ->
```
gdpelican: This was unused, heh

angular/core/components/notifications/notifications.haml
```diff
@@ -9,6 +9,6 @@
       %ul.selector-list
         %li.selector-list-header{translate: "notifications.header", role: "heading"}
         %li.selector-list-item.media{ng-class: "{'lmo-active': !notification.viewed}", ng-repeat: "notification in notifications() | orderBy: '-createdAt' track by notification.id"}
-          %a.selector-list-item-link{class: "navbar-notifications__{{notification.kind()}}", lmo-href-for: "notification.relevantRecord()", lmo-href-action: "{{notification.actionPath()}}", ng-click: "broadcastThreadEvent(notification)"}
+          %a.selector-list-item-link{class: "navbar-notifications__{{notification.kind()}}", lmo-href-for: "notification.actionModel()", lmo-href-action: "{{notification.actionPath()}}", ng-click: "broadcastThreadEvent(notification)"}
```
gdpelican: We're in the somewhat odd position of determining the href of the link from outside the item's controller, so the notification needs to know about where it's linking to in `actionModel()`

This could be improved, probably.

angular/core/components/thread_page/activity_card/discussion_edited_item_controller.coffee
```diff
@@ -1,6 +1,6 @@
-angular.module('loomioApp').controller 'DiscussionEditedItemController', ($scope) ->
-  discussion = $scope.event.discussion()
-  version = $scope.event.version()
+angular.module('loomioApp').controller 'DiscussionEditedItemController', ($scope, Records) ->
+  version    = Records.versions.find($scope.event.eventable.id)
+  discussion = Records.discussions.find(version.discussionId)
```
gdpelican: Each event, in addition to serializing out the relevant records, now comes with an 'eventable' hash which looks like

```
eventable: {
  id: 1,
  type: 'discussion'
}
```

That makes it pretty easy to find an event's eventable through the `Records` interface.

gdpelican: Actually, looking at this with fresh eyes, it should probably be

```
# event_model.coffee
eventable: ->
  @recordStore[@eventable.type].find(@eventable.id]
```
which I think we do with notifications.

Anyway this is a solid step up for now.

angular/core/models/event_model.coffee
```diff
@@ -4,53 +4,24 @@ angular.module('loomioApp').factory 'EventModel', (BaseModel) ->
...
-      'motion_outcome_created':         'proposal',
-      'new_comment':                    'comment',
-      'comment_liked':                  'comment',
-      'comment_replied_to':             'comment',
-      'user_mentioned':                 'comment',
-      'invitation_accepted':            'membership',
-      'new_coordinator':                'membership'
-    }
+    @eventTypeMap:
+      group:              'groups'
```
gdpelican: Simplifies the event lookup a lot by going off of what kind of eventable is is, rather than what event kind it is.

(this says, 'if the event type is discussion, use Records.discussions to find it by eventable id)

angular/core/models/notification_model.coffee
```diff
@@ -7,19 +7,27 @@ angular.module('loomioApp').factory 'NotificationModel', (BaseModel) ->
...
+    model:      -> @event().model()
+    kind:       -> @event().kind
 
     group: ->
-      @event().group()
+      switch @model().constructor.singular
+        when 'group' then @model()
+        else              @model().group()
+
+    discussion: ->
```
gdpelican: I found it handy to have a helper function for an event's discussion and group.

It enforces the hierarchy! But is quite handy.

angular/core/services/mention_service.coffee
```diff
@@ -0,0 +1,14 @@
+angular.module('loomioApp').factory 'MentionService', (Records, Session) ->
+  new class MentionService
+    applyMentions: (scope, model) ->
+      scope.updateMentionables = (fragment) ->
```
gdpelican: Turns out the code for mentions works everywhere; this is pasted straight from the existing code.

app/controllers/api/restful_controller.rb
```diff
@@ -48,18 +48,29 @@ def public_records
...
+      respond_with_collection scope: scope, serializer: serializer, root: root
     else
       respond_with_errors
     end
   end
 
-  def resource_options(scope:, serializer:, root:)
-    if @event.is_a?(Event)
-      { resources: [@event], scope: scope, serializer: EventSerializer, root: :events }
+  def respond_with_collection(scope: default_scope, serializer: resource_serializer, root: serializer_root)
```
gdpelican: Okay, I've highlighted the 'serializing events vs serializing models' here a bit more, although it's the same logic.

- If @event is an Event, then we're serializing out events.
- The way we do this, is wrap up the events into an EventCollection, then serialize that
- An EventCollection serializer is real simple; it `has_many :events`
- The reason we do this is that it allows the individual events to choose their own serializer (A new_motion event can serialize a motion, a discussion_edited can serialize a version, etc.) This simplifies the logic tremendously (check out `app/serializers/events/*_serializer.rb`), and there's no way to do this with ActiveModel::ArraySerializer, which requires a single `each_serializer` option to specify one serializer for the whole array.

app/extras/queries/users_to_mention_query.rb
```diff
@@ -0,0 +1,5 @@
+class Queries::UsersToMentionQuery
+  def self.for(model)
+    model.mentioned_group_members - model.group.users.mentioned_in(model)
```
gdpelican: This is a bit different, but much easier to track. It goes 'The users to mention are the group members mentioned in this body, sans the group members who have been mentioned in this model before (<-- users who have a 'user_mentioned' event associated with this model)'

I _think_ this is way better than how we were doing it before, which was to look at the previous iteration and determine a diff based on the two sets of mentions.

The subtle change in behaviour here is we lose is the case 'I mention @bob. He is notified. I remove @bob and mention @alice. She is notified. I remove @alice and mention @bob again. He is notified again.'

^^ I think not having that is perfectly acceptable behaviour, and quite possibly better.

app/models/event.rb
```diff
@@ -26,6 +26,12 @@ class Event < ActiveRecord::Base
 
   acts_as_sequenced scope: :discussion_id, column: :sequence_id, skip: lambda {|e| e.discussion.nil? || e.discussion_id.nil? }
 
+  def active_model_serializer
+    "Events::#{eventable.class.to_s.split('::').last}Serializer".constantize
```
gdpelican: This is a hook into active model serializers, which lets the model say how it should be serialized.

It looks at the eventable class to determine if there should be any special serializer behaviour (like, if it's a motion eventable, we need say 'root: :proposals' for example), otherwise go with the Events::BaseSerializer

app/models/events/comment_replied_to.rb
```diff
@@ -5,16 +5,4 @@ def self.publish!(comment)
            eventable: comment,
            created_at: comment.created_at).tap { |e| EventBus.broadcast('comment_replied_to_event', e, comment.parent.author) }
   end
-
-  def group_key
```
gdpelican: All of these event-specific methods can go; they were only used in beta.

app/serializers/comment_serializer.rb
```diff
@@ -2,10 +2,10 @@ class CommentSerializer < ActiveModel::Serializer
   embed :ids, include: true
   attributes :id, :body, :mentioned_usernames, :created_at, :updated_at, :parent_id, :parent_author_name, :versions_count
 
-  has_one :author, serializer: UserSerializer, root: 'users'
+  has_one :author, serializer: UserSerializer, root: :users
```
gdpelican: I was running into a thing where AMS was serializing out a 'users' key and a :users  key, which b0rked things pretty badly. Normalized on symbols because that's what AMS defaults to.

app/serializers/event_serializer.rb
```diff
@@ -1,132 +0,0 @@
-class EventSerializer < ActiveModel::Serializer
```
gdpelican: So good to be rid of this!

app/services/comment_service.rb
```diff
@@ -37,14 +37,13 @@ def self.destroy(comment:, actor:)
...
   def self.update(comment:, params:, actor:)
-    new_mentions = comment.new_mentions_in(params[:body])
     comment.edited_at = Time.zone.now
     comment.body = params[:body]
 
     return false unless comment.valid?
     actor.ability.authorize! :update, comment
     comment.save!
 
-    EventBus.broadcast('comment_update', comment, new_mentions)
```
gdpelican: Getting rid of the `new_mentions_in` method also means we can normalize the EventBus broadcasts to 'model, params, actor'

Eventually of course we want to move to named parameters, but for now this is a good thing.

PR #3402, opened 2016-05-30
Change new user email defaults

mhjb:
I'm confused. I made this change, but it isn't reflected when I look at Patrick's email settings:

image

### DIFF
--- a/db/migrate/20160530084434_set_new_email_defaults.rb
+++ b/db/migrate/20160530084434_set_new_email_defaults.rb
@@ -1,0 +1,6 @@
+class SetNewEmailDefaults < ActiveRecord::Migration
+  def change
+    change_column :users, :email_on_participation, :boolean, default: false
+    change_column :users, :default_membership_volume, :integer, default: 2
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


gdpelican:
Hm, default will only take place for new users (it doesn't touch existing records) so you'll have to make sure patrick has been wiped and recreated before you see the changes. You could try visiting `/development/email_settings_as_logged_in_user` to ensure the default is picked up correctly.

Also, I don't think we can get away with editing `schema.rb` directly, so we'll want to write a migration:

```
rails g migration set_new_email_defaults # => makes a new migration
```
```
# db/migrate/{{big_fat_timestamp}}_set_new_email_defaults.rb
class SetNewEmailDefaults < ActiveRecord::Migration
  def self.change
    change_column :users, :email_on_participation, :boolean, default: false
  end
end
```

relevant docs: http://edgeguides.rubyonrails.org/active_record_migrations.html#creating-a-standalone-migration

mhjb:
thanks very much James.

mhjb:
Thanks again James. It's looking good now I think.

image

Would it make sense to put it on clone or test to check that emails do what they oughta?

gdpelican:
Sure, I've pushed this to clone for testing.

mhjb:
Thanks James. FYI, @HSalmon is picking this up now

gdpelican:
@HSalmon since this is a well-tested part of the app, we'll need to update
the tests to reflect the new intended functionality.

gdpelican:
Closing in favour of #3403

PR #3403, opened 2016-05-31
Update default email settings

HSalmon:


HSalmon:
Sorry, didn't realise @mhjb was already working on this. Have updated tests and build seems to be passing. Would someone mind doing a CR to make sure I haven't missed anything? @gdpelican, @robguthrie?

app/services/discussion_service.rb
```diff
@@ -28,6 +28,7 @@ def self.create(discussion:, actor:)
     return false unless discussion.valid?
 
     discussion.save!
+    DiscussionReader.for(discussion: discussion, user: actor).set_volume! :loud
```
gdpelican: Is this a change in behaviour? Just trying to understand why we didn't need this before but we do now.

HSalmon: It is a change in behaviour because email on participation used to be true by default, but this PR makes it false by default

gdpelican: Hm, we already do `set_volume_as_required!` on discussion create, I think we can just change that to being loud and take this out of the service

```
  config.listen('discussion_create') do |discussion, actor|
    DiscussionReader.for(discussion, actor).set_volume!(:loud)
  end

  config.listen('discussion_update', 'comment_like') do |model, actor|
    DiscussionReader.for_model(model, actor).set_volume_as_required!
  end
```

HSalmon:
I'm not sure what's up with all of these failures @gdpelican, _sigh_

mhjb:
It's going to be neat to have this changed:

image

gdpelican:
🍏  Thanks @HSalmon and @mhjb !

PR #3412, opened 2016-06-09
Allow pasting of attachments

gdpelican:
Now you can paste attachments with ctrl+v, from anywhere in the form!

(Works for comments and discussions)

http://recordit.co/ErjHJsxUEf.gif


robguthrie:
merge it!!!

PR #3417, opened 2016-06-15
Allow experimental, group-specific plugins

gdpelican:


angular/tasks/app.coffee
```diff
@@ -19,7 +19,11 @@ module.exports = ->
       htmlmin(),                                  # minify html
       template(                                   # store html templates in angular cache
         module: 'loomioApp',
-        transformUrl: (url) -> "generated/components/#{url}"
+        transformUrl: (url) ->
```
robguthrie: could you put a source code comment and explain this bit?

config/plugins.yml
```diff
@@ -1,7 +1,12 @@
 loomio_webhooks:
-  repo:    loomio/loomio_webhooks
-  version: master
+  repo:         loomio/loomio_webhooks
+  version:      master
```
robguthrie: isn't version, branch now?

angular/tasks/app.coffee
gdpelican: Yep, sure. It's a hacky workaround through the angular router, which I'm hoping to find a better solution for.

Basically the templates which come through the core app are getting the format `generated/components/my_component/my_component.html`, while the plugin ones are getting `generated/components/my_component.html`

That's all fine until you go to add a new component to the router, which insists that this:
```
{path: '/tags', components: 'tagsPage'}
```
map to 
```
generated/components/tags_page/tags_page.html
```

which I haven't found an override for yet. Thus this hacks the plugin templates to use the intended component template url.

robguthrie: Thanks. In umyesno I'm using a new version of angular router and it's great.

PR #3422, opened 2016-06-16
NameError - undefined local variable or method instantiate_resouce

sicambria:
Fix typo


gdpelican:
Hm, I think you'll also have to upgrade to the latest version of snorlax (we're currently gemfile locked on 0.1.3), to make this work; this typo is there to match a typo in that repository, which has since been fixed :D

gdpelican:
I'll grab this in this PR: #3464

PR #3423, opened 2016-06-16
failed to run gulp vendor

loganfreeman:
due to wrong path of marked.js


gdpelican:
Loomio supports node version 4.2.6, we haven't made the jump to 5.0.0 yet (Node 5 did away with nested node module folders), so you'll have to use something like NVM to ensure you're on 4.2.6 :)

PR #3435, opened 2016-06-22
Fix griddler unauthorized errors

gdpelican:


gdpelican:
This makes Griddler not blow up when the proper reply keys aren't supplied. It means we'll get far fewer noise on errbit.loomio.io

robguthrie:
thank you

PR #3444, opened 2016-06-25
Fix newrelic errors

gdpelican:
This fixes the three most common and worthless errors coming through newrelic at the moment:

- People visiting '/notifications/dropdown_items' (presumably as old beta tabs that have been left open somewhere) or '/u/:key/username' (which isn't a route anymore)
- Receiving 'bytesize is nil' error through ahoy_email (which is fixed in the latest version)
- Visiting a subdomain which is not associated with a group (this now 404's properly instead of erroring on `ActiveRecord::RecordNotFound`)


gdpelican:
Sorry for the quick pushes and merges recently, but I'm really looking forward to not getting newrelic emails about how Loomio is erroring 5% of the time every day.

PR #3447, opened 2016-06-29
Weekly analytics emails for group coordinators

gdpelican:
screen shot 2016-06-29 at 3 00 43 am

These emails will be sent out weekly to coordinators of 'engaged' Groups (>2 discussions, >1 member, discussion activity within the last month) where 'analytics_enabled' is set to true (this flag is editable in the admin panel)

TODO:

- [x] Where does the contact us link go?
- [x] Where does the 'you can pick a plan here' link go?
- [ ] I'm not certain which font the mockups are using, but it looks different than what's in the email now. Also probably there's some styling cleanup to be done in general.

- [ ] post-deploy: Ensure job is running weekly and at a time in the week we like

@mhjb @robguthrie


mhjb:
**Get in touch** could go to https://loomio.org/contact

But maybe nicer would be just inviting people to reply to the email. What email address will these come from? Slack's come from feedback@

mhjb:
**Pick a plan**: ideally it would behave like this button:

image

mhjb:
The font in the mockups is Roboto; it's what Inbox uses for the emails Loomio sends at present. 

Could we have it say e.g. "1 proposal" and "2 proposals"?

gdpelican:
Replying to this email's out of scope for this iteration. I think the rest of this I can take care of tomorrow.

PR #3463, opened 2016-07-04
Add link to sign up to sign in page

jhass:

### DIFF
--- a/app/views/devise/sessions/new.html.haml
+++ b/app/views/devise/sessions/new.html.haml
@@ -18,6 +18,8 @@
 
           .signin_container
             = render "form"
+            %p.sign_up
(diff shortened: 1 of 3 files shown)


app/views/devise/sessions/new.html.haml
```diff
@@ -18,6 +18,8 @@
 
           .signin_container
             = render "form"
+            %p.sign_up
+              != t :create_account, sign_up_link: link_to(t(:sign_up_link), new_user_registration_path)
```
gdpelican: Cool, I didn't know about `!=` as a raw helper for haml.

gdpelican:
Perfect, thanks @jhass

PR #3469, opened 2016-07-07
Pass scope to warden authenticate

gdpelican:


gdpelican:
This is a fix for the strange logout behaviour. I'm going to push to clone to test.

PR #3485, opened 2016-07-11
Only run weekly jobs once a week

gdpelican:


mhjb:
Does this mean I'll get some emails on Monday? (I've turned group analytics on for a couplea groups)

Issue #3068, opened 2016-03-07
Logging out should be synced across tabs


rdbartlett:
Just saw Michael log out of Loomio in one tab, then absentmindedly try to start a thread in another tab that was already open. There was no error message, just the modal stayed open and wouldn't submit.

Not sure of the best solution to this one.

We could do something that updates all your tabs when you log out? Or if you are logged out and try to submit a form we could kick you out at that point?

gdpelican:
We'll want to kick em after they submit a form I think; otherwise we'll have to log them out of everything on all devices, which I don't think is that desirable. (I don't care to check and see if a client is on the same machine as the one that logged out)

This falls under the umbrella of 'better support for unhappy paths'

PR #3495, opened 2016-07-13
Log out all sessions on signout

gdpelican:
Once you log out, all other sessions will see this modal, and after clicking 'ok' will be logged out.

screen shot 2016-07-13 at 3 24 04 am

for #3068


robguthrie:
Looks good, James. 🍏

PR #3497, opened 2016-07-13
fix broken styling on safari and apply truncated for super long string

reaksmey1:
1. Fix old styling which is not support on safari
2. Apply truncate in case on long string

### DIFF
--- a/angular/core/components/group_page/group_previous_proposals_card/group_previous_proposals_card.coffee
+++ b/angular/core/components/group_page/group_previous_proposals_card/group_previous_proposals_card.coffee
@@ -16,21 +16,3 @@ angular.module('loomioApp').directive 'groupPreviousProposalsCard', ->
 
     $scope.canShowMore = ->
       $scope.group.closedMotionsCount > 3
-
-angular.module('loomioApp').filter 'cut', ->
-  (value, wordwise, max, tail) ->
-    if !value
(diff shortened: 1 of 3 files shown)


angular/core/components/group_page/group_previous_proposals_card/group_previous_proposals_card.coffee
```diff
@@ -16,3 +16,21 @@ angular.module('loomioApp').directive 'groupPreviousProposalsCard', ->
 
     $scope.canShowMore = ->
       $scope.group.closedMotionsCount > 3
+
+angular.module('loomioApp').filter 'cut', ->
```
gdpelican: We have a filter for this called 'truncate', check out `core/filters/truncate_filter.coffee` :)

We use it in subgroups_card.haml

gdpelican: @reaksmey1 Any update on this? Would be great to merge this in, with that minor change!

reaksmey1: @gdpelican yes, the pull request is updated

gdpelican:
Looks good to me, thanks, @reaksmey1 !

PR #3498, opened 2016-07-13
Load all memberships on memberships page

gdpelican:


robguthrie:
mmmm yea. Seems like, longterm, RecordStore needs a better way to define a remotely available collection. 🍏

PR #3501, opened 2016-07-14
Ahoy fixes

gdpelican:


angular/core/services/ahoy_service.coffee
```diff
@@ -1,27 +1,23 @@
...
-    , (path) ->
-      properties =
-        url: $location.absUrl()
-        title: document.title
-        page: $location.path()
+      ahoy.trackClicks()
+      ahoy.trackSubmits()
+      ahoy.trackChanges()
 
-      ahoy.track properties
```
gdpelican: This was the primary bug here, we wanted to be doing `ahoy.track '$view', properties`

Issue #3508, opened 2016-07-15
z-index of navbar is off


gdpelican:
I'd expect to see a pretty shadow:

![screen shot 2016-07-14 at 11 49 40 pm](https://cloud.githubusercontent.com/assets/750477/16863122/b18d40a4-4a1d-11e6-94d1-3672755a1a14.png)

PR #3511, opened 2016-07-15
Sidebar css tweaks

gdpelican:
- Add box shadow to sidebar desktop
- Make toolbar appear above other page elements
- Increase navbar height variable slightly to account for slightly taller navbar

**Before:**
screen shot 2016-07-15 at 12 46 25 am

**After:**
screen shot 2016-07-15 at 12 44 04 am

(ignore the broken profile pic; that's because of my dev env :) )


gdpelican:
Fixes #3508 #3506

PR #3519, opened 2016-07-17
count code contributors per organization

dachary:
The FLOSS Contributor organization groups all contributors that are not
affiliated to any known organization.

Signed-off-by: Loic Dachary <loic@dachary.org>

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,5 @@
+James Kiesel <james@loomio.org> <james.kiesel@gmail.com>
+Robert Guthrie <rob@guthr.ie> <rguthrie@gmail.com>
+Hannah Salmon <hr.salmon@gmail.com>
+Nati <nati@loomio.org>
+Matthew Bartlett <mhjbnz@gmail.com>
--- a/.organizationmap
+++ b/.organizationmap
@@ -1,0 +1,23 @@
+#
+# Display the number of contributors active in the past six months, per organization
+#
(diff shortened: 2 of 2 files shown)


robguthrie:
Hi @dachary, can you tell me why should we put this into the codebase? 
I have concerns that it's not really helpful, and will be out of date as soon as it's comitted.

The repo is intended to hold Loomio code, I think github already does a good job of displaying metadata about the code.

dachary:
The .mailmap file is generally useful because it helps avoid duplicates when displaying or looking for commit authors, reason why it is found in a number of repositories [such as git itself](https://github.com/git/git/blob/master/.mailmap). The .organizationmap is less common but serves a similar purpose, for the organizations and is easier to maintain with the code base (see [the file in ceph for instance](https://github.com/ceph/ceph/blob/master/.organizationmap)).

The authors who contribute to the code base are expected to update this files so that it eventually becomes a well maintained list of active contributors from which statistics can be extracted. And lawyers love it when the authors and their affiliation is well maintained because it makes their life much easier during due diligence.

My personal incentive for helping maintain this list is to provide a verifiable way to measure which organization has influence over the code base of a given software.

Does that answer your question ?

robguthrie:
Thanks for the clear explanation.

Are you willing to maintain this list into the future? We have not talked committed to this process within the team, so it's likely to remain unmaintained.

dachary:
I'm willing to maintain this in the future. I've also established this as a [desirable metric in wikidata](https://www.wikidata.org/wiki/Wikidata:WikiProject_Informatics/FLOSS#developer) and I'm hopeful it will be an additional incentive for co-maintainers.

robguthrie:
Ok. While it's maintained I'm happy to include it. Thank you.

PR #3527, opened 2016-07-19
Support safari 8 for #4384

gdpelican:


robguthrie:
oh wow. nice work champ. 🍏

Issue #3426, opened 2016-06-20
Our connection to Intercom is buggy


mhjb:
Talk to me for an example

mhjb:
Intercom is also missing group info for lots of users

mhjb:
I lack confidence in Intercom's numbers (for e.g. MAUs) at the moment — there's a big divergence between those and Metabase's.

mhjb:
<img width="408" alt="groups started per day according to intercom our database" src="https://cloud.githubusercontent.com/assets/1380991/16600457/41cfa23e-435b-11e6-95d4-f1e19bd9e64b.png">

gdpelican:
I'm going to close this as it's pretty unspecific. I'd love to hear about specific intercom issues, if any are left.

robguthrie:
Good call. I've done lots of work on this recently and I reckon it's valid to close.

PR #3546, opened 2016-07-26
Pass app id correctly to intercom

gdpelican:


app/views/angular/_intercom.html.haml
```diff
@@ -4,7 +4,5 @@
 
   :javascript
     (function(){var w=window;var ic=w.Intercom;if(typeof ic==="function"){ic('reattach_activator');ic('update',intercomSettings);}else{var d=document;var i=function(){i.c(arguments)};i.q=[];i.c=function(args){i.q.push(args)};w.Intercom=i;function l(){var s=d.createElement('script');s.type='text/javascript';s.async=true;
-    s.src='https://widget.intercom.io/widget/{app_id}';
+    s.src='https://widget.intercom.io/widget/#{@appConfig['intercomAppId']}';
```
robguthrie: yikes, was this broken before?

gdpelican: Yeah I don't quite get this one yet.

We do `IntercomService.boot` from within the app, which calls `window.Intercom.boot()` with a bunch of data including the intercom app id and the user id. So I'm not quite sure if or why we need this script. I'm going to test a little further to see if I can just take this out entirely; it'd be nicer to be doing this in-app entirely rather than having garbage third party scripts on the page.

gdpelican: Nah, we need it. I reckon this is probably the cause of #3426 too.

PR #3550, opened 2016-07-29
Attempt to fix attachment links

HSalmon:


gdpelican:
Thanks for the fix attempt, Hannah. After pawing at it for a little bit, I found a slightly simpler approach where we simply make lmo-href sensitive to the 'target: _blank' attribute.

PR #3559, opened 2016-08-02
Revert devise update

HSalmon:


robguthrie:
Hannah I reckon keep that errors line in there for now.

Can you depoly to loomioclone and confirm normal and facebook login work?

robguthrie:
Thanks for doing this @HSalmon but james and I have come up with another approach. Closing

Issue #3569, opened 2016-08-03
Yeserday on loomio email - proposal closed shows vote buttons


robguthrie:
![image](https://cloud.githubusercontent.com/assets/486367/17381380/ac4fb326-5a1f-11e6-966e-0125f3138e80.png)

It should not show vote buttons at this stage.

'scuse the private information :)

robguthrie:
but they look so good!

gdpelican:
Should be fixed now.

PR #3574, opened 2016-08-04
Remove buttons from closed proposal section in missed yesterday email

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3569

screen shot 2016-08-04 at 2 02 06 pm


app/models/motion.rb
```diff
@@ -140,7 +140,7 @@ def members_count
     if voting?
       group.memberships_count
     else
-      self[:members_count]
+      self[:members_count] || 0
```
gdpelican: Was this causing trouble somewhere?

gdpelican:
Looks perfect! 🍏

app/models/motion.rb
robguthrie: I suggested this approach but now I think the db schema should be changed to default: 0 with null: false

Hannah if you don't mind, can you try giving that a go?

PR #3575, opened 2016-08-04
Griddler reply split points

robguthrie:
@gdpelican this is tested and working.

I've modified Griddler so we can insert our own reply split regexes. Then I include 2 new regexes.

The first will match any line which has:
"(Loomio)" and the notifications email address and ends with a colon.

This will match most of the comments which have failed to match in the past.

The second one will match a series 8 of invisible characters, which we put at the top of the thread mailer layout. So should we fail to recognise their "In reply to" line from their mail client, we'll then surely match the next invisible line.. so the damage won't go far.

I've tested this from google inbox -> loomiotest.org and it works well, but I don't have any automated tests.. Can you think of an easy way to add a test for this functionality?


config/initializers/griddler.rb
```diff
@@ -1,6 +1,10 @@
 Griddler.configure do |config|
   config.processor_class = EmailProcessor # CommentViaEmail
   config.processor_method = :process # :create_comment (A method on CommentViaEmail)
-  config.reply_delimiter = '-- REPLY ABOVE THIS LINE --'
+  config.reply_delimiter = ['-- REPLY ABOVE THIS LINE OR DIE --']
```
robguthrie: can delete this line

gdpelican:
My thoughts would be:

- Take a raw email body that loomio.org sent you, copy it to a file (new_comment.email)

- Change it so the headers and body match whatever scenario you'd like to test (new_comment_from_outlook.email)

- Write a test which creates a mail message, stub out the body and read it from the test file
```
let (:mail) { Mail::Message.new(from: 'example@ex.com', body: File.read('new_comment.email')) }
```

- Pass that email to our processor, and make sure the email comes out with the body missing from the other side.
```
expect(Email::Processor.new(mail).process!.body).to_not contain 'Fizzy Whiskers, Batman!'
```

robguthrie:
@gdpelican I worked out how to test this! Can I have an apple?

gdpelican:
Looks great, good work! :green_apple:

Issue #3549, opened 2016-07-28
On thread page, when I select invite people from FAB, this group should be selected


rdbartlett:
Currently it is just selecting the first group in my groups list.

gdpelican:
Just marking this as a 'feature', since it's never behaved that way. There's several things I want to do to make the FAB actions behave more intuitively.

HSalmon:
@gdpelican, does this mean you think this should be a bigger job than just this one improvement?

gdpelican:
I'd be happy to just knock this one off, but there's an architecture question in my head around how we want to best handle this.

We've been running into some situations recently (like 'how do I tell which group to highlight in the sidebar', or 'should I display this plugin which only appears for some groups?'), which point to the need for storing what the 'current group', and/or 'current thread' is in the application state. This is another one of those things where we have this idea of 'what group am I currently interacting with? What thread?', and it's the same whether you're on the memberships page, the previous proposals page, etc.

We've skirted with this idea some, but it's not fully realised yet. The simplest implementation is this:

- Everytime we visit a new page, we broadcast 'current component', along with some info about the page. (this works today)
- When we broadcast that information, we could pass the group / thread through with it

```
$rootScope.broadcast 'currentComponent', {page: 'threadPage', group: @discussion.group(), discussion: @discussion}
```
if a page isn't particularly associated with a group, then we don't pass that info
```
$rootScope.broadcast 'currentComponent', {page: 'dashboardPage'}
```

Then, when we listen for the broadcast, we set the current group on the AppConfig

```
$scope.$on 'currentComponent', (data) ->
  AppConfig.currentGroup = data.group.id
```

Which means that the FAB can access that info when trying to determine which thing to preselect when opening the modal

```
ModalOpen ThreadForm, discussion: -> Records.discussions.build(groupId: AppConfig.currentGroupId)
```

So that's the idea... I'm just 100% on wanting to implement it yet.

rdbartlett:
Sounds great James. I'll put it on the Groomers agenda. I agree with your
sense that it's not top priority.

robguthrie:
Sounds good to me. Did a quick mental walk though an alternative where we broadcast 'currentGroup' but that gets dumb when you navigate to components that don't specify a group, like user settings. I like what you've suggested because it explicitly clears currentGroup if you navigate away from a group context.

HSalmon:
[...]

PR #3576, opened 2016-08-04
Select correct group in thread and invitation forms launched from FAB

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3549

### DIFF
--- a/angular/core/app.coffee
+++ b/angular/core/app.coffee
@@ -77,6 +77,7 @@ angular.module('loomioApp').controller 'ApplicationController', ($scope, $timeou
     MessageChannelService.subscribe()
 
   $scope.$on 'currentComponent', (event, options = {}) ->
+    Session.currentGroup = options.group
     $scope.pageError = null
     $scope.$broadcast('clearBackgroundImageUrl')
     ScrollService.scrollTo(options.scrollTo or 'h1')
--- a/angular/core/components/discussion_form/discussion_form.coffee
+++ b/angular/core/components/discussion_form/discussion_form.coffee
@@ -2,6 +2,7 @@ angular.module('loomioApp').factory 'DiscussionForm', ->
   templateUrl: 'generated/components/discussion_form/discussion_form.html'
   controller: ($scope, $controller, $location, discussion, Session, Records, AbilityService, FormService, MentionService, AttachmentService, KeyEventService, PrivacyString, EmojiService) ->
     $scope.discussion = discussion.clone()
+    $scope.discussion.groupId = Session.currentGroupId()
 
     if $scope.discussion.isNew()
       $scope.showGroupSelect = true
--- a/angular/core/components/invitation_form/invitation_form.coffee
+++ b/angular/core/components/invitation_form/invitation_form.coffee
@@ -6,7 +6,7 @@ angular.module('loomioApp').factory 'InvitationForm', ->
       _.filter Session.user().groups(), (group) ->
         AbilityService.canAddMembers(group)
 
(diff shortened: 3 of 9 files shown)


angular/core/app.coffee
```diff
@@ -77,6 +77,7 @@ angular.module('loomioApp').controller 'ApplicationController', ($scope, $timeou
     MessageChannelService.subscribe()
 
   $scope.$on 'currentComponent', (event, options = {}) ->
+    Session.currentGroup = options.group
```
gdpelican: Putting it on the Session seems like a good call to me 🍏

angular/core/components/discussion_form/discussion_form.coffee
```diff
@@ -2,6 +2,7 @@ angular.module('loomioApp').factory 'DiscussionForm', ->
   templateUrl: 'generated/components/discussion_form/discussion_form.html'
   controller: ($scope, $controller, $location, discussion, Session, Records, AbilityService, FormService, MentionService, AttachmentService, KeyEventService, PrivacyString, EmojiService) ->
     $scope.discussion = discussion.clone()
+    $scope.discussion.groupId = Session.currentGroupId()
```
gdpelican: I think I'd move this out into the places where we call
```
ModalService.open DiscussionForm, discussion: Records.discussions.build(Session.groupId)
```

My thoughts there are that could envision us wanting to start a thread that wasn't necessarily the current group's thread; maybe there's a button in the sidebar for each group which says 'start a thread by clicking on me!', and we get stuck because the discussion form itself is locked to the current group.

angular/core/components/invitation_form/invitation_form.coffee
```diff
@@ -6,7 +6,7 @@ angular.module('loomioApp').factory 'InvitationForm', ->
       _.filter Session.user().groups(), (group) ->
         AbilityService.canAddMembers(group)
 
-    $scope.form = Records.invitationForms.build(groupId: group.id or (_.first($scope.availableGroups()) or {}).id)
+    $scope.form = Records.invitationForms.build(groupId: Session.currentGroupId() or (_.first($scope.availableGroups()) or {}).id)
```
gdpelican: Maybe same thing here; if we ensure we're always passing in the currentGroup to the InvitationForm, this code will work as-is.

HSalmon: We're passing in an id though, rather than a group, so group.id wouldn't work. Should I still update it?

gdpelican: Oh yeah. I was thinking that, instead of any of the forms knowing about Session.currentGroup, they're simply told which group they're rendering, so that we always have the option of rendering a form which isn't the current group.

angular/core/components/start_menu/start_menu_option.coffee
```diff
@@ -12,7 +12,7 @@ angular.module('loomioApp').directive 'startMenuOption', ->
         when 'startGroup'
           ModalService.open GroupForm, group: -> Records.groups.build()
         when 'startThread'
-          ModalService.open DiscussionForm, discussion: -> Records.discussions.build(groupId: $scope.currentGroupId())
+          ModalService.open DiscussionForm, discussion: -> Records.discussions.build(groupId: Session.currentGroupId())
```
gdpelican: Should we lose the `$scope.currentGroupId` method as well?

robguthrie:
Looks good to me. Thanks @HSalmon. Is this ready to merge?

robguthrie:
@HSalmon are you able to add a single e2e test that confirms the behavior?

gdpelican:
Looks good to me! 🍏

PR #3577, opened 2016-08-04
untested pass at adding proposal outcomes count

robguthrie:


db/schema.rb
```diff
@@ -11,27 +11,26 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160629045209) do
+ActiveRecord::Schema.define(version: 20160804083238) do
```
gdpelican: Hm, I wonder what we can do to mitigate these whitespace schema changes from coming through all the time.

Is it that we have different editor settings, or just that our schema is old and rails is trying to update it to sane defaults?

gdpelican:
I reckon counter_cache has worked out pretty great, huh? 🍏

db/schema.rb
robguthrie: I think our schemas between dev environments are actually different. This might be a difference in postgres version or the way our databases were created?

robguthrie:
Yeah. I reckon it's been real good.

PR #3578, opened 2016-08-04
Intercom user has profile photo, belongs to paying group

robguthrie:


app/models/user.rb
```diff
@@ -283,6 +283,10 @@ def send_devise_notification(notification, *args)
     I18n.with_locale(locale) { devise_mailer.send(notification, self, *args).deliver_now }
   end
 
+  def belongs_to_paying_group
+    groups.paying.any?
```
gdpelican: Since we serialize out all the memberships with your current user, I think we can do this on the clientside without an additional serialization query

```
# user_model.coffee
belongsToPayingGroup: ->
  _.any @groups(), (group) -> group.subscriptionKind == 'paid'
```

Issue #3456, opened 2016-07-04
Discussion not inheriting email settings of group


alannallama:
I have a group set to 'no emails', in which I started a thread...
<img width="616" alt="screen shot 2016-07-04 at 1 17 42 pm" src="https://cloud.githubusercontent.com/assets/5498878/16548732/f145a0ba-41e9-11e6-8386-ed34e9531994.png">
However, I got a comment reply emailed to me...
<img width="744" alt="screen shot 2016-07-04 at 1 18 23 pm" src="https://cloud.githubusercontent.com/assets/5498878/16548748/0fb910e0-41ea-11e6-8bda-199adcd5039f.png">
So I went and checked this discussions settings and found that it was set to "all activity"...
<img width="614" alt="screen shot 2016-07-04 at 1 18 04 pm" src="https://cloud.githubusercontent.com/assets/5498878/16548757/4e48893a-41ea-11e6-83f1-677938abb3bc.png">

I did not manually change the settings on this discussion. It seems to have incorrectly changed to "all activity" on its own?

gdpelican:
We set discussions which you create to loud, the idea being that you're
interested in how folks respond to your own discussion.

alannallama:
Oh OK so I guess this isn't a bug, sorry. This isn't my preferred behavior but not a bug :)

gdpelican:
Yeah, we've heard that complaint before, from @rich I think. So it's not
just you; there's probably some work to be done around making that obvious
and/or configurable.

robguthrie:
Hi, I've just chatted to @alannallama about this. I want to chime in a say that I think this IS a bug.

We have a checkbox "Subscribe on participation" which indicates if you should get emails when you particiapte (or start) a thread. In Alanna's case, she has this set to false, so she should not get emails for threads she starts.

rdbartlett:
Agreed this behaviour feels like a bug.

mhjb:
Agreed

gdpelican:
Fixed pending deploy

PR #3581, opened 2016-08-04
Don't set to loud unless site setting

gdpelican:
For #3456


robguthrie:
hooooray! thanks james! 🍏  💚

PR #3582, opened 2016-08-04
Move sidebar to component

gdpelican:


gdpelican:
This will require a little smoke testing, but I think we want this to happen.

robguthrie:
nice one. we had endless scope problems when we tried to do this. it was super weird. did you learn anything relating to that?

gdpelican:
Hm, I didn't. My hurdles with this PR were:

- Converting the haml from ruby-haml flavoured to node-haml flavoured was a pain. This is one of the big reasons I want to make angular.html.haml as tiny as possible, so we don't have to deal with two haml parsers that are totally different (massive confusion)
- I removed the '.lmo-sidebar-wrapper' section so that %md-sidenav and .lmo-main-content could be siblings; otherwise I was having weird problems rendering the sidebar in the right place.

PR #3586, opened 2016-08-06
Update intercom_service.coffee

rdbartlett:
Remove extra slash. Currently being sent thru as "admin_link":"https://www.loomio.org//admin/groups/VKlSony8"

### DIFF
--- a/angular/core/services/intercom_service.coffee
+++ b/angular/core/services/intercom_service.coffee
@@ -5,7 +5,7 @@ angular.module('loomioApp').factory 'IntercomService', ($rootScope, $window, App
     key: group.key
     name: group.name
     description: group.description.substring(0, 250)
-    admin_link: AppConfig.baseUrl+"/admin/groups/#{group.key}"
+    admin_link: AppConfig.baseUrl+"admin/groups/#{group.key}"
     subscription_kind: group.subscriptionKind
     subscription_plan: group.subscriptionPlan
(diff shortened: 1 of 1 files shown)


angular/core/services/intercom_service.coffee
```diff
@@ -5,7 +5,7 @@ angular.module('loomioApp').factory 'IntercomService', ($rootScope, $window, App
     key: group.key
     name: group.name
     description: group.description.substring(0, 250)
-    admin_link: AppConfig.baseUrl+"/admin/groups/#{group.key}"
+    admin_link: AppConfig.baseUrl+"admin/groups/#{group.key}"
```
robguthrie: @rdbartlett thanks for the fix, but I think it's better to fix it in AppConfig.baseURL.. seems to me like we'd want any url or piece of url to make sense by itself.

robguthrie: Looking at angular_helper.. it's not so easy. I wonder if we need a compose url method that removes double-ups. uh.. why is this hard, @gdpelican what do you think?

gdpelican: I think my preferred syntax would be

```
LmoUrlService.route(model: @group, absolute: true, namespace: 'admin')
```

Looks like we handle that case already in lmo_url_service on line 13:
```
  "/".concat(route).replace('//', '/')
```

PR #3587, opened 2016-08-07
Namespace LmoUrlService for intercom urls

gdpelican:
In lieu of #3586


gdpelican:
I tested this like so:

screen shot 2016-08-07 at 11 56 04 am

PR #3588, opened 2016-08-07
Material modals

gdpelican:


robguthrie:
Woha.. This is way less change than I expected.

gdpelican:
Yeah me too. The only fiddly bit is that we have to tell each modal what '$close' is, which we do manually with these lines:

```
$scope = $rootScope.$new(true)   # create a new scope and isolate it so it doesn't know about its parent
$scope.$close = $mdDialog.cancel # set $scope.$close to the modal cancel function
$mdDialog.alert(scope: $scope)   # pass the scope through to the modal
```

which the ModalService handles pretty nicely.

robguthrie:
Maybe time to change the modal architecture so we include the ModalService
in the Modal itself?

PR #3597, opened 2016-08-08
fix NoMethodError: undefined method `formatted_offset'

robguthrie:
on marketing#index

https://errbit.loomio.io/apps/57215524657272000e000001/problems/57a8a56a6572720f72000454


gdpelican:
Whoa I had no idea that code existed. Anyway this should be good, although I've been preferring ruby's `&.` over rails's `try` method... strictly because it's less characters and is a little less jarring to read.

PR #3599, opened 2016-08-09
[Needs CR, QA] Truncate long comment

HSalmon:
expand-comment-3


HSalmon:
I'll push this up to a test server asap, but in the interim, could probably do with a CR - @gdpelican, @robguthrie, do either of you have capacity?

angular/core/components/thread_page/activity_card/new_comment.scss
```diff
@@ -6,3 +6,36 @@
 .new-comment__in-reply-to {
   a { color: $primary-text-color; }
 }
+
+.new-comment__truncated {
+  max-height: 180px;
+  overflow: hidden;
+  margin-bottom: 0 !important;
+  p { margin-bottom: 0 }
```
gdpelican: Why the need for important here?
Also, does this work without the trailing semicolon?

angular/core/components/thread_page/activity_card/new_comment_item_controller.coffee
```diff
@@ -1,5 +1,8 @@
 angular.module('loomioApp').controller 'NewCommentItemController', ($scope, $rootScope, $translate, Records, Session, ModalService, EditCommentForm, DeleteCommentForm, AbilityService, TranslationService, RevisionHistoryModal) ->
   $scope.comment = Records.comments.find($scope.event.eventable.id)
+  $scope.commentCollapsed = true
+  $scope.commentTooLong = $scope.comment.cookedBody().length > 600
```
gdpelican: I'm uncomfortable with saying 'if it's longer than 600 characters, truncate to 175px height'. I think we need to pick an implementation here, either height-based or length-based.

HSalmon: I'm totally happy for that. Could you suggest a way to implement?

HSalmon: @robguthrie, do you have a sense of the best way to implement a specifically height-based or length-based solution?

angular/core/components/thread_page/activity_card/new_comment.scss
HSalmon: Because there are margins on the other classes associated with this particular element. They were overriding the css I was trying to add. Is that all good?

gdpelican:
We made a plugin out of this, yay! #3628

Issue #3594, opened 2016-08-08
Vote links in email don't go to vote modal


mhjb:
![](http://g.recordit.co/21LCzaeuVv.gif)

Chrome / OS X

Wondering how this slipped through our QA process…

rdbartlett:
I would have called this Medium priority as it is a Core workflow, a bit munted. While we get around this bug prioritisation system, would you mind spelling out your thought process Matt?

HSalmon:
@mhjb, it was broken before the thread mailer deploy

mhjb:
I wondered about 'a bit munted' too — but in this case it seemed like a core workflow that just wasn't working. On reflection the ease of the workaround probably does warrant 'a bit munted' rather than horridly confusing or whatever the higher-level one is (haven't got the doc to hand).

HSalmon:
I'm having a gander at this now :eyes:

gdpelican:
Fixed pending deploy

PR #3600, opened 2016-08-09
Fix vote by email

HSalmon:
Fix for: https://github.com/loomio/loomio/issues/3594

I introduced this bug when I tried to fix issue https://github.com/loomio/loomio/issues/3169 with this PR: https://github.com/loomio/loomio/pull/3273

### DIFF
--- a/angular/core/components/thread_page/thread_page_controller.coffee
+++ b/angular/core/components/thread_page/thread_page_controller.coffee
@@ -3,6 +3,7 @@ angular.module('loomioApp').controller 'ThreadPageController', ($scope, $routePa
 
   @requestedProposalKey = $routeParams.proposal or $location.search().proposal
   @requestedCommentId   = parseInt($routeParams.comment or $location.search().comment)
+  $location.url($location.path()) unless $location.search().position
 
   handleCommentHash = do ->
     if match = $location.hash().match /comment-(\d+)/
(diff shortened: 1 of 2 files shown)


angular/core/components/thread_page/thread_page_controller.coffee
```diff
@@ -3,7 +3,7 @@ angular.module('loomioApp').controller 'ThreadPageController', ($scope, $routePa
 
   @requestedProposalKey = $routeParams.proposal or $location.search().proposal
   @requestedCommentId   = parseInt($routeParams.comment or $location.search().comment)
-  $location.url($location.path())
+  $location.url($location.path()) unless $location.search().position
```
robguthrie: Can you explain the behaviour here, @HSalmon?
When will we see a search().position be present, what is the user doing in that case?

HSalmon: Voting via email

HSalmon: Actually, it occurred to me that there are probably other times when we don't want to clear the url e.g. when a person is following a link in a membership request email (i.e. someone has requested access to one of their groups).

gdpelican:
Pretty sure we want to clear the route params, but we want to do it later, (ie, after we've called `openVoteModal`)

How about doing it at the end of the, possibly now ill-named, `performScroll` method?

PR #3602, opened 2016-08-09
Move to simpler single pane sidebar

gdpelican:
For #3598 

I've just removed the additional scroll area from the sidebar, which should make it much simpler to edit going forward.


robguthrie:
Do we need to do some browser compatibility testing on this?

I don't get this PR right now.. I see the sidebar is now one panel.. What's the difference with this?

gdpelican:
Er, this is the implementation of your suggestion here:
https://github.com/loomio/loomio/issues/3598

With one panel, we don't have to deal with crazy scrolling issues.

robguthrie:
Heh. Thanks James. The thing I'm finding confusing about this is that the sidebar already operates as one panel for me in chrome.

As this screenshot kind of suggests:
image

gdpelican:
Yeah. The hope is that this operates the same on chrome, but works on Firefox because there aren't multiple scrolling divs.

robguthrie:
just updated/merged this against master.. I'll deploy soon

gdpelican:
Alright to merge this @robguthrie ?

robguthrie:
yeah! 🍏

PR #3603, opened 2016-08-10
SpikeSperiment - Migrate user service

robguthrie:
We've always put off doing this work. Today I felt inspired to give it a crack. 

@gdpelican If you have time, can you describe, loosely what kind of pitfalls you expect and what level of testing you expect to see to feel ok about this.

And generally if you think this approach is ok.


gdpelican:
Do you have energy to push this experiement over the edge, or shall we close it @robguthrie ?

PR #3611, opened 2016-08-10
use safe values for membersCount and votersCount

### LINKED ISSUE #3604: 64 + 13 != 75
image

Gotta run so don't have more detail at this stage. Just noticed this on prod on `d/RIMCfuaT`

robguthrie:
should fix #3604


robguthrie:
I've tested this locally and it seems good to go.

Issue #3609, opened 2016-08-10
Alignment of links in sidebar


rdbartlett:
In Chrome:

![image](https://cloud.githubusercontent.com/assets/970124/17570284/e093755c-5f9f-11e6-9863-e80cbe7bae2d.png)

My Unread and at least one of my Parent groups is being indented by a few pixels

HSalmon:
Hope the PR referenced above will also fix the parent group indentation problem too, I can't reproduce that

HSalmon:
Fixed pending deploy

PR #3621, opened 2016-08-12
Fix alignment of unread and subgroups

HSalmon:
screen shot 2016-08-12 at 1 02 18 pm

Hope this will also fix the parent group misalignment @rdbartlett describes here: https://github.com/loomio/loomio/issues/3609. Can't reproduce


robguthrie:
looks good. thanks!!! 🍏

Issue #3629, opened 2016-08-15
"Deactivate account" option not showing on profile page


rdbartlett:
I'm trying to understand exactly what's changed here. 

As I recall, we used to show "deactivate account" to everyone. Clicking the button opens a modal that will either let you deactivate, or explain that you need to appoint a new group coordinator before you can leave.

Now we're just hiding the button from people who can't deactivate, which is resulting in support requests because people don't understand why they can't deactivate.

Here's the design from last year, I'm not sure exactly how far we got with implementing this though, so I think we can either follow this design, or revert the last change that hid this option.

![user settings page](https://cloud.githubusercontent.com/assets/970124/17683141/9d0c5480-63a4-11e6-8e3f-1051dc37e48c.png)
![user settings - about deactivation](https://cloud.githubusercontent.com/assets/970124/17683142/9d0d32e2-63a4-11e6-97a1-fcc6db9cd41e.png)
![user settings - deactivate group](https://cloud.githubusercontent.com/assets/970124/17683140/9d0c3b58-63a4-11e6-815d-1f3bc866ec48.png)
![user settings - deactivate account](https://cloud.githubusercontent.com/assets/970124/17683139/9d0be37e-63a4-11e6-8657-6644869445a3.png)

*Process point: this is an example of where I want a label to indicate something between a bug and a feature.*

gdpelican:
Re process point I think the discussion tag is appropriate for things like these.

HSalmon:
This issue has been resolved!

PR #3632, opened 2016-08-16
Reveal deactivate account option to all

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3629
deactivate-profile-1

### DIFF
--- a/angular/core/components/only_coordinator_modal/only_coordinator_modal.coffee
+++ b/angular/core/components/only_coordinator_modal/only_coordinator_modal.coffee
@@ -1,0 +1,12 @@
+angular.module('loomioApp').factory 'OnlyCoordinatorModal', ->
+  templateUrl: 'generated/components/only_coordinator_modal/only_coordinator_modal.html'
+  controller: ($scope, $location, Session, LmoUrlService) ->
+
+    $scope.groups = ->
+      _.filter Session.user().groups(), (group) ->
+        _.contains(group.adminIds(), Session.user().id) and
+        !group.hasMultipleAdmins
+
+    $scope.redirectToGroup = (group) ->
+      $location.path LmoUrlService.group(group)
+      $scope.$close()
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -780,8 +780,5 @@ DEPENDENCIES
   uglifier (~> 3.0.0)
   webmock
 
-RUBY VERSION
-   ruby 2.3.0p0
-
 BUNDLED WITH
    1.12.5
--- a/angular/core/components/deactivate_user_form/deactivate_user_form.haml
+++ b/angular/core/components/deactivate_user_form/deactivate_user_form.haml
@@ -1,4 +1,4 @@
-%form{ng-submit: "submit()"}
+%form.deactivate-user-form{ng-submit: "submit()"}
   .lmo-disabled-form{ng-show: "isDisabled"}
   .modal-header
     %modal_header_cancel_button
--- a/angular/core/components/deactivation_modal/deactivation_modal.coffee
+++ b/angular/core/components/deactivation_modal/deactivation_modal.coffee
@@ -1,0 +1,9 @@
+angular.module('loomioApp').factory 'DeactivationModal', ->
+  templateUrl: 'generated/components/deactivation_modal/deactivation_modal.html'
+  controller: ($scope, AbilityService, ModalService, DeactivateUserForm, OnlyCoordinatorModal) ->
+
+    $scope.submit = ->
+      if AbilityService.canDeactivateUser()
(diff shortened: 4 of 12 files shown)


angular/core/components/only_coordinator_modal/only_coordinator_modal.coffee
```diff
@@ -0,0 +1,12 @@
+angular.module('loomioApp').factory 'OnlyCoordinatorModal', ->
+  templateUrl: 'generated/components/only_coordinator_modal/only_coordinator_modal.html'
+  controller: ($scope, $location, Session, LmoUrlService) ->
+
+    $scope.groups = ->
+      _.filter Session.user().groups(), (group) ->
+        _.contains(group.adminIds(), Session.user().id) and
```
gdpelican: We should move this into AbilityService I think.

```
isOnlyCoordinator (group) ->
  @canAdministerGroup(group) and !group.hasMultipleAdmins
```

gdpelican:
I'd really rather have this information show up on the original deactivate modal, rather than going through the mental gymnastics of clicking 'Yes, I want to deactivate' only to be told I can't.

Can we say something like:

```
When you deactivate your account:
- stuff
- happens

You are currently unable to deactivate because you a coordinator etc.
```

in a single modal, with a disabled 'Yes I want to deactivate' button?

HSalmon:
@gdpelican, I was just building this according to the design. @rdbartlett, what do you think about James' suggestion above?

rdbartlett:
I'm happy with the original design, and I'm happy with James's suggestion too. Having a bit of friction in the deactivation process seems like a good thing.

On the original issue I was intending to open a conversation, rather than jumping straight to building something. Specifically I was trying to figure out what's changed.

rdbartlett:
Ideally we would have had the discussion before the building. Now that it's built to match the design though, I reckon it is good enough to merge.

We can add another ticket to the inbox if we want to do another iteration

HSalmon:
Thanks @rdbartlett, I do wonder whether this could do with a smoke-test though...

HSalmon:
@gdpelican, @rdbartlett, @mhjb, do any of you have capacity to smoke-test this branch?

rdbartlett:
Can you get it up on a test env for me to play with?

mhjb:
ditto for me

mhjb:
It works properly but gives an odd message at the end:

image

This might be out of scope, but why do we remove people's names from people's contributions when they deactivate their account? Scenario: I'm running a board of directors on Loomio. A past member contributed in various ways including voting on important proposals, but then their term was up and they moved on. They weren't using Loomio in any other context so deactivated their account. Why should their name be removed from the record of our past important decisions?

rdbartlett:
Tested on Chrome:

With multiple groups, confirmed I am told to close them before I can deactivate. Confirmed theg groups link to the right place.

With no groups, confirmed I can deactivate the account.

I think the flash text should change from "Your account is not activated yet." to "There is a deactivated account associated with this email address. If you want to reactivate your account, [contact us](#)."

@mhjb I'd say your question is definitely out of scope for this job, and interesting to discuss. The rationale for making that choice was to protect the privacy of people who have commented in public threads. 

(Process wise I think it is really useful to distinguish blocking feedback from non-blocking. None of my feedback is blocking and I'm happy to see this deployed as is, or preferably with an updated flash text.)

HSalmon:
Thanks very much for testing @mhjb and @rdbartlett, I'll update the flash message.

HSalmon:
screen shot 2016-08-19 at 3 45 26 pm

HSalmon:
I think it's ready to go! @rdbartlett, @mhjb, @robguthrie (do you green apple it, Rob)?

mhjb:
Sorry — I should have been more specific in my feedback. The new message is great, but I think it's odd to see it immediately upon deactivating. I wonder if when you successfully deactivate your account, you should be sent to the home page, not a login page.

rdbartlett:
Matt this is built into Devise, I think it is too much effort to change
anything more than the words.

mhjb:
Sorry which bit is built into Devise?

rdbartlett:
The behaviour that says the flash message you see on deactivating is the
same as when you try to login with a deactivated account. We get a fair
number of support requests from people who are trying to log in with
deactivated accounts so this feels like an improvement.

mhjb:
Yeah but I was suggesting upon deactivation we send people to www.loomio.org, rather than to https://www.loomio.org/users/sign_in — is that behaviour built into Devise too?

HSalmon:
@robguthrie, any chance I could get a green apple on this if you're all good with it?

robguthrie:
🍏 !!!!!

PR #3633, opened 2016-08-17
Explain mute on muted threads page

HSalmon:
[Trello card here](https://trello.com/c/D0ll9PJw/1720-1-explain-muted-threads-on-empty-muted-threads-page)

screen shot 2016-08-17 at 12 52 59 pm

screen shot 2016-08-17 at 12 53 28 pm


angular/core/components/dashboard_page/dashboard_page_controller.coffee
```diff
@@ -68,4 +70,7 @@ angular.module('loomioApp').controller 'DashboardPageController', ($rootScope, $
 
   $scope.$on 'currentUserMembershipsLoaded', => @setFilter()
 
+  @showLargeImage = -> $mdMedia("gt-sm")
+  @showSmallImage = -> $mdMedia("sm") || $mdMedia("xs")
```
gdpelican: Feels like this should be

```
ng-show: "showLargeImage()"
ng-show: "!showLargeImage()"
```

so we can lose `showSmallImage`

angular/core/components/dashboard_page/dashboard_page.scss
```diff
@@ -55,3 +57,14 @@
 .dashboard-page__show-more a:hover{
   color: $primary-text-color;
 }
+
+.dashboard-page__mute-image--large {
+  img { max-width: 730px; }
```
gdpelican: Where does this number come from? Could we get away with $small-max (768px) instead of adding an arbitrary value, perhaps adding margins to the img if need be?

HSalmon:
Updated explain muted modal because it still referenced the filter threads dropdown:

screen shot 2016-08-18 at 12 07 00 pm

angular/core/components/dashboard_page/dashboard_page.scss
HSalmon: That was the width of the text (I think due to the single column layout class). I've updated it now though :)

HSalmon:
@gdpelican, do you think you might be able to QA this branch when you're back on board?

robguthrie:
Can you confirm there are tests around this first mute experience? I'm happy to QA and merge this.. 

You can use loomio-test if you like.

gdpelican:
I'm finding a disconnect between having examples with actual people and having examples with Loomiobot. Which do we want?

angular/core/components/dashboard_page/dashboard_page_controller.coffee
```diff
@@ -1,9 +1,11 @@
-angular.module('loomioApp').controller 'DashboardPageController', ($rootScope, $scope, Records, Session, LoadingService, ThreadQueryService, AbilityService, AppConfig, $routeParams, ModalService, GroupForm) ->
+angular.module('loomioApp').controller 'DashboardPageController', ($rootScope, $scope, Records, Session, LoadingService, ThreadQueryService, AbilityService, AppConfig, $routeParams, $mdMedia, ModalService, GroupForm) ->
 
   $rootScope.$broadcast('currentComponent', { page: 'dashboardPage', filter: $routeParams.filter })
   $rootScope.$broadcast('setTitle', 'Recent')
   $rootScope.$broadcast('analyticsClearGroup')
 
+  @userHasMuted    = -> Session.user().hasMuted
```
gdpelican: Hrm. This should be an experience. I wonder if migrating hasMuted to an experience is worth doing now.

angular/core/components/mute_explanation_modal/mute_explanation_modal.scss
```diff
@@ -1,3 +1,7 @@
+.mute-explanation-modal {
+  max-width: 500px;
```
gdpelican: I think we want to try to stick to our existing variables as much as possible. What about $tiny-max-px, which is 480px?

angular/core/components/inbox_page/inbox_page.scss
```diff
@@ -7,7 +7,7 @@
 }
 
 .inbox-page__heading{
-  padding: $cardPaddingSize;
+  margin: 20px 0 20px 13px;
```
gdpelican: Why this change to the inbox page?

HSalmon: It makes the inbox page heading consistent with the dashboard page headings

HSalmon:
@gdpelican, I've updated this PR. Can you QA locally, or do you want me to push it up to a test env?

HSalmon:
@robguthrie, this has test coverage too - in dashboard spec

gdpelican:
Don't we need to remove the has_muted call from the user serializer? I think this will error otherwise

db/migrate/20160831015641_remove_has_muted_from_users.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveHasMutedFromUsers < ActiveRecord::Migration
+  def change
+    User.where(has_muted: true).each do |user|
+      user.experienced!("threadMuted")
```
robguthrie: mutedThread above vs threadMuted here. I think that's a mistake

I think the best name is the threadMuted one.. because it puts the most important/categorising word first..

HSalmon: Happy to change, was just thinking about how you would say it aloud, e.g. _user experienced mutingThread_

gdpelican:
This worked alright for me, thanks Hannah! 🍏

db/migrate/20160831015641_remove_has_muted_from_users.rb
robguthrie: yeah.. I think your preference sounds better too.

PR #3648, opened 2016-08-28
Tweaks to material modals

gdpelican:
Changes in this PR:

- Make all modals min-width 480px except on tiny screens (where it pops down to 100% width)
- Fix edit comment modal so the emoji popup works and the text area expands correctly.
- Fix the height so it auto-expands to 80% of screen height, but scrolls the modal-body if it's larger than that.


robguthrie:
Hey @gdpelican we've been fire fighting problems with the modals over the last week, and it's confirmed that we must QA this stuff properly with all the browsers and devices before we merge it.

So far we've been finding that safari mobile and safari desktop don't behave properly when the screen is too small. 

One suggestion I have is make the min hight 90 or even 95% rather than 80.. it's pretty cramped on small screens right now.

gdpelican:
I'll deploy this to clone and start a thread to QA it before I finish today.

This was pushed to clone for a loong time, and the only feedback I got was
'meh, the animation seems cool!' Although there should have been a thread
for it (the QA talk happened in Slack), this did get several eyes on it
before it was merged.

robguthrie:
Sorry you had a bad time. I suggest we all start to name more explicitly what we want people to test.

Thanks for pushing to clone: In this case I think we now know that we want to be paying special attention to firefox and safari mobile and desktop versions.

robguthrie:
@mhjb can you please test this with your iphone4 and tell us if the login modal bug is resovled?

robguthrie:
@mhjb and if it is not... can you post a screenshot in here

mhjb:
Alas still can't log in on iPhone 4 — login button has migrated above the fold (yay) but unclickable (fallen behind 'third-party…' label?).

robguthrie:
I have a fix for the login button issue now.. will ciricle back soonish

robguthrie:
Let's merge and deploy this. I feel confident it's an improvement in every way.

gdpelican:
Did this get deployed? Comparing it against master makes it look like there are no changes anymore..

robguthrie:
I don't think it did actually...

robguthrie:
Would be good to check and be sure.

gdpelican:
I just checked in the master branch and found code that was in this branch, so I assumed it had been manually merged somehow. That jives with github saying all these commits happened, but there's no file changes as well.

gdpelican:
Also the material modals look much better in safari on prod :)

PR #3649, opened 2016-08-28
Adjustments to weekly analytics emails

gdpelican:
- Left aligned names in table
- Language tweak for title
- Replace table with contact us message in case of no activity
- Fix for admins with multiple analytics emails

### DIFF
--- a/app/views/user_mailer/analytics.text.haml
+++ b/app/views/user_mailer/analytics.text.haml
@@ -3,7 +3,7 @@
 
 = raw t(:"email.analytics.stats", @stats)
 
-- if @stats.has_activity
+- if @stats[:has_activity]
   = "Name | Motions created | Discussions created | Comments created | Votes created"
   = '---------------------------------------------------------------------------'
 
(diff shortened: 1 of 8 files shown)


app/views/user_mailer/analytics.text.haml
```diff
@@ -1,16 +1,19 @@
-= "Your weekly summary for #{@stats[:since].strftime('%m/%d')} - #{@stats[:till].strftime('%m/%d')}"
+= raw t(:"email.analytics.title", since: @stats[:since].strftime('%d %B'))
 = '---------------------------------------------------------------------------'
 
 = raw t(:"email.analytics.stats", @stats)
 
-= "Name | Motions created | Discussions created | Comments created | Votes created"
-= '---------------------------------------------------------------------------'
+- if @stats[:has_activity]
+  = "Name | Motions created | Discussions created | Comments created | Votes created"
```
robguthrie: I guess this should be translated?

Also you could use tab characters to align the columns, maybe.

gdpelican: I translated the header; not concerned enough with the text version of the email to do tab spacing; can wait until this becomes an actual feature.

Issue #3652, opened 2016-08-29
Updating a motion causes the 'closing_soon' event to be published multiple times


gdpelican:
Causing multiple webhooks to be sent out:
![screen shot 2016-08-28 at 6 59 20 pm](https://cloud.githubusercontent.com/assets/750477/18037763/8dcd9888-6d51-11e6-9400-11ee09cfa626.png)

And notifications:
![screen shot 2016-08-28 at 6 59 58 pm](https://cloud.githubusercontent.com/assets/750477/18037770/a4b24ee0-6d51-11e6-9131-8b54b9260a62.png)

Repro steps:

- Start a proposal in a new discussion
- Update the closing time of the motion
- Trigger the 'proposal closing soon' job
- See that two events are created

Example in production:
https://www.loomio.org/d/Jy9HklOB/proposal/1Phieykm

gdpelican:
I believe this to be fixed; will reopen if we see it crop up again.

PR #3654, opened 2016-08-29
Fix multiple proposal closing soon events

gdpelican:
For #3652

### DIFF
--- a/app/models/motion.rb
+++ b/app/models/motion.rb
@@ -55,15 +55,19 @@ class Motion < ActiveRecord::Base
   scope :chronologically,          -> { order('created_at asc') }
   scope :with_outcomes,            -> { where('motions.outcome IS NOT NULL AND motions.outcome != ?', '') }
 
-  scope :closing_soon_not_published, -> {
+  # recency threshold is the minimum length of time for a new motion_closing_soon event to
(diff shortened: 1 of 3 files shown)


app/models/motion.rb
```diff
@@ -55,15 +55,19 @@ class Motion < ActiveRecord::Base
   scope :chronologically,          -> { order('created_at asc') }
   scope :with_outcomes,            -> { where('motions.outcome IS NOT NULL AND motions.outcome != ?', '') }
 
-  scope :closing_soon_not_published, -> {
+  # recency threshold is the minimum length of time for a new motion_closing_soon event to
+  # be published. So, if the proposal is extended by 2 hours, we don't want to publish another
+  # motion closing soon event, but if it's extended by 10 days, we do.
+  scope :closing_soon_not_published, ->(timeframe, recency_threshold = 2.days.ago) do
      voting
-    .joins("LEFT OUTER JOIN events e ON e.eventable_id = motions.id AND e.eventable_type = 'Motion'")
```
gdpelican: This join was the culprit. We don't use it, and it can cause the same motion to be selected twice (because it has multiple events)

robguthrie:
🍏  nice one

Issue #3596, opened 2016-08-08
Cmd+click on sidebar opens two tabs


rdbartlett:
On FF 48 and Chrome 52, when I cmd+click sidebar links (e.g. to group pages, recent, and unread), it opens two tabs, one to the intended path, and the other to `https://www.loomio.org/#` (the marketing site).

gdpelican:
Fixed pending deploy

PR #3659, opened 2016-08-29
Fix double open for cmd click

gdpelican:
For #3596


gdpelican:
Angular has a weird requirement where %md-button turns into an `a` tag if it has an href, and a `button` if it doesn't. I guess that makes sense, but it messes with our styling a little. This should make it good.

Issue #3656, opened 2016-08-29
Group logos should always be displayed as a square


rdbartlett:
Here's an example of this being slightly broken on the email_preferences page:

![image](https://cloud.githubusercontent.com/assets/970124/18039216/b525e608-6df4-11e6-9861-fd4a82e7ab30.png)

gdpelican:
We've got a 'group_avatar' directive which (I believe) works properly. We just want to move all the times we do some garbage like

```
%img.box--medium{ng-src: "{{group.logoUrl()}}"
```

to use that directive.

```
%group_avatar{group: group}
```

HSalmon:
@rdbartlett, I can't see this on the email settings page and I'm a member of that group...weird

gdpelican:
Fixed pending deploy

PR #3661, opened 2016-08-30
Ensure group logos are always square on email settings page

HSalmon:
Fix for: https://github.com/loomio/loomio/issues/3656

screen shot 2016-08-30 at 12 34 42 pm


gdpelican:
🍏

Issue #3663, opened 2016-08-30
Don't display inbox loaded threads before dashboard threads


gdpelican:
Here's a gif of Loomio loading for me:

![lpkyvondsy](https://cloud.githubusercontent.com/assets/750477/18104125/390269e0-6ebf-11e6-83ec-69d5c0cd7fad.gif)

Note that there is a 'jump' from some dashboard threads loaded, to all threads loaded. I consider this very crappy.

I think the easiest thing to do here is, upon loading the dashboard for the first time, display a loading spinner until the actual dashboard query has come back and we're sure that it's fully populated.

PR #3664, opened 2016-08-30
Wait for dashboard to load before displaying threads

gdpelican:
For #3663


robguthrie:
seems good. Thanks @gdpelican 🍏

PR #3680, opened 2016-09-05
QA Checklist Template

robguthrie:
The intention is that this file provides a template, which you can paste into the discussion context of a QA check thread so that people can check off what they have tested, and we can see what level of QA we have achieved before pressing deploy.


robguthrie:
I'm going to merge this immediately, but I welcome additions or suggestions to it. Feel free to comment on this issue after I merge.

PR #3691, opened 2016-09-12
Iterate comment truncation

HSalmon:
PR on loomio_truncate_comment: https://github.com/loomio/loomio_truncate_comment


angular/core/components/thread_page/activity_card/new_comment.haml
```diff
@@ -26,7 +26,7 @@
   .thread-item__body.new-comment__body.lmo-markdown-wrapper{marked: "comment.cookedBody()"}
   %translation.thread-item__body{ng-if: "translation", translation: "translation", field: "body"}
 
-  %outlet{name: "after-comment-body", model: "comment"}
+  %outlet{name: "after-comment-body", model: "comment", aria-hidden: "true"}
```
gdpelican: Hmm. Since we can't be sure _what_ will go into this outlet (what if someone wants to put some other content in there, like a translation or something?), I don't think I agree with aria-hiding this.

Is it possible to throw the aria-hidden label within the plugin?

PR #3699, opened 2016-09-13
Fix ab testing issues with e2e tests. Support clearer segment recording.

robguthrie:


app/helpers/development/ninties_movies_helper.rb
```diff
@@ -26,7 +26,7 @@ def jennifer
                                email: 'jennifer_grey@example.com',
                                username: 'jennifergrey',
                                password: 'gh0stmovie',
-                               angular_ui_enabled: true)
+                               angular_ui_enabled: true).tap {|u| u.experienced! 'welcomeModal'}
```
gdpelican: Noice. 🍏

app/services/group_service.rb
```diff
@@ -1,6 +1,6 @@
 module GroupService
-  def self.ab_test_alternative?(group)
-    group.segment_seed % 2 == 0
+  def self.ab_testing_enabled?
+    Rails.env.production? or ENV['LOOMIO_AB_TESTING_ENABLED']
```
gdpelican: I'd lean towards just looking at ENV, rather than whether it's production or not. What's the likelihood a third party install is going to care about AB testing?

robguthrie: Yea.. walking home tonight, I suddendly realised that the production condition was wrong.. I thought to myself.. this is why we have code review!

app/services/group_service.rb
```diff
@@ -9,11 +9,16 @@ def self.create(group:, actor: )
     return false unless group.valid?
 
     if group.is_parent?
+      if ab_testing_enabled?
+        group.segments['bx_choose_plan'] = [true, false].sample
```
gdpelican: Things like this make me wish we'd made the push to extract our payment system into a plugin while we were doing the loomio_org plugin.

app/controllers/development_controller.rb
```diff
@@ -173,34 +174,20 @@ def setup_busy_discussion_with_signed_in_user
...
+    membership = Membership.find_by(user: patrick, group: group)
+    group.add_member! jennifer
     sign_in patrick
-    redirect_to group_url(test_group)
+    redirect_to group_url(group)
   end
 
-  def setup_group_with_overdue_trial
+  def setup_group_and_select_plan
+    test_group.segments['bx_choose_plan'] = true
```
gdpelican: Just a noticing that this feels quite similar to experiences; ie a hash associated with a model which is going to fill up with a bunch of (random?) values which affect how the app behaves. Not the same thing at all, but similar.

robguthrie: Yea.. cool noticing. I'm on the fence.. maybe slightly in favour of combining them actually. It seems like ab tests are about experiences... and if we were to do this.. then we don't have more BX specific schema on the models, and we get to extend experiences to Group.

PR #3711, opened 2016-09-13
Admin chargify subscription link

HSalmon:
[Trello card](https://trello.com/c/2BjWcOra/1872-2-link-to-chargify-subscription-page-from-group-loomio-admin-page)

chargify_subscription_url


app/admin/subscriptions.rb
```diff
@@ -14,6 +14,24 @@
...
+    attributes_table do
+      row :id
+      row :kind
+      row :expires_at
+      row :trial_ended_at
+      row :activated_at
+      row :chargify_subscription_id do |subscription|
+        if subscription.chargify_subscription_id
+          link_to subscription.chargify_subscription_id,
+                  ChargifyService.new(subscription.chargify_subscription_id).subscription_url, target: '_blank'
```
gdpelican: I don't think I quite see the merit of including ChargifyService in this. Options:

- Create a helper method, `chargify_subscription_url(subscription)`
- Put a `chargify_url` method on subscription.rb, `subscription.chargify_url`
- Create a SubscriptionDecorator which wraps a subscription `subscription.decorate.chargify_url` (most correct, but we don't do this anywhere else yet.)

I'd reckon #1 would be best for now.

app/helpers/subscription_helper.rb
```diff
@@ -0,0 +1,6 @@
+module SubscriptionHelper
+
+  def chargify_subscription_url(subscription)
+    ChargifyService.new(subscription.chargify_subscription_id).subscription_url
```
gdpelican: Heh, I think you misunderstood me; what I was thinking was

```
def chargify_subscription_url(subscription)
  return unless Rails.application.secrets.chargify_app_name
  "http://#{Rails.application.secrets.chargify_app_name}.chargify.com/subscriptions/#{subscription.chargify_id}"
end
```

That way we don't have to add those methods into the service.

gdpelican: You could even pull the whole link_to out into a helper

```
def chargify_subscription_url(subscription)
  return unless Rails.application.secrets.chargify_app_name
  link_to subscription.chargify_id, "http://#{Rails.application.secrets.chargify_app_name}.chargify.com/subscriptions/#{subscription.chargify_id}"
end
```

HSalmon:
@gdpelican, I think I've implemented your feedback :)

gdpelican:
Looks good to me, thanks! 🍏  I just put in a check to make sure we don't show the link if there's no subscription id

PR #3722, opened 2016-09-16
minor: fix typo that changes meaning (not vs. now)

azul:
you're now ready to install Loomio :sunflower:


robguthrie:
Thank you @azul!

PR #3724, opened 2016-09-18
[WIP] Watching

robguthrie:
I'm having a go at implementing a system of notfiying people based on if they are watching a thread. (rather than volume)

There will be a `uses_watching` column on group, so that we can enable this notification model per group for testing purposes.

Design doc here:
https://hackmd.loomio.io/MwM2E4HYoDgWgCYIGyTgFgeAjHARgKbACsckAxuMssQEx7IjYhA=


robguthrie:
closing this for now.. Going to explore another option for solving the same problem

PR #3735, opened 2016-09-20
Don't serialize out attachments / mentions for dashboard / inbox

gdpelican:
Currently our dashboard and inbox queries are getting killed by loading up a bunch of attachments:

**Dashboard**
screen shot 2016-09-20 at 5 45 47 pm

**Inbox**
screen shot 2016-09-20 at 5 46 52 pm

Over 55% of the time spent in the request is used doing an N+1 query for attachments which we don't use on these pages at all.

Here I've introduced a pair of new serializers for use with dashboard / inbox which ignores attachments (and, as a bonus, @mentions, which also incur a db hit and we also don't care about) for Discussions and Motions, which should halve the time it takes to process these queries. Wee performance!

On the plus side, it looks like folks are getting good mileage out of the discussion attachments feature, with about 2500 discussion attachments since we rolled out the feature June 9th, or about 15% of the 16.8k discussions created in that time. (There have been 90 motion attachments in that time, or 1% of the 7.5k motions created.)

### DIFF
--- a/app/controllers/api/discussions_controller.rb
+++ b/app/controllers/api/discussions_controller.rb
@@ -12,13 +12,13 @@ class API::DiscussionsController < API::RestfulController
   def dashboard
     raise CanCan::AccessDenied.new unless current_user.is_logged_in?
     instantiate_collection { |collection| collection_for_dashboard collection }
-    respond_with_collection
+    respond_with_collection(serializer: Dashboard::DiscussionSerializer)
   end
 
   def inbox
     raise CanCan::AccessDenied.new unless current_user.is_logged_in?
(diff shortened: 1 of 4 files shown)


gdpelican:
As another note, it looks like we have attachment N+1 issues in a lot of queries in the app, including notifications and the event feed. These are the most prominent queries, but I'm thinking it might make sense to not include attachments with Discussions, and instead either bundle them up specifically when we're doing discussion#show, or grab them in a separate request (maybe along with all the attachments for that discussions's comments and proposals), since they only show up on the thread page itself.

robguthrie:
Awesome. You're sooo good at AMS stuff. Thanks James. 🍏

PR #3738, opened 2016-09-22
[wip] Migrate user service

HSalmon:
@gdpelican, would be good to get your thoughts on this one. Totally open to it not being the right way to go, mostly concerned about risk to other data in the database. Interested in strategies we can use to test this to make sure it doesn't have side effects [Rob]


app/services/migrate_user_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    schema.each_pair do |table, columns|
+      Array(columns).map(&:to_s).each do |column_name|
+        lines << "UPDATE #{table} SET #{column_name} = #{@new_id} WHERE #{column_name} = #{@old_id}"
+      end
+    end
+    lines
+  end
+
+  def update_paperclip_versions_sql
+    "UPDATE versions SET whodunnit = '#{@new_id}' WHERE whodunnit = '#{@old_id}'"
```
gdpelican: How would this be different than adding `{version: :whodunnit}` to end of the schema? Also, what does this have to do with paperclip? It seems this will update all versions whether the models are attachments or not.

app/services/migrate_user_service.rb
```diff
@@ -0,0 +1,101 @@
+class MigrateUserService
+  def initialize(new_id:, old_id:)
+    @new_id = new_id
+    @old_id = old_id
+    @lines = []
+  end
+
+  def schema
```
gdpelican: I'm not sure I follow the name 'schema' here. What about `columns_to_transfer`?

app/services/migrate_user_service.rb
```diff
@@ -0,0 +1,101 @@
+class MigrateUserService
+  def initialize(new_id:, old_id:)
+    @new_id = new_id
+    @old_id = old_id
+    @lines = []
```
gdpelican: Looks like @lines isn't used

app/services/migrate_user_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    schema.each_pair do |table, columns|
+      Array(columns).map(&:to_s).each do |column_name|
+        lines << "DELETE FROM #{table} WHERE #{column_name} = #{@old_id}"
+      end
+    end
+    lines
+  end
+
+  def commit!
+    update_sql.each do |line|
```
gdpelican: I can't quite follow this. So we're updating all of the schema columns from the old id to the new id, then deleting all the schema columns with the old id (which should have been moved already?). Do we need to do both of these things?

db/migrate/20160922023739_add_unique_constraint_to_memberships.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueConstraintToMemberships < ActiveRecord::Migration
+  def change
+    add_index :memberships, [:group_id, :user_id], unique: true
```
gdpelican: We should check in prod and make sure we don't have any memberships which are violating this constraint currently.

app/services/migrate_user_service.rb
```diff
@@ -0,0 +1,101 @@
...
+      group.motions.each do |motion|
+        motion.update_votes_count
+        motion.update_voters_count
+        motion.update_vote_counts!
+      end
+    end
+  end
+
+  def update_version_owners
+    PaperTrail::Version.where(item_type: ['Discussion', 'Comment', 'Motion'], whodunnit: "#{@old_id}").update_all(whodunnit: "#{@new_id}")
```
gdpelican: I don't think you need this if the SQL in `update_paperclip_versions_sql` is run

app/services/migrate_user_service.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+    end
+    lines
+  end
+
+  def commit!
+    update_sql.each do |line|
+      begin
+        ActiveRecord::Base.transaction do
+          ActiveRecord::Base.connection.execute line
```
gdpelican: I think all of the SQL that gets run in this method should be bundled up in a transaction; we _really_ don't want a half-complete user migration to happen.

app/services/migrate_user_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    @lines = []
+  end
+
+  def schema
+    {
+      visits: :user_id,
+      ahoy_events: :user_id,
+      ahoy_messages: :user_id,
+      groups: :creator_id,
+      memberships: [:user_id, :inviter_id],
```
gdpelican: How will you handle the case of the @old_id and @new_id already belonging to the same group? Seems like that would violate the new [:user_id, :group_id] constraint on Membership

robguthrie: That's why we catch errors and keep each SQL in a seperate transaction. So in the case where there is a constraint violation, we can safely ignore it.

app/services/migrate_user_service.rb
robguthrie: Yeah.. I reckon you're right

app/services/migrate_user_service.rb
robguthrie: Yeah.. it will.. Is that ok?

app/services/migrate_user_service.rb
robguthrie: In the case where the update fails (eg: old and new user belonged to the same group) the update for memberships would have failed.. so there would be a dangling membership record unless we delete it.

app/services/migrate_user_service.rb
robguthrie: My approach to this was to allow some updates to fail, as they must be violating constraints.. in which case they get swept up in the delete afterwards.

robguthrie:
Thanks for your review James. I'm happy to talk it over and also to consider other approaches. I realise it's a kind of crazy way to update records.. but I think it's also worth considering it.

app/services/migrate_user_service.rb
gdpelican: OoOh. Okay, I think I understand better now.

app/services/migrate_user_service.rb
gdpelican: Yeah, seems good 🍏

gdpelican:
Looks sweet to me. Didn't grok that the delete was to sweep up transactions, seems alright, and it's tested yay! 🍏

robguthrie:
Wow.. Almost too supportive :) Thanks James.

I think we need to consider testing for "side effects" ie: create a bunch of records that should remain unchanged and verify that that is the case.

app/services/migrate_user_service.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def update_paperclip_versions_sql
+    "UPDATE versions SET whodunnit = '#{@new_id}' WHERE whodunnit = '#{@old_id}'"
+  end
+
+  def delete_sql
+    lines = []
+    schema.each_pair do |table, columns|
+      Array(columns).map(&:to_s).each do |column_name|
+        lines << "DELETE FROM #{table} WHERE #{column_name} = #{@old_id}"
```
robguthrie: add limit 1

PR #3745, opened 2016-09-29
More intuitive thread querying for group page [WIP]

gdpelican:


gdpelican:
This is alright, but I think I can do better with a more fully realized think about pagination.

PR #3746, opened 2016-09-29
Cancel subscription when group archived

HSalmon:
[Trello card](https://trello.com/c/NYu7KACQ/1917-8-people-deactivate-their-account-thinking-that-it-will-cancel-their-subscription)


app/services/group_service.rb
```diff
@@ -41,6 +41,7 @@ def self.update(group:, params:, actor:)
 
   def self.archive(group:, actor:)
     actor.ability.authorize! :archive, group
+    SubscriptionService.new(group, actor).end_subscription!
```
gdpelican: Should be in the event_bus listeners, listening for `group_archive` I reckon

HSalmon: It was initially, but it meant the event was being fired after the record was deleted and we need to reference the record to get the chargify_subscription_id. We need to call end_subscription! before the group is archived.

HSalmon: Could we create a before hook or something for the event bus?

gdpelican: Hm, we definitely shouldn't be deleting either the Group or Subscription record in `archive!`; it looks like we're just setting the `archived_at` date on the group, so not sure why you wouldn't be able to find it again.

HSalmon: Hmmm, there was an issue with permissions. I'm happy to take a look again

HSalmon: I think, from memory, it was because the memberships are archived therefore you don't have permission to change the subscription...or something like that

gdpelican: Looks like it bypasses the perm check if you don't pass in an actor to `SubscriptionService#new`

```
def initialize(group, actor = nil)
  actor.ability.authorize!(:choose_subscription_plan, group) if actor
```

HSalmon:
@gdpelican - latest commit incorporates your feedback

config/secrets.yml
```diff
@@ -34,8 +34,8 @@ development:
   <<: *default
   secret_key_base: asjdahkjda
   devise: abc123
-  chargify_app_name: loomio-test
-  chargify_api_key: fakeapikey
+  chargify_app_name: <%= ENV['CHARGIFY_APP_NAME'] %>
```
gdpelican: You can actually just take those two lines out; we'll inherit the ENV thingies from the default block above

.env.development
```diff
@@ -1,8 +1,8 @@
 # These are the default development settings.
 # To make changes to your specific instance, please modify .env
 
-CANONICAL_HOST=localhost
-CANONICAL_PORT=3000
+# CANONICAL_HOST=localhost
+# CANONICAL_PORT=3000
```
gdpelican: What was the reason for removing these? (I don't know either why they're there, or why we'd take them out :D)

HSalmon: It's because we had to set up an ssh tunnel to test out the chargify stuff in development. I can remove that stuff

config/secrets.yml
HSalmon: Can do

HSalmon:
This probably needs testing on clone or test (whichever is setup to handle chargify-related stuff). Are you up for that @gdpelican?

app/services/group_service.rb
```diff
@@ -42,6 +42,6 @@ def self.update(group:, params:, actor:)
   def self.archive(group:, actor:)
     actor.ability.authorize! :archive, group
     group.archive!
-    EventBus.broadcast('group_archive', group, actor)
+    EventBus.broadcast('group_archive', group)
```
gdpelican: We should keep this line here, because there may be some other thing that comes along which needs to know about who archived the group.

gdpelican:
🍏

Issue #3737, opened 2016-09-20
Loading discussion for print should not appear when entire discussion is already loaded


gdpelican:
If I have a discussion with less than 30 events, I shouldn't have to wait for all of the events to re-load before the print dialog appears. We should check to see if the firstSequenceId and lastSequenceId events are loaded before displaying the loading message

Issue #3736, opened 2016-09-20
Loading discussion message shouldn't be printed


gdpelican:
Printing on Chrome OSX leaves the 'loading discussion' message right in the middle of the printed page, which is less than ideal.

![screen shot 2016-09-20 at 7 36 21 pm](https://cloud.githubusercontent.com/assets/750477/18683777/9803a552-7f69-11e6-81b2-a2c82cdd8718.png)

gdpelican:
Looks like this is only the case when there's no activity, which makes it pretty edge case. Anyway the modal should have 'lmo-no-print' on it to prevent it from ever showing.

PR #3748, opened 2016-09-30
Only load records if they are not already loaded

HSalmon:
Addresses https://github.com/loomio/loomio/issues/3737 and https://github.com/loomio/loomio/issues/3736

### DIFF
--- a/angular/core/models/discussion_model.coffee
+++ b/angular/core/models/discussion_model.coffee
@@ -97,6 +97,9 @@ angular.module('loomioApp').factory 'DiscussionModel', (DraftableModel, AppConfi
       item = _.max @events(), (event) -> event.sequenceId or 0
       item.sequenceId
 
+    allEventsLoaded: ->
+      @recordStore.events.find(discussionId: @id).length == @itemsCount
+
     membership: ->
       @recordStore.memberships.find(userId: AppConfig.currentUserId, groupId: @groupId)[0]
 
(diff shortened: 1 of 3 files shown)


angular/core/models/discussion_model.coffee
```diff
@@ -97,6 +97,9 @@ angular.module('loomioApp').factory 'DiscussionModel', (DraftableModel, AppConfi
       item = _.max @events(), (event) -> event.sequenceId or 0
       item.sequenceId
 
+    allEventsLoaded: ->
+      @recordStore.events.find(discussionId: @id).length == @itemsCount
```
gdpelican: Ah, interesting, hadn't thought of it that way. This totally works though 🍏

gdpelican:
Another couple nitpicks:

- Hitting 'ctrl+P' (as opposed to selecting print from the dropdown) opens the print window immediately without loading the discussion. Wonder if there's a way we can override that? (Don't worry about that for this PR)
- I'd love to see a loading spinner next to the loading entire discussion message, and maybe some additional messaging like 'this may take a little while depending on the size of the discussion); it looks like a good-sized discussion takes about 7 seconds or so to load fully.

gdpelican:
Another thought to worry about later: We don't load all proposals for this, so if someone were printing out a discussion because they wanted a list of all the decisions which had been made, there's a chance they wouldn't get them.

HSalmon:
screen shot 2016-10-03 at 12 21 44 pm

@gdpelican

gdpelican:
🍏

PR #3749, opened 2016-10-02
Fix motion close

robguthrie:
temporary fix for https://errbit.loomio.io/apps/57215524657272000e000001/problems/57eb6ada657272003a0002ca

until we merge "merge user accounts" and fix the constraint issue.


gdpelican:
:green_apple:

PR #3752, opened 2016-10-03
Implement accessibility suggestions

HSalmon:
[Trello card](https://trello.com/c/pBR8ELIP/1976-accessibility-tweaks)


robguthrie:
This looks like a good start, we'll need to extract the english into translation strings though.

It strikes me that we could do this without defining ariaLabel for each modal. For instance if we add 
```
angular_welcome_modal:
  arial_label: "Welcome Dialog"
```

to each of the translation blocks.

Then we could add
```
arialLabel: $translate.instant(snake_case_name + ".aria_label")
```
into the currentModal =  bit in `modal_service.coffee`

We just need a method to return a snake_case version of the modal name

robguthrie:
Happy to pair on this with you tomorrow.

HSalmon:
Thanks for your suggestions, one thing that made me unsure about this approach is that we don't have access to the modal name at that point

HSalmon:
One thing I was thinking I could do is use regex to get the modal name from the templateUrl, then it would already be snakecase too.

robguthrie:
yea. That's what I'd do too. If you wanna have a go at that I'd recommend
something like http://regexr.com/ to test out your regex.

robguthrie:
🍏

PR #3755, opened 2016-10-04
Refactor start group controller

gdpelican:
The start group action is currently one of the slowest routes in our app:

screen shot 2016-10-03 at 8 02 52 pm

Which sucks because it's right at the front door of the Buyer experience.

This is a refactor to kick off the actual group creation and admin invitation (which are relatively slow and pretty safe to assume successful) into a side job, so that the request itself can be quick like 🐰 .

### DIFF
--- a/app/controllers/start_group_controller.rb
+++ b/app/controllers/start_group_controller.rb
@@ -8,9 +8,7 @@ class StartGroupController < ApplicationController
 
   def create
     if group_with_creator.errors.empty?
-      GroupService.create(group:  group_with_creator.group,
-                          params: group_with_creator.creator,
-                          actor:  current_user_or_visitor)
+      GroupService.create(group:  group_with_creator.group, actor: group_with_creator.creator)
     else
       render :new
     end
--- a/app/controllers/start_group_controller.rb
+++ b/app/controllers/start_group_controller.rb
@@ -21,6 +21,6 @@ class StartGroupController < ApplicationController
   end
 
   def redirect_to_dashboard
-    redirect_to_dashboard_path(start_group: true) if current_user_or_visitor.is_logged_in?
+    redirect_to dashboard_path(start_group: true) if current_user_or_visitor.is_logged_in?
   end
 end
--- a/app/controllers/start_group_controller.rb
+++ b/app/controllers/start_group_controller.rb
@@ -7,8 +7,8 @@ class StartGroupController < ApplicationController
   end
 
   def create
-    if group.with_creator.errors.empty?
-      StartGroupJob.perform_later(group_with_creator)
+    if group_with_creator.errors.empty?
+      StartGroupJob.perform_later(params)
     else
       render :new
     end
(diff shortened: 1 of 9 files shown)


gdpelican:
Would be good to get a CR of this one @robguthrie

robguthrie:
Hrrrmm.. 

I like most of the refactor here. The code seems to be improving. But I'm not sure we should background record creation. I'm happy to background stuff that sends emails and connects to external things, but I feel that we should be able to start groups inline.

I'm not totally opposed to this but ... I don't want to see us doing this over the app.. Record creation usually needs to complete before we can respond.

It would be interesting to dig into it a little further and find out what about group creation is slow.. is it the counters? Do we need indexes?  

I guess I'm keen to make the system sweat and find out why rather than push the problem to the background.

gdpelican:
Cool, I've updated it to do that, as well as putting the name / email stuff into a LoggedOutUser, which gets passed around nicely through our event system as the actor.

gdpelican:
I'm trying to figure out why our create and update actions are so slow. It's particularly bad for groups (because we need to set up a default group cover, a subscription, and example content), but I think the culprit is something else, because even routes that don't really do anything are taking a while, like Discussions#create or Votes#update

Maybe it's taking a long time to put stuff in the Event table? I also wonder if it's an anti-pattern to be responding with the event, because it would be much faster to do something like

```
thing is created
creation triggers side jobs to handle side effects
respond with thing
```

rather than what we do now:
```
thing is created
event is created
respond with event
```

But I want to hold off on doing anything until I'm more sure of why things are slow.

gdpelican:
Also, having looked a little into our angular performance, it may be more useful for me to just spend more time there. Realistically we're looking at shaving maximum 400ms off of some of the API calls, whereas we could get similar gains (and potentially more) by figuring out why the digest cycle is running so often on the frontend.

I'll spin up a profile once I feel more confident about what I'm seeing, but we're doing something like spinning a $digest 20 times on initial page nav, each of which takes between 75 - 175ms, which as a result makes the page load feel mega slow.

robguthrie:
Re: Returning events. I've though about that a lot, and came to the conclusion that returning events was the right way to reduce extra work - You have already have a reference to the _thing_ because you passed it into the service. Where as you need a reference to the event to continue with the work, if we did not pass it out, then we would have to do another find.

If we don't return the event, then I'm tempted to suggest that we return nothing, rather than a reference to an object you already have a reference to.

Anyway, I'm open to your thoughts, but it does not seem relevant to perf unless we change it to involve more work.

---

Re: Is it slow to insert events? I believed it could be (large table with many indices), but I looked into NR and I don't believe it is right now:
image

---

I'd be happy to help understand angular performance. I think one area we might find room for improvement is when we import recrds. If we could pause the digest loop until the import is complete we'd reduce the digest cycles over data that's about to change anyway.

robguthrie:
🍏

robguthrie:
Also I'm keen to try some kind of memoizing function over the whole record store so that repeat digest cycles are way faster than the first

gdpelican:
Ya I haven't found a very digestable way to get more granular than that controller method, which is taking 1.5 seconds, which is way too long for the ruby code portion of the thing. One of these days we'll get kcachegrind to show us something we understand hopefully.

PR #3759, opened 2016-10-05
Allow inclusion of plugin assets on rails views

gdpelican:


gdpelican:
It would be really good for @HSalmon to give this a quick go (there are instructions in the PR), to see if it meets the use case we're looking for

HSalmon:
@gdpelican, happy to

PR #3764, opened 2016-10-07
[wip] Move assets to plugin

HSalmon:


HSalmon:
@gdpelican, can we talk about what remains to be done at some point?

gdpelican:
If you deploy to clone I'll smoke test it and we can merge.

HSalmon:
I'm using clone for group activation card, but will deploy if I can get sign-off on that feature

PR #3765, opened 2016-10-07
Delay bulk mail jobs

gdpelican:


gdpelican:
Hm, I'm a little nervous about this one, but don't know why. Happy to merge if you're happy with it @robguthrie (and I've already set BULK_EMAIL_DELAY to 60 sec in prod, so this will take effect on deploy)

config/initializers/event_bus.rb
```diff
@@ -22,7 +22,8 @@
 
   # send bulk emails after events
   Event::BULK_MAIL_KINDS.each do |kind|
-    config.listen("#{kind}_event") { |event| SendBulkEmailJob.perform_later(event.id) }
+    delay = ENV.fetch('BULK_MAIL_DELAY', 0).to_i.seconds
+    config.listen("#{kind}_event") { |event| SendBulkEmailJob.set(wait: delay).perform_later(event.id) }
```
robguthrie: What is the default priority? I'd like it to be top priority after the delay.

robguthrie: actually.. why don' we pass in the event rather than the event_id? then we would not have to have this fuzzy fail thing...right?

app/jobs/send_bulk_email_job.rb
```diff
@@ -1,6 +1,6 @@
 class SendBulkEmailJob < ActiveJob::Base
   def perform(event_id)
-    event = Event.find(event_id)
+    return unless event = Event.find_by(id: event_id)
```
robguthrie: then over here we could test if the event still exists?

robguthrie:
Yea.. something is fishy about this. Sorry my comments are scattered.

Honestly.. I wonder if we should stop this line of work, and move to "archive" comments rather than delete, then test if comment is archived before sending.

config/initializers/event_bus.rb
gdpelican: ActiveJob can't pass objects, only primitive types.

gdpelican: Actually, that might not be quite right. But it does throw DeserializationError when trying to reconstruct the event object, so finding it by id was easier.

config/initializers/event_bus.rb
```diff
@@ -22,7 +22,8 @@
 
   # send bulk emails after events
   Event::BULK_MAIL_KINDS.each do |kind|
-    config.listen("#{kind}_event") { |event| SendBulkEmailJob.perform_later(event.id) }
+    delay = ENV.fetch('BULK_MAIL_DELAY', 0).to_i.seconds
```
gdpelican: Yep. We have `LOOMIO_AB_TEST`, but I think I prefer `LMO_BULK_MAIL_DELAY` ??

PR #3769, opened 2016-10-09
Remove before_save callback checking for group creator

gdpelican:


gdpelican:
I'd super-love a CR for this, @robguthrie 

The reason I got in here was that I want to get rid of any `before_` and `after_` filters in `group.rb`, which are really slowing down group creation (and, in some cases, saving the group in general, which is something we do a lot.)

That memberships query that we saw in NewRelic being run dozens of times across several requests, was because of this line:

```
self[:creator_id] || admins.any?
```
which is called in a before_save callback, meaning it gets run quite often, and hits the DB every time if `creator_id` isn't set.

robguthrie:
YIIKES! I've been guilty of this before. I would love to add it to our QA code review section: Don't do any database hits in before_ or after_ queries. Bad rob.

I'll review today.

robguthrie:
And the note would say: Don't do db stuff in before and after save, instead put it into the service.

gdpelican:
I would go so far as to say 'don't write before or after callbacks on the model, unless you're _really_ sure they need to happen _every_ time the model is manipulated.

robguthrie:
Yep. And maybe include reasoning about how hard it is to 'see' them because there is not a procedural order..

app/models/group.rb
```diff
@@ -369,8 +335,7 @@ def discussion_private_default
 
   def add_member!(user, inviter=nil)
     begin
-      save!
-      memberships.find_or_create_by(user: user) { |m| m.inviter = inviter }
+      tap(&:save).memberships.find_or_create_by(user: user) { |m| m.inviter = inviter }
```
robguthrie: changing from save! to save changes behaviour. recommend you keep save!

gdpelican: Good call. Not to mention we definitely don't want to be trying to add members to an invalid group, which seems like a really scary path to be going down.

Issue #3383, opened 2016-05-24
Links with @ in them get eaten and poorly digested 


rdbartlett:
I want to include this link in a comment but whichever way I try it gets borked. 

“Digital Tools Open Up Taiwan’s Democratic Imaginations” @audreyt https://medium.com/@audrey.tang/digital-tools-open-up-taiwans-democratic-imaginations-d8f80432305c

rdbartlett:
![screenshot_2016-05-24-09-15-47](https://cloud.githubusercontent.com/assets/970124/15495480/82b38d8e-2190-11e6-8f16-6b3f08fc0b68.png)

audreyt:
Workaround 1: https://medium.com/%40audrey.tang/digital-tools-open-up-taiwans-democratic-imaginations-d8f80432305c

Workaround 2: https://medium.com/p/d8f80432305c :-)

rdbartlett:
Haha whoops I didn't mean to mention you but thanks for the tip!

gdpelican:
I haven't said this before out loud, but have often thought it and will say it now for transparency's sake:

I have very little desire to muck around inside or around our markdown engine, because I feel that our markdown is a) pretty good-enough, and b) not used all that often by our users (ref mhjb's charts on the matter). So while I acknowledge that these kinds of things are bugs, they're not the kind of bugs I'm all that interested in fixing, short of several pointed complaints from non power users. <3

gdpelican:
I looked at this again. I think I want to write a `Mention` model, which will hang between a comment and a user, to resolve this, but it's too much work to prioritize right now (most of the work is around backwards compat). However, there's some perks to writing including:

- resolving #2782 
- allowing for some quite interesting metabase queries around who's mentioning whom (which sounds like something I'd like to unleash a dataviz gal / guy on)

gdpelican:
Turns out this isn't the case anymore, hooray our longest standing bug is now fixed!

PR #3780, opened 2016-10-12
Fix links with @ mentions in them

gdpelican:
For #3383


robguthrie:
don't know if this is good or not. happy to trust you.

gdpelican:
Yeah I'm not totally sure myself. It makes sense to me that mentions should have a space character in front of them in order to be mentions (this of course precludes links with @ in them), but I wonder if I'm missing some thing here.

gdpelican:
I'm closing this; links with @ don't feel important enough to risk if this breaks things.

PR #3781, opened 2016-10-12
Group activation progress bar

HSalmon:
Enable group activation progress card
[loomio_group_progress_card repo](https://github.com/loomio/loomio_group_progress_card)


angular/test/protractor/helpers/page_helper.coffee
```diff
@@ -40,9 +40,6 @@ module.exports = new class PageHelper
   findFirst: (selector) ->
     element.all(By.css(selector)).first()
 
-  clickLast: (selector) ->
```
gdpelican: I think we still need this; I can see a couple tests which still use this function

HSalmon: There were two identical versions of it in the file

gdpelican:
Ah, gotcha.

angular/core/models/group_model.coffee
```diff
@@ -169,3 +178,10 @@ angular.module('loomioApp').factory 'GroupModel', (DraftableModel, AppConfig) ->
 
     isSubgroupOfSecretParent: ->
       @isSubgroup() && @parent().privacyIsSecret()
+
+    setupComplete: ->
```
gdpelican: Either is ok with me. I've been curious about decorating JS models for a little bit now, but haven't had the occasion to get it to work yet.

HSalmon: Happy to make that change. Tests don't seem to be working unfortunately - would you mind walking me through that?

gdpelican:
Hey @HSalmon I've got a couple nitpicks (these are in reference to the plugin code, but there's not a PR there so no easy place to comment)

- I'm on the fence a bit as to whether to make this a membership experience, rather than a user one. It's kinda unintuitive to dismiss a card once, and then never show it again for any other group that you're a part of... seems like we'll get the maximum bang for our buck by showing this to any user that has an 'incomplete' group where they haven't dismissed the card yet.
- The 'dismissed' option with `display: none` is a little odd. Did you try
```
show: ->
  user.hasExperienced('dismissGroupCard')
```
?

- We can't merge this with loomio.school in the card link, since loomio.school isn't good yet. Suggest swapping out that link for a blog post of some kind. @rdbartlett?

PS: In the buyer experience branch, I've added support for adding ENV to a plugin's config. So in `config.yml`, you could write:

```
loomio_school_link: <%= ENV['LOOMIO_SCHOOL_LINK'] %>
```
And access it on the frontend with
```
AppConfig.pluginConfig('loomio_group_progress_card').loomio_school_link
```

We can do this after that buyer experience stuff is merged.

rdbartlett:
Yep we'll have a blog post ready soon. I'd say hold of deploying til we
have the article published.

gdpelican:
If a deploy's not imminent, I may port the ENV stuff over to this branch so we can store links in ENV now, because that is nifty. 

@HSalmon @robguthrie 
What are your thoughts on user vs membership experience? To be super clear,

A: User experience:
- I create a group, and dismiss the progress card. I never see the card ever again, no matter how many groups I start

B: Membership experience:
- Every time I create a group, I'm presented with a reminder of the 6 things I need to do in order to help activate my group.

A) strikes me as being slightly less annoying, but I think B) will result in way more group activations.

rdbartlett:
FWIW I had imagined it working like B.

robguthrie:
I think B is the right way to go too.

robguthrie:
I think you should use "URL" rather than Link, as a link is the gui element.

gdpelican:
B it is! Should be a pretty simple swap to

```
user.membershipFor($scope.group).hasExperienced('dismissGroupCard')
```

HSalmon:
Yeah, thanks folks, I'd actually already implemented this - just haven't pushed it up :)

PR #3783, opened 2016-10-13
Use cache queries instead of has_many for comment serialization

gdpelican:
Looking at improving the performance of the events#index route (the one which loads events for the activity card)

Here's a recent extra-long query (this one took 2,700 ms):
screen shot 2016-10-13 at 1 08 47 am

A huge chunk of that time was taken querying for attachments and users (likers, in this case)

To combat this, I've added some queries to the EventCollection class, which will allow us to make a single query for all likers, attachments, or mentions (3 queries total) in the specified events, rather than having to do single queries for each of them (this added up to ~50 queries in the request above)

### DIFF
--- a/app/models/event_collection.rb
+++ b/app/models/event_collection.rb
@@ -10,6 +10,32 @@ class EventCollection
   end
 
   def serialize!(scope = {})
-    Events::ArraySerializer.new(self, scope: scope).as_json
+    Events::ArraySerializer.new(self, scope: default_scope.merge(scope)).as_json
+  end
+
+  private
+
+  def default_scope
+    { cache: { likers: likers, attachments: attachments, mentions: mentions } }
+  end
+
+  def likers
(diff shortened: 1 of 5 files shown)


app/models/event_collection.rb
```diff
@@ -10,6 +10,32 @@ def any?
...
+
+  def likers
+    @likers ||= User.joins(:comment_votes)
+                    .select('users.*').select('comment_votes.comment_id')
+                    .where('comment_votes.comment_id': event_comment_ids)
+                    .group_by(&:comment_id)
+  end
+
+  def mentions
+    @mentions ||= Comment.find(event_comment_ids).map { |c| [c.id, c.mentioned_usernames] }.to_h
```
gdpelican: I believe this will allow us, in another iteration, to be smart about who we highlighting as mentions (ie, only highlight members of the group, or people who have actually been notified of the mention)

PR #3795, opened 2016-10-18
Tweak language + add character count

HSalmon:
[Trello card](https://trello.com/c/gD1klyvH/2046-2-tweak-ask-to-join-form-language-add-character-count)

screen shot 2016-10-18 at 3 33 38 pm

screen shot 2016-10-18 at 3 38 39 pm

### DIFF
--- a/app/models/membership_request.rb
+++ b/app/models/membership_request.rb
@@ -8,6 +8,8 @@ class MembershipRequest < ActiveRecord::Base
 
   validates :group, presence: true
 
+  validates_length_of :introduction, maximum: 250
+
   belongs_to :group
   belongs_to :requestor, class_name: 'User'
   belongs_to :user, foreign_key: 'requestor_id' # duplicate relationship for eager loading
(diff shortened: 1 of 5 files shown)


app/models/membership_request.rb
```diff
@@ -8,6 +8,8 @@ class MembershipRequest < ActiveRecord::Base
 
   validates :group, presence: true
 
+  validates_length_of :introduction, maximum: 250
```
gdpelican: I'm nervous about this - there's about 5000 records in the database that have more than 250 sized introductions. What will happen to them?

HSalmon: @robguthrie, @rdbartlett, thoughts?

rdbartlett:
The intention is to change user behaviour. 

The field currently says "Introduce yourself", so people are writing great introductions. But then there is no clean UX to take that introduction content and share it with the group. E.g. once you've approved their request to join, there's no link to get to the membership_requests page, where the intro content lives.

So we want to change this field so it is more like 'knock knock let me in'.

I don't know about the best way to handle the old records, but I would have thought the length validation is only going to get in the way when records are updated right?

app/models/membership_request.rb
gdpelican: Pretty sure you want this:
http://stackoverflow.com/a/28603293

Issue #3793, opened 2016-10-17
open new window when clicking "select" on a paid plan on the 'pick a plan' modal


alannallama:
Currently if you click 'select' on a paid option on the pricing plans modal screen, it opens in the current window and takes you out of your Loomio group. Should open the Chargify payment process in a new tab/window.

Issue #3792, opened 2016-10-17
Clicking "Let's talk" on plan select modal closes modal


mhjb:
… whereas it should stay open, because that's what people are wanting to talk to us via Intercom one.

PR #3796, opened 2016-10-18
Open chargify form in new tab, prevent modal from closing when contact form opens

HSalmon:
Attempt to address:
https://github.com/loomio/loomio/issues/3792
&
https://github.com/loomio/loomio/issues/3793

Could maybe do with a smoke-test on a test env with access to Chargify and Intercom


gdpelican:
This looks good to me. I've updated the new buyer experience plugin to reflect this code as well.

Issue #3033, opened 2016-02-29
Strip markdown for meta tags


mhjb:
![image](https://cloud.githubusercontent.com/assets/1380991/13406958/feff96d2-df8a-11e5-8a6f-79fc5f9a1eb7.png)

robguthrie:
Interesting. Maybe we can strip markdown for the meta tags.

gdpelican:
I'll take care of this in the logged out angular branch.

gdpelican:
Doesn't appear to be an out-of-the-box thingy for ruby, which is a little odd.

I think I'm going to port over this thingy from node:
https://github.com/stiang/remove-markdown

Yay making a gem!

mhjb:
Here's kinda the reverse issue — it would be nice if shareable links for public groups had the group's metadata:

![image](https://cloud.githubusercontent.com/assets/1380991/16864634/84bdb5b4-4ab2-11e6-9c91-48bf615b2567.png)

PR #3801, opened 2016-10-19
Strip markdown for meta tags

gdpelican:
For #3033 

I just found out Redcarpet comes bundled with a markdown stripper! 🍰 🎉


mhjb:
Worth doing this to search results too?

PR #3804, opened 2016-10-19
Fold inbox and notifications query into current user data [WIP|Experiment]

gdpelican:


angular/core/components/proposal_redirect/proposal_redirect_controller.coffee
```diff
@@ -1,11 +1,12 @@
...
   $rootScope.$broadcast('currentComponent', 'proposalRedirect')
   Records.proposals.findOrFetchById($routeParams.key).then (proposal) =>
     Records.discussions.findOrFetchById(proposal.discussionId).then (discussion) =>
-      params = {proposal: proposal.key}
+      params = {}
       if $location.search().position?
         params.position = $location.search().position
-      $location.url LmoUrlService.discussion discussion, params
+        $location.search 'position', null
+      $timeout -> $location.path LmoUrlService.proposal proposal, params
```
gdpelican: This broke and I don't know why. Pulling it out of the digest loop fixed it, tho

app/controllers/api/profile_controller.rb
```diff
@@ -54,7 +54,7 @@ def resource_serializer
     if current_user == restricted_user
       Restricted::UserSerializer
     else
-      CurrentUserSerializer
+      Full::UserSerializer
```
gdpelican: Since we have the concept of `Full` serializers now (ie, serializers which include child records), makes sense to move the CurrentUserSerializer under that umbrella.

app/models/current_user_data.rb
```diff
@@ -0,0 +1,43 @@
+CurrentUserData = Struct.new(:user, :restricted) do
+  def data
+    serializer.new(user, scope: serializer_scope, root: :current_user).as_json.tap do |json|
+      json[:current_user].except!(:group_ids, :membership_ids)
```
gdpelican: It interferes with our `groupIds()` method on the frontend if these get serialized out. (Normally when you use `has_many` with ActiveModel::Serializers, it includes the ids of the child record with the parent. We don't actually care about that though.)

app/models/current_user_data.rb
```diff
@@ -0,0 +1,43 @@
...
+    if restricted
+      Restricted::UserSerializer
+    else
+      Full::UserSerializer
+    end
+  end
+
+  def serializer_scope
+    { memberships: memberships }.tap do |hash|
+      hash.merge!(
```
gdpelican: Only serialize out unread threads, readers, and notifications if the user is fully logged in

gdpelican: Always serialize out memberships, because restricted users need em and logged out users don't have any.

gdpelican:
@robguthrie A CR here would be great.

Do you agree that rolling information into the initial payload is a good way to go? I think this'll shave some significant time off of our page load, but keen to test that hypothesis on clone once it's free.

app/models/current_user_data.rb
robguthrie: Looks good.. except .. I thought we did not have a concept of readers in the client. We mush readers into discussions eh? Would to mind refereshing me as to why we're sending readers over?

robguthrie:
Happy to give it a go. Looks like an improvement when we don't have to check if records are loaded in some places.

I was kind of expecting us to move towards more, smaller API requests, but happy to go this way if it makes sense.

Are you having a good time with the new AMS?

app/models/current_user_data.rb
gdpelican: Oh, yes, it's a bit odd, eh. When the DiscussionSerializer is doing it's thang, it checks the scope to see if there is a `:reader_cache`, and pulls readers from that, so that we can do one reader query for _all_ discussions being serialized, rather than one query for _each_ discussion being serialized (this is the same reasoning behind serializing out all memberships, or all notifications here, rather than passing them through `has_many`).

This scope is also shared by every model that is serialized, so if a user is serializing out a bunch of memberships, the UserSerializer and MembershipSerializer are looking at the same scope.

So, we're not actually serializing out discussion readers separately, just loading them together so that the DiscussionSerializer can pick through the ones that it wants without making a separate query.

(We need to do this because there's no inbox query anymore, so this is where we get all unread threads for a user now, and in order for the client to think they're unread it needs the reader data fused onto the discussion)

gdpelican:
I would have thought that too, and I won't deploy this without a pretty solid perf test on clone which says this is better, but since we're doing sequential requests on every page load, makes sense to at least try this approach.

robguthrie:
sounds great. 🍏

gdpelican:
We're still on ancient AWS, but I think we've figured out all the weird, undocumented back alleys towards 

A couple of attempts to upgrade to latest have been less than successful IIRC because embedding ids isn't very well supported

gdpelican:
Heh, a quick run of this on clone has this running _way_ slower because it's loading up, like, all unread discussions ever (1500 for me). I'll have to tweak and try again at some point.

PR #3807, opened 2016-10-20
optimise images

rdbartlett:
I ran ImageOptim across app/assets/images and losslessly saved 126.5KB oh my goodness!

No big deal, but I reckon it's worth doing from time to time.


gdpelican:
Hey @rdbartlett would you be willing to repeat this process for `public/img`? That's where all the images we use in-app live.

PR #3809, opened 2016-10-25
Serve correct background image for viewport size

gdpelican:


gdpelican:
Mind CR'ing this @robguthrie ? It's simple but I think it does the trick well enough for the first time around.

rdbartlett:
This makes me happy.

I wonder if we should take this opportunity to add more sizes? I'm especially thinking of HD screens, e.g a 13" Retina is 2560px wide.

angular/core/services/viewport_service.coffee
```diff
@@ -0,0 +1,10 @@
+angular.module('loomioApp').factory 'ViewportService', ($window) ->
+  new class ViewportService
+
+    width: ->
```
robguthrie: I'm uncomfortable about taking one "width" which is an int and returning another which is a string representing some size.

I think i'd more more happy with 'size' rather than 'width'.. or even approximateSize.

how does that sound?

robguthrie:
Extra sizes are a different problem.. but now much easier to solve..

PR #3811, opened 2016-10-25
locales update

natiloomio:


gdpelican:
Thanks @natiloomio ; we'll have to have a chat about how to get your process to handle translations in plugins at some point.

PR #3813, opened 2016-10-26
update qa checkist

robguthrie:
as discussed at last product meeting:
Test on production after deploy of code which touches core workflows.


gdpelican:
🍏

PR #3816, opened 2016-10-27
[Wip] Intro carousel

HSalmon:
[PR on loomio_group_progress_card plugin](https://github.com/loomio/loomio_group_progress_card/pull/2)


angular/core/components/group_page/group_page.haml
```diff
@@ -1,20 +1,21 @@
...
-      .lmo-group-column-right
-        %outlet{name: "before_group_column_right", model: "groupPage.group"}
-        %gift_card{group: "groupPage.group"}
-        %upgrade_plan_card{group: "groupPage.group"}
-        %membership_requests_card{group: "groupPage.group"}
-        %members_card{group: "groupPage.group"}
-        %group_previous_proposals_card{group: "groupPage.group"}
-        %subgroups_card{group: "groupPage.group"}
-        %group_help_card{group: "groupPage.group"}
+    %outlet{name: "before_group_page", model: "groupPage.group"}
```
gdpelican: Whoa, I don't think we can nest things inside outlets.

Also, I know it's a little strange, but our convention so far has been `string-of-beads` outlet names, rather than `train_cars`

gdpelican:
I had a good time playing with this on clone. A couple nit picks:

- There's still a dip on slide transition, but I was able to fix it by making the 'close' button absolutely positioned in the top right
- cursor: pointer on the 'jewels' at the bottom of the slider

gdpelican:
@HSalmon I'm just seeing one failing test after merging master in (looks like an old one). Care to check that out and then we'll merge this?

PR #3831, opened 2016-11-01
Add animations to toast message

gdpelican:
Looks like this: https://drive.google.com/open?id=0B7rFWNdVhCDFNVFXUmp0N0JUUFU

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -7,4 +7,3 @@ SMTP_DOMAIN=localhost
 SUPPORT_EMAIL="support <support@localhost>"
 # SMTP_SERVER=localhost
 # SMTP_PORT=1025
-FLASH_TIMEOUT_SHORT=1000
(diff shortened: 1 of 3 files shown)


gdpelican:
Hey @rdbartlett @mhjb , what do you think?

rdbartlett:
I think it looks fine, and the movement will likely make the message more noticeable.

I was expecting to see it work like it does in the spec though, i.e. popping up like toast: https://material.google.com/components/snackbars-toasts.html#snackbars-toasts-specs

mhjb:
+1 for Material-style

gdpelican:
Yeah I'm not planning on putting more Material into the app until we make a decision as a team that this is something we want to adopt more fully. Also the material components we have put in have been broken and not x-browser compatible.

mhjb:
Interesting — sounds like a useful conversation to have. But here I'm just thinking imitate Material in how the animation looks, rather than actually adding a Material component

PR #3846, opened 2016-11-04
missing translations for vote buttons

fauno:
we had a missing translation for the commitment alt button and then i noticed the others were in english :)


gdpelican:
Thanks @fauno !

PR #3857, opened 2016-11-08
enable data export from admin panel

robguthrie:
This adds a "toggle data export" button to the groups show page.

It also adds a features object much like the experiences object, for storing which features are enabled and any details we need about them.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -66,8 +66,7 @@ class Ability
     end
 
     can :export, Group do |group|
-      user.is_admin? or
-      (user_is_admin_of?(group.id) && group.enabled_beta_features.include?('export'))
+      user_is_admin_of?(group.id) && group.features['dataExport']
     end
 
     can [:members_autocomplete,
--- a/app/admin/groups.rb
+++ b/app/admin/groups.rb
@@ -151,6 +151,15 @@ ActiveAdmin.register Group do
       end
     end
 
+    panel 'Data export' do
+      export_enabled = group.features['dataExport']
+      form action: toggle_export_admin_group_path(group), method: :post do |f|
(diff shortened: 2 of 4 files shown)


robguthrie:
I've tested this. I don't think there is much to worry about.. @gdpelican can you tell me if this needs more QA or anything?

gdpelican:
Looks great to me. 

Do you think we could (would it be smart to, not should we do it right now) move some other flags, like whether a group is experimental or not, into features?

robguthrie:
yes, I do think we should do that. it's a nice way to keep plugin specific
stuff out of the main schema.

robguthrie:
deploying to clone now.

gdpelican:
What's the difference between 'features' and 'experiences' on groups?

robguthrie:
In my mind, features are things that the group can do or use. Like, which comment form widget they are using. Or if they have data export.

Experiences is where we record that the group has done something, like seen a modal, or pushed a button.

robguthrie:
Would could mix them up.. but I don't think that would be helpful in the long run.

gdpelican:
Sounds good to me

app/models/ability.rb
```diff
@@ -66,8 +66,7 @@ def initialize(user)
     end
 
     can :export, Group do |group|
-      user.is_admin? or
```
mhjb: @robguthrie how come you took this out?

robguthrie: i guess it was because it's an undocumented thing.

Happy to put it back in if you like. Sorry for the inconvenience.

mhjb: I understand user.is_admin = true means user has access to loomio.org/admin, eh?

robguthrie: yes

PR #3858, opened 2016-11-09
Improve mobile experience of thread page

HSalmon:
[Trello card](https://trello.com/c/VJFfuJnN/2069-8-improve-mobile-thread-page)
screen shot 2016-11-09 at 1 46 45 pm
screen shot 2016-11-09 at 1 46 55 pm

### DIFF
--- a/angular/core/components/thread_page/context_panel/context_panel.coffee
+++ b/angular/core/components/thread_page/context_panel/context_panel.coffee
@@ -1,0 +1,52 @@
+angular.module('loomioApp').directive 'contextPanel', ->
+  scope: {discussion: '='}
+  restrict: 'E'
+  replace: true
+  templateUrl: 'generated/components/thread_page/context_panel/context_panel.html'
+  controller: ($scope, $rootScope, $window, $timeout, AbilityService, Session, ModalService, ChangeVolumeForm, DiscussionForm, ThreadService, MoveThreadForm, PrintModal, DeleteThreadForm, RevisionHistoryModal) ->
+
+    $scope.showContextMenu = ->
+      AbilityService.canChangeThreadVolume($scope.discussion)
+
+    $scope.canChangeVolume = ->
+      Session.user().isMemberOf($scope.discussion.group())
+
+    $scope.openChangeVolumeForm = ->
+      ModalService.open ChangeVolumeForm, model: => $scope.discussion
+
+    $scope.canEditThread = ->
+      AbilityService.canEditThread($scope.discussion)
+
+    $scope.editThread = ->
+      ModalService.open DiscussionForm, discussion: => $scope.discussion
+
+    $scope.muteThread = ->
+      ThreadService.mute($scope.discussion)
+
+    $scope.unmuteThread = ->
+      ThreadService.unmute($scope.discussion)
+
+    $scope.canMoveThread = ->
+      AbilityService.canMoveThread($scope.discussion)
+
+    $scope.moveThread = ->
+      ModalService.open MoveThreadForm, discussion: => $scope.discussion
+
+    $scope.requestPagePrinted = ->
+      unless discussion.allEventsLoaded()
+        ModalService.open PrintModal, preventClose: -> true
+        $rootScope.$broadcast 'fetchRecordsForPrint'
+      else
+        $timeout -> $window.print()
+
+    $scope.canDeleteThread = ->
+      AbilityService.canDeleteThread($scope.discussion)
+
+    $scope.deleteThread = ->
+      ModalService.open DeleteThreadForm, discussion: => $scope.discussion
+
+    $scope.showLintel = (bool) ->
+      $rootScope.$broadcast('showThreadLintel', bool)
+
+    $scope.showRevisionHistory = ->
+      ModalService.open RevisionHistoryModal, model: => $scope.discussion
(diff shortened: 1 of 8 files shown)


angular/core/components/thread_page/context_panel/context_panel.coffee
```diff
@@ -0,0 +1,52 @@
...
+      ThreadService.unmute($scope.discussion)
+
+    $scope.canMoveThread = ->
+      AbilityService.canMoveThread($scope.discussion)
+
+    $scope.moveThread = ->
+      ModalService.open MoveThreadForm, discussion: => $scope.discussion
+
+    $scope.requestPagePrinted = ->
+      unless discussion.allEventsLoaded()
```
gdpelican: Hi I'm going to say this but you can ignore it if you like.

I'm not a fan of  the leading 'unless'; it doesn't read well in my head and gets the story of the method backwards. 'If' is optimistic; it says, 'this is what should happen, but if it doesn't we've got a backup'. Which is a much better sentence than 'unless it's sunny today, we will stay inside, otherwise we will go to the park'

So this imo should be
```
if allEventsLoaded
  # print
else
  # load and print
```

angular/core/components/thread_page/context_panel/context_panel.haml
```diff
@@ -0,0 +1,56 @@
+.blank
```
gdpelican: Why is the `.blank` necessary?

angular/core/components/thread_page/thread_page.haml
```diff
@@ -2,81 +2,44 @@
...
+        %outlet{name: "before-thread-page-column-right", model: "threadPage.discussion"}
+      .thread-page__members
+        %members_card{group: "threadPage.discussion.group()"}
+
+    .thread-page__main-content--large{ng-if: "threadPage.windowIsLarge"}
+      .lmo-thread-column-left
+        %context_panel{discussion: "threadPage.discussion"}
+      .lmo-thread-column-right
+        %outlet{name: "before-thread-page-column-right", model: "threadPage.discussion"}
+        %section.start-proposal-card.lmo-no-print{in-view: "threadPage.proposalButtonInView($inview)", in-view-options: "{debounce: 200}", ng-if: "threadPage.canStartProposal()", aria-label: "{{ 'start_proposal_card.title' | translate }}"}
```
gdpelican: Seems like we should pull this out into a directive as well, to avoid duplicating it.

angular/core/components/thread_page/thread_page_controller.coffee
```diff
@@ -1,6 +1,9 @@
-angular.module('loomioApp').controller 'ThreadPageController', ($scope, $routeParams, $location, $rootScope, $window, $timeout, Records, MessageChannelService, KeyEventService, ThreadService, ModalService, DiscussionForm, MoveThreadForm, DeleteThreadForm, ScrollService, AbilityService, Session, ChangeVolumeForm, PaginationService, LmoUrlService, TranslationService, RevisionHistoryModal, ProposalOutcomeForm, PrintModal) ->
+angular.module('loomioApp').controller 'ThreadPageController', ($scope, $routeParams, $location, $rootScope, $window, $timeout, $mdMedia, Records, MessageChannelService, KeyEventService, ModalService, ScrollService, AbilityService, Session, PaginationService, LmoUrlService, TranslationService, ProposalOutcomeForm) ->
   $rootScope.$broadcast('currentComponent', { page: 'threadPage'})
 
+  @windowIsSmall = $mdMedia('sm') || $mdMedia('xs')
+  @windowIsLarge = $mdMedia('gt-sm')
```
gdpelican: Why two variables? Why not

```
ng-if: "windowIsLarge"
ng-if: "!windowIsLarge"
```
?

angular/core/components/thread_page/context_panel/context_panel.coffee
HSalmon: Do you mind if I consider this out of scope for this feature and make the change soon?

angular/core/components/thread_page/thread_page_controller.coffee
HSalmon: Truue

angular/core/components/thread_page/thread_page.haml
HSalmon: Sorry @gdpelican, what are you referring to here? Proposals?

gdpelican: Ya this whole start proposal card is duplicated.

HSalmon:
@gdpelican, I think I've implemented all of yer recommended changes. Whaddya reckon - is this ready to go?

gdpelican:
Erg, alright, let's give it a go, although I have to say I have a bad feeling about this.

If I were to go in and give it another go, I'd probably look at getting rid of `lmo-thread-page-left` and `lmo-thread-page-right`, and instead simply have markup that looked like:

```
.lmo-main-content
  %context_card
  %activity_card
  %members_card
```

Then we could use the [flexbox order](https://css-tricks.com/almanac/properties/o/order/) property based on a media query to reorder elements, without needing to resort to javascript and duplicating DOM nodes.

HSalmon:
@gdpelican, my initial approach was to use flexbox order, but it proved more difficult than I'd anticipated.

PR #3873, opened 2016-11-16
add a basic count of activity in last 30 days for groups.

robguthrie:
This is requested by mhjb so we can sort explore page bettererer.

@HSalmon how would I tell the groups api controller to sort the results by recent_activity_count?


HSalmon:
Sorry, I think this is more of a @gdpelican question, but I think you could do something like what the api/discussions_controller does:

```def index
    load_and_authorize(:group, optional: true)
    instantiate_collection { |collection| collection.sorted_by_importance }
    respond_with_collection
end```


app/extras/queries/visible_discussions.rb

```def sorted_by_latest_activity
    @relation = @relation.order(last_activity_at: :desc)
    self
end```

File: app/jobs/count_group_recent_activity_job.rb [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+  queue_as :low_priority
+
+  def query
+    "UPDATE groups SET
+     recent_activity_count =
+      (5* (SELECT count(discussions.id)
+           FROM discussions
+           WHERE discussions.group_id = groups.id
+           AND discussions.created_at > current_date - 30)) +
+      (5 * (SELECT count(motions.id)
```
gdpelican: Why 5x here?

db/schema.rb
```diff
@@ -543,6 +544,7 @@
 
   add_index "memberships", ["created_at"], name: "index_memberships_on_created_at", using: :btree
   add_index "memberships", ["group_id", "user_id", "is_suspended", "archived_at"], name: "active_memberships", using: :btree
+  add_index "memberships", ["group_id", "user_id"], name: "index_memberships_on_group_id_and_user_id", unique: true, using: :btree
```
gdpelican: Ok with adding this index?

gdpelican:
@HSalmon Ya that's close.

I'm going to start making some documentation of some kind on how all this stuff works.

When we run `instantiate_collection`, it does 4 things (in this order:)

1. Gets all possible records visible to the user by calling `accessible_records`
2. Yields to the block given to `instantiate_collection` <- (this is where we do ordering, or where conditions, etc.
3. Applys a timeframe (since / until) to the records if timeframe params are given (we don't use this atm)
4. Applys a limit and offset to the records if they are given, default is offset: 0, limit: `default_page_size`

So to order things we'd want something like

```
def index
  instantiate_collection { |collection| collection.search_for(params[:q).order(recent_activity_count: :desc) }
  respond_with_collection
end
```

app/jobs/count_group_recent_activity_job.rb
robguthrie: it's made up. a starting point for fine tuning. I'm guessing that motions are 5 times more important than comments.

db/schema.rb
robguthrie: no, that's a mistake. Thanks for catching that.

robguthrie:
Thanks for that info @gdpelican. Sounds like we want an explore action so we can implement a sort without changing the behvaviour of index.

robguthrie:
I've finished this now. On deploy we need to add support for daily_tasks to be run every day.

robguthrie:
@gdpelican ok for merge?

gdpelican:
Yeah!

juliencarnot:
Sorry for digging this conversation out, was thinking about #4404 , but recent_activity_count is set to 0 for all groups I participate in...

@robguthrie mentionned "On deploy we need to add support for daily_tasks to be run every day.", was this set up?

PR #3877, opened 2016-11-16
Ensure events are first thing loaded by thread page

gdpelican:
Currently events are being loaded after the active proposal, previous proposal page, tags, etc, which is delaying the perceived load time of the page by ~500ms. This makes it so that events are the first query to go out after the thread itself is loaded.


gdpelican:
I've added a loading content thingy to this branch, which looks like this:

screen shot 2016-11-21 at 12 36 53 pm

PR #3882, opened 2016-11-17
Perf improvements for thread preview [WIP / maybe not that useful]

gdpelican:
I loaded up a full dashboard's worth of threads on my development instance, and got 3599 watchers. That's a lot! It's recommended to stay under 2000 to keep the digest loop snappy.

So, I made some improvements:
- A full dashboard with production code right now: 3599 watchers (-0)
- Assume user names and avatar urls will not change: 3331 watchers (- 268)
- Assume groups names and discussion titles will not change: 3181 watchers (-150)
- Translate thread preview translations once per collection, rather than once per thread: 2931 watchers (-250)
- Assume thread URL, and screen reader only information (title, author name) will not change: 2781 watchers (-250)

2781 / 3599 = 23% fewer watchers
However, somewhat disconcertingly, this didn't result in a $digest loop performance improvement (it hovered around 650 - 670ms both with and without these changes applied.)

Soooo... ¯\_(ツ)_/¯ ?

### DIFF
--- a/angular/core/components/thread_preview/thread_preview.coffee
+++ b/angular/core/components/thread_preview/thread_preview.coffee
@@ -1,5 +1,5 @@
 angular.module('loomioApp').directive 'threadPreview', ->
-  scope: {thread: '='}
+  scope: {thread: '=', translations: '='}
   restrict: 'E'
   templateUrl: 'generated/components/thread_preview/thread_preview.html'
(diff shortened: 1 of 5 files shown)


gdpelican:
I'll close this for now. I don't think it's improving much.

Issue #3874, opened 2016-11-16
Angry, jumpy character limit warning on position statements


alannallama:
This must have changed recently because I just encountered it for the first time...

<img width="570" alt="screen shot 2016-11-16 at 2 54 20 pm" src="https://cloud.githubusercontent.com/assets/5498878/20331853/f30346c2-ac0c-11e6-83fa-c548bb1c9dbd.png">

1) The size of the modal jumps when this message appears, which is distracting (especially because the user experience is to then go edit it shorter and try again, meaning it's jumping back and forth).
2) This text is really angry sounding and off putting (especially in red). This is a facilitation intervention and should be carefully worded and presented as such. I suggest something like "Character limit reached. Short position statements create a succinct summary. If you want to say more, post a comment".

mhjb:
In fact most of a year old… https://github.com/loomio/loomio/commit/e0f45edd7ccced10f62913ef6571f4214794c4f6

PR #3889, opened 2016-11-21
Make modal widths consistent, update charater limit message for vote form

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3874

screen shot 2016-11-21 at 1 47 22 pm

### DIFF
--- a/angular/core/components/thread_page/vote_form/vote_form.haml
+++ b/angular/core/components/thread_page/vote_form/vote_form.haml
@@ -1,26 +1,26 @@
-.vote-form
-  %form{ng-submit: "submit()",  ng-disabled: "vote.processing", name: "voteForm" }
-    .lmo-disabled-form{ng-show: "isDisabled"}
-    .modal-header
-      %modal_header_cancel_button
-      %h1.lmo-h1.modal-title{translate: "vote_form.heading"}
+%md-dialog.vote-form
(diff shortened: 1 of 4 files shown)


gdpelican:
I think this is going to require a little more thought. What about these modals?
screen shot 2016-11-21 at 10 02 24 am

screen shot 2016-11-21 at 10 03 02 am

screen shot 2016-11-21 at 10 04 26 am

HSalmon:
@gdpelican, I understand what you're saying but personally I'm not sure that it matters for some of the modals that were smaller before. I spent some time on it yesterday and couldn't get it to work the way I wanted it to. Do you wanna have a stab? Or provide some additional guidance?

gdpelican:
I was able to size particular modals by adding an `%md_dialog` element to the root of the template.

So for the discussion form, I could do

```
%md_dialog.discussion-form
  %form{...}
```
and in the css
```
.discussion-form {
  width: 60%; // or whatever's right for that modal)
}
```

Does that help?

HSalmon:
Ok, @gdpelican, I just fixed the vote-form and removed the blanket md-dialog width:

fix-vote-form-width

gdpelican:
Cool, looks good to me.

If you feel emboldened you could do that same for the discussion form (which tweaks out when you go from 'edit' to 'preview') in another PR, or I'll grab it sometime next week.

PR #3907, opened 2016-11-29
Inbox-style sidebar update

mhjb:
image
image


gdpelican:
Whoa, fantastic work, kudos!

I'm on my phone at the moment and can't give a proper QA, but will do so as soon as I get the chance. <3

mhjb:
Thanks :)

gdpelican:
Alright, I've had a go. I think it's really great.

My primary concern is around the background of the sidebar changing, for two reasons that I could see:

- The new background color makes the default group logo all but invisible
screen shot 2016-11-29 at 1 31 23 am
- Inbox doesn't overlap the content the way ours does when the sidebar is open; I think this feels like a bit of an odd state:
screen shot 2016-11-29 at 1 42 37 am

Inbox by comparison shoves all that stuff over to the side on mobile and waits for you to close the sidebar:
screen shot 2016-11-29 at 1 43 13 am


Shall we push it up to clone for a proper QA, and get some more eyes on it?

gdpelican:
I pushed this up to loomioclone, if you want to raise a QA thread about it @mhjb

mhjb:
Thanks very; will do

mhjb:
@gdpelican any feelings about swapping out the current default icon with …

image

image

?

mhjb:
Re. Inbox pushing the main content over: it's interesting. Feels a bit of a compromise either way, in that on a narrow screen Inbox pushes the main content past the edge of the window…

image

mhjb:
thar you go @gdpelican

mhjb:
This is ready to merge I think @gdpelican

mhjb:
I'd love to squash it, but I can't remember how and can't find the old tech manual where the process was laid out

angular/core/components/sidebar/sidebar.scss
```diff
@@ -17,42 +28,58 @@
...
+  font-size: 22px;
+  color: #666;
+}
+
+.fa-envelope-o {
+  font-size: 20px;
+}
+
+.fa-inbox {
+  color: #5cb85c;
```
gdpelican: If you're up for it, I'd suggest defining these as color variables in `0_mixins.scss` (I'm going to merge anyway tho :P)

Also, whenever I see a new 'grey' color, I wonder if we really need it, since we already have ~10 flavours of grey defined in that file.

gdpelican:
No more need for squashing; github's taken care of that for us now:
screen shot 2016-12-07 at 3 57 50 am

gdpelican:
Great work @mhjb !

mhjb:
You're great!

mhjb:
And awake at 4am?!

angular/core/components/sidebar/sidebar.scss
mhjb: roger wilco; the two colours are already define actually; I'll go hunting for them

PR #3920, opened 2016-12-06
Prevent users from editing context when they mean to comment

HSalmon:
[Trello card](https://trello.com/c/CZKcHUX6/2209-users-editing-title-and-context-instead-of-posting-comment-in-intro-thread)

screen shot 2016-12-07 at 11 15 57 am

### DIFF
--- a/angular/core/components/thread_page/context_panel/context_panel.haml
+++ b/angular/core/components/thread_page/context_panel/context_panel.haml
@@ -52,5 +52,5 @@
     .context-panel__attachment{ng-repeat: "attachment in discussion.attachments() | orderBy:'createdAt' track by attachment.id"}
       %attachment_preview{attachment: "attachment", mode: "context"}
 
-  %button.context-panel__add-comment.lmo-no-print{ng-click: "scrollToCommentForm()", translate: "thread_context.add_comment"}
+  %button.context-panel__add-comment.lmo-no-print{ng-click: "scrollToCommentForm()", translate: "thread_context.add_comment", ng-if: "canAddComment()"}
   %button.context-panel__edit.lmo-no-print{ng-click: "editThread()", translate: "thread_context.edit_thread", ng-if: "canEditThread()"}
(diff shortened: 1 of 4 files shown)


angular/core/components/thread_page/context_panel/context_panel.haml
```diff
@@ -52,4 +52,5 @@
     .context-panel__attachment{ng-repeat: "attachment in discussion.attachments() | orderBy:'createdAt' track by attachment.id"}
       %attachment_preview{attachment: "attachment", mode: "context"}
 
-  %a.context-panel__edit-link.lmo-card-minor-action.lmo-no-print{href: "", ng-click: "editThread()", translate: "thread_context.edit_thread", ng-if: "canEditThread()"}
+  %button.context-panel__add-comment.lmo-no-print{ng-click: "scrollToCommentForm()", translate: "thread_context.add_comment"}
```
gdpelican: This needs to appear only if you can comment, yeah? (a la `/development/view_open_group_as_visitor`)

angular/core/components/thread_page/context_panel/context_panel.scss
```diff
@@ -33,6 +33,20 @@
...
+  @include lmoBtnLink;
+  @include fontHelveticaMedium;
+  @include fontSmall;
   cursor: pointer;
+  text-transform: uppercase;
+  letter-spacing: .3px;
+  margin-right: 30px;
+}
+.context-panel__add-comment {
+  color: #5cb85c;
```
gdpelican: Let's pull this out into a color mixin. Also, are we certain we need a new color for this?

HSalmon: It's not a new colour, it's the colour from the start proposal button etc, but I couldn't find a variable for it in the app

gdpelican: Ooooh. Probably a bootstrap color variable (since we still use bootstrap, heh.)

Wonder if there's a variable defined in there for it...

mhjb:
design has this in Roboto — worth sticking to I reckon; feels more action-y from analogy with other Material apps

image

mhjb:
too little padding when there's an attachment…

image

mhjb:
sorry for the verbosity — one more thing: the [Material spec](https://material.google.com/components/cards.html#cards-content-blocks) has 8px padding around actions + 16px margin above

PR #3922, opened 2016-12-06
Close sidebar when user selects print option

HSalmon:
Address: https://github.com/loomio/loomio/issues/3897

close-sidebar-on-print


gdpelican:
Seems like a decent solution. Thanks @HSalmon

PR #3929, opened 2016-12-08
Add functionality for reminding undecided members

HSalmon:
[Trello card](https://trello.com/c/lrEGaiBg/1867-mention-undecided-members)

remind undecided

### DIFF
--- a/angular/core/components/thread_page/comment_form/comment_form.coffee
+++ b/angular/core/components/thread_page/comment_form/comment_form.coffee
@@ -5,9 +5,10 @@ angular.module('loomioApp').directive 'commentForm', ->
   replace: true
   controller: ($scope, $rootScope, FormService, Records, Session, KeyEventService, AbilityService, MentionService, AttachmentService, ScrollService, EmojiService, ModalService, SignInForm) ->
 
-    $scope.$on 'remindUndecided', (event, undecided) ->
-      undecided = _.flatten _.values(undecided)
-      undecided = _.map undecided, (member) -> "@" + "#{member}"
+    $scope.$on 'remindUndecided', (event) ->
+      return unless $scope.discussion.activeProposal()
+      ScrollService.scrollTo('.comment-form__comment-field')
+      undecided = _.map $scope.discussion.activeProposal().undecidedMembers(), (member) -> "@#{member.username}"
       $scope.comment.body = undecided.join(', ')
 
     $scope.showCommentForm = ->
--- a/angular/core/components/thread_page/comment_form/comment_form.coffee
+++ b/angular/core/components/thread_page/comment_form/comment_form.coffee
@@ -5,6 +5,11 @@ angular.module('loomioApp').directive 'commentForm', ->
   replace: true
   controller: ($scope, $rootScope, FormService, Records, Session, KeyEventService, AbilityService, MentionService, AttachmentService, ScrollService, EmojiService, ModalService, SignInForm) ->
 
+    $scope.$on 'remindUndecided', (event, undecided) ->
+      undecided = _.flatten _.values(undecided)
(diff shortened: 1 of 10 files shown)


angular/core/components/thread_page/undecided_panel/undecided_panel.coffee
```diff
@@ -15,3 +15,7 @@ angular.module('loomioApp').directive 'undecidedPanel', ->
 
     $scope.hideUndecided = ->
       $scope.undecidedPanelOpen = false
+
+    $scope.remindUndecided = ->
+      ScrollService.scrollTo('.comment-form__comment-field')
```
gdpelican: I'd suggest the scroll should happen inside the comment_from `$on`, since we'll want to be scrolling there anytime we manually manipulate the comment text

angular/core/components/thread_page/comment_form/comment_form.coffee
```diff
@@ -5,6 +5,11 @@ angular.module('loomioApp').directive 'commentForm', ->
   replace: true
   controller: ($scope, $rootScope, FormService, Records, Session, KeyEventService, AbilityService, MentionService, AttachmentService, ScrollService, EmojiService, ModalService, SignInForm) ->
 
+    $scope.$on 'remindUndecided', (event, undecided) ->
+      undecided = _.flatten _.values(undecided)
+      undecided = _.map undecided, (member) -> "@" + "#{member}"
```
gdpelican: What are your thoughts on, instead of broadcasting all those undecided members, doing something like

```
return unless $scope.discussion.activeProposal()
_.map $scope.discussion.activeProposal().undecidedMembers(), (member) -> "@#{member.username}"
```

config/locales/client.en.yml
```diff
@@ -326,7 +326,9 @@ en:
 
   undecided_panel:
     aria_label: 'Undecided members'
-    show_undecided: Show undecided members
+    show: Show
+    remind: Remind
```
gdpelican: Hrm. I'd suggest that 'Remind' on its own might not make it very clear what this does. Maybe 'Send Reminder'? Or maybe a button like this with a title?

Here's some thoughts:
screen shot 2016-12-08 at 4 21 29 am
screen shot 2016-12-08 at 4 29 37 am
screen shot 2016-12-08 at 4 32 24 am

@mhjb Do you have thoughts on this?

gdpelican:
I'd also suggest that this wants a speedbump of some kind somewhere, since there are going to be lots  of times where a button to @mention 200 people is not the right thing, and is in fact actively dangerous to have around. Slack has that message where it says 'Are you _sure_ you want to mention 400 people in 12 time zones?', so maybe we could display a warning or change the way it looks when you're planning on notifying a large number of people with this feature.

config/locales/client.en.yml
mhjb: Show & Remind feel like different order things. I'ma ponderin'. Leaning towards sticking it in the Options menu. Super-user feature. Could be fun to advertise it to coordinators via Intercom in-app message

HSalmon:
@gdpelican, I've implemented your suggested code improvements but not the suggested aesthetic changes. 

@robguthrie, @mhjb and I talked about this today. I consulted Rob about the lack of convergent thinking about the design of this particular feature and his thinking was that with a feature this small we can afford to deploy and iterate afterwards. It might provide us with more useful feedback to fuel the next iteration than we would otherwise have access to, right now it seems kinda speculative.

Rob, Matt feel free to chime in if you have things to add :)

HSalmon:
@gdpelican, @robguthrie would you mind engaging with the previous comment? Thanks

robguthrie:
I agree with what you've said.

Happy for you to own the job and make iterations based on user feedback.

PR #3930, opened 2016-12-08
Hide cancel button for screen readers, fix vote form aria-label

HSalmon:
[Trello card](https://trello.com/c/QxpppjBB/2236-ensure-cancel-is-not-first-tab-option-when-modal-launches)

### DIFF
--- a/angular/core/components/archive_group_form/archive_group_form.haml
+++ b/angular/core/components/archive_group_form/archive_group_form.haml
@@ -1,7 +1,7 @@
 %form{ng-submit: "submit()"}
   .lmo-disabled-form{ng-show: "isDisabled"}
   .modal-header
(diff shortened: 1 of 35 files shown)


gdpelican:
PRs like this make me so sad we never figured out how to do a modal template.

In any case, thanks for going through and doing this, @HSalmon

gdpelican:
Out of curiousity, would it work to simply put one aria-hidden: true on the %button in the %modal_cancel_button directive?

```
%button{type: "button", class: "close", ng-click: "$close()", aria-hidden: "true"}
```

HSalmon:
I tried that but i found I could still tab to it and voiceover didn't read
anything out - it was like a blank element.

Issue #3937, opened 2016-12-08
Can't @mention people while editing comment


mhjb:
![image](https://cloud.githubusercontent.com/assets/1380991/21029580/4d329586-bdff-11e6-9648-0fe9cf0d00da.png)

HSalmon:
Fixed pending deploy

PR #3940, opened 2016-12-09
Enable mentions for edit comment form

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3937


gdpelican:
Gah so easy.

Issue #3914, opened 2016-12-01
Long string doesn't wrap on previous proposals page


rdbartlett:
![image](https://cloud.githubusercontent.com/assets/970124/20777931/49bfdb42-b7cf-11e6-9bb9-b5fa8bfe1e67.png)

HSalmon:
Fixed pending deploy

PR #3941, opened 2016-12-09
Hide overflow on previous proposal body

HSalmon:
Addresses: https://github.com/loomio/loomio/issues/3914

Not sure that this is _the_ solution but it is _a_ solution...


gdpelican:
Mind posting up a screenshot, @HSalmon ?

HSalmon:
@gdpelican, is this sufficient?

screen shot 2016-12-16 at 8 41 12 am

PR #3951, opened 2017-02-01
move to 2.4.0

robguthrie:


fauno:
does it work? i upgraded to 2.4.0 and got lots of deprecation warnings :P

robguthrie:
no. looks like rails 4 won't work with 2.4. We'll need to move to rails 5 first.

PR #3986, opened 2017-03-21
Reload Poll colums to get the group_id attribute

fauno:
Had to do this to prevent migration from failing by missing group_id method :)


gdpelican:
Thanks!

PR #3998, opened 2017-03-26
Dot vote stance chart

robguthrie:
Displays each dot-vote-stance as its own little chart.


robguthrie:
image

angular/core/components/poll/common/bar_chart_panel/poll_common_bar_chart_panel.haml
```diff
@@ -1,5 +1,3 @@
 .poll-common-bar-chart-panel
   %h3.lmo-card-subheading{translate: "poll_common.results"}
-  .poll-common-bar-chart-panel__chart-container
-    .poll-common-bar-chart-panel__bar{ng-repeat: "option in poll.pollOptions() track by option.id", ng-style: "{'background-image': backgroundImageFor(option), 'background-size': percentageFor(option)}"}
-      {{barTextFor(option)}}
+  %poll-common-bar-chart{poll: "poll", size: "normal"}
```
gdpelican: We tend to use underscores for directives, so `%poll_common_bar_chart`

angular/core/components/poll/dot_vote/votes_panel_stance/poll_dot_vote_votes_panel_stance.coffee
```diff
@@ -0,0 +1,21 @@
+angular.module('loomioApp').directive 'pollDotVoteVotesPanelStance', (PollService, RecordLoader) ->
+  scope: {stance: '=', poll: '='}
```
gdpelican: It looks like poll isn't used in here? Also if we did need I reckon `$scope.stance.poll()` would be the way to do it.

angular/core/components/poll/common/votes_panel/poll_common_votes_panel.haml
```diff
@@ -5,6 +5,6 @@
       %md-option{ng-repeat: "opt in sortOptions", ng-value: "opt", translate: "poll_common_votes_panel.{{opt}}"}
   .poll-common-votes-panel__no-votes{ng-if: "!hasSomeVotes()", translate: "poll_common_votes_panel.no_votes_yet"}
   .poll-common-votes-panel__has-votes{ng-if: "hasSomeVotes()"}
-    %poll_common_directive{stance: "stance", name: "votes_panel_stance", ng-repeat: "stance in stances() track by stance.id"}
+    %poll_common_directive{stance: "stance", poll: "poll", name: "votes_panel_stance", ng-repeat: "stance in stances() track by stance.id"}
```
gdpelican: Do we need to pass poll in here?

app/helpers/email_helper.rb
```diff
@@ -90,4 +90,13 @@ def percentage_for(poll, index)
       (100 * poll.stance_counts[index].to_f / poll.stance_counts.max).to_i
     end
   end
+
+  def stance_choice_percentage_for(stance, stance_choice)
+    scores = stance.stance_choices.pluck(:score)
```
gdpelican: I think `max` here might be slow. What about
```
max_score = stance.stance_choices.maximum(:score).to_i
if max_score.to_i <= 0
  0
else
  (100 * stance_choice.score.to_f / max_score).to_i
end
```

?

robguthrie: Why do you think it might be slow?

robguthrie: maximum is what I'm looking for though.

gdpelican: Just that I'm pretty sure doing `MAX(column)` in postgres is way faster than doing `Array.max` in ruby

PR #4005, opened 2017-03-31
Generalize caches

gdpelican:


robguthrie:
Wow. that's beautiful. Nice work.

PR #4008, opened 2017-04-03
Account for polls on dashboard and group page

gdpelican:
Allows polls to show up in the correct order when sorting the dashboard / group page

NB that this includes the Caches work from #4005


robguthrie:
Wow. Looks great

PR #4012, opened 2017-04-06
Meeting time zone

gdpelican:
Adds option to select time zone when creating and displaying meeting poll


gdpelican:
This is ready for a look @robguthrie 

I've gone with the 'local time' approach, so we detect your timezone via JS and use that as the default when viewing or creating polls. You can then change from there if you'd like, to a (short; around 35) list of pre-defined options.

I've also put this time zone stuff into the emails, so it falls back to the poll's defined time zone if the user hasn't defined one.

One bug: It's currently parsing date-only poll options into full times now, need to have a look at that in a bit.

gdpelican:
Also, moment now has a `guess()` function, which I think means we may not need jstz for this.

Issue #4022, opened 2017-04-12
loomio.org email marked as spam


fauno:
```
X-Spamd-Result: default: False [10.60 / 15.00]
 RAMBLER_URIBL(4.50)[stanford.edu.uribl.rambler.ru]
 PHISHING(4.00)[attending.io->loomio.org, stanford.edu->loomio.org]
 R_BAD_CTE_7BIT(3.00)[]
 R_DKIM_ALLOW(-0.20)[loomio.org]
 BAYES_HAM(-0.40)[10.17%]
 DMARC_NA(0.00)[loomio.org]
 URIBL_BLOCKED(0.00)[imgur.com.multi.uribl.com, stanford.edu.multi.uribl.com, attending.io.multi.uribl.com, loomio.org.multi.uribl.com]
 MIME_GOOD(-0.10)[multipart/alternative, text/plain]
 R_SPF_ALLOW(-0.20)[+a:smtp.loomio.org]
```

numbers in parenthesis are scores, positive ones lead to email being marked as spam.

* the phishing score was given because non-loomio urls are printed as they are but linked through a loomio's ahoy address (i'm guessing is for tracking clicks), which is a phishing technique.

* the bad content transfer encoding is set at 7 bit and i'm guessing the contents are not. here's a discussion about cte: https://stackoverflow.com/a/28531705

* the rambler uribl score depends on the links printed, but the previous two scores are enough (7 points!) for rspamd to greylist and then mark them as spam.

i guess if you can fix the encoding you could get away with the phishy urls? :P

gdpelican:
Have you still been having troubles with this, @fauno ? We've switched over to Amazon SES on our end, which I believe has helped with our spamminess.

fauno:
a few still got classified as spam, but way less than before :)

thunderbird also tells me email notifications can be scam, probably
because of ahoy's hidden urls

-- 
}(:=

gdpelican:
Cool, I'm going to close this for now, we'll see how we go going forward.

We're discovering more and more that as much as we say that everyone's on slack and facebook... really EVERYONE is on email, so it's really important we keep on top of this stuff. Thanks for raising.

PR #4023, opened 2017-04-13
Use multipart/alternative for email encoding

gdpelican:
I'm not sure why we're using 7bit currently; maybe ahoy email or premailer or something does this?
https://rails-bestpractices.com/posts/2010/08/05/use-multipart-alternative-as-content_type-of-email/

### DIFF
--- a/app/mailers/base_mailer.rb
+++ b/app/mailers/base_mailer.rb
@@ -7,7 +7,8 @@ class BaseMailer < ActionMailer::Base
   add_template_helper(PrettyUrlHelper)
 
   NOTIFICATIONS_EMAIL_ADDRESS = "notifications@#{ENV['SMTP_DOMAIN']}"
-  default :from => "Loomio <#{NOTIFICATIONS_EMAIL_ADDRESS}>"
+  default from: "Loomio <#{NOTIFICATIONS_EMAIL_ADDRESS}>",
+          'Content-Transfer-Encoding' => 'multipart/alternative'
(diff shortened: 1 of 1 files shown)


gdpelican:
@robguthrie How do you feel about this change? Ref #4022

robguthrie:
I feel fine about the idea, unsure if it will break stuff in practice. 
Give it a go?

fauno:
Content-Type is different from Content-Transfer-Encoding, and every part can have a different CTE.  I just checked the parts and the header and text/plain part are CTE 7bit and the text/html part is quoted-printable... while github notifications are all 7bit.

what happens if i use unicode? can you check this notification? áéíóú, 🙃

fauno:
by looking at the function that throws the error, i'm guessing it's because of the mixed CTEs [7bit,7bit,quoted-printable]: https://github.com/vstakhov/rspamd/blob/71d52e806a4780d95eb0322544330e31a7a375d9/src/libmime/mime_expressions.c#L1277

i also found a github notification where the main cte was 7bit but being written in correct spanish (with tildes and all) the parts were quoted-printable and it was marked as bad_cte_7bit by rspam, so it may be too picky.

fauno:
asked on rspamd :)

https://github.com/vstakhov/rspamd/issues/1590

gdpelican:
Okay, we're not going to go with this for now. Hopefully with the move to SES we won't have these deliverability issues going forward.

PR #4065, opened 2017-05-10
Add new decision to sidebar, decisions to group page

robguthrie:


robguthrie:
WIP

gdpelican:
I think this creates a couple more problems than it solves at the moment, so will hold off in lieu of 'add poll to discussion'

PR #4069, opened 2017-05-10
support translation api key not present

robguthrie:
proposed fix for https://www.loomio.org/d/nceoSVZ1/error-on-loomio-docker-container-at-startup


gdpelican:
Oh. Yep!

PR #4086, opened 2017-05-16
Live update on poll page

gdpelican:


robguthrie:
@gdpelican I wonder if it's worth creating a poll channel, so polls that are not part of a group still live update.

gdpelican:
That was the idea behind this PR; should be working in prod now

robguthrie:
duh.. I totally missed those lines. nice work

PR #4091, opened 2017-05-17
Remove unused code

gdpelican:


gdpelican:
This removes cohorts, themes, categories, and group requests from the admin panel. None have been touched since 2015. I figure if we _really_ need to look at these things, we can do so in metabase. Any objections @robguthrie ?

robguthrie:
none at all. thanks

PR #4093, opened 2017-05-18
Tidy share form, move decisions card to top of group page

robguthrie:


angular/core/components/poll/common/share/link_form/poll_common_share_link_form.haml
```diff
@@ -1,17 +1,10 @@
 %section
```
gdpelican: I'd like to get to a point where we don't have a whole bunch of directives for different poll displays, and instead have a single bit of code which takes a poll search query somehow and displays the polls it finds

```
%poll_collection_card{model: "discussion", query: "closedPolls"}
%poll_collection_card{model: "group", query: "polls"}
```
etc.

angular/test/protractor/group_spec.coffee
```diff
@@ -100,7 +100,7 @@ describe 'Group Page', ->
 
       xit 'open group displays previous proposals', ->
         page.loadPath('view_open_group_as_visitor')
-        page.expectText('.group-previous-polls-card', 'Let\'s go to the moon!')
+        page.expectText('.group-decisions-card', 'Let\'s go to the moon!')
```
gdpelican: Wonder if we can uncomment this test? :)

angular/core/components/poll/common/share/link_form/poll_common_share_link_form.haml
robguthrie: yea, my thoughts too. I added a nice to have ticket

PR #4096, opened 2017-05-20
change contact form to use intercom api

robguthrie:
should probably pull this out in to the marketing plugin


Gemfile.lock
```diff
@@ -593,6 +594,10 @@ GEM
     request_store (1.1.0)
     responders (2.2.0)
       railties (>= 4.2.0, < 5.1)
+    rest-client (2.0.2)
```
gdpelican: Why not use HTTParty, since we have it already?

app/controllers/contact_messages_controller.rb
```diff
@@ -13,9 +13,8 @@ def create
   	@contact_message = ContactMessage.new(permitted_params.contact_message)
     @contact_message.user = current_user if current_user.is_logged_in?
   	if @contact_message.save
-      ContactMessageMailer.delay(priority: 2).contact_message_email(@contact_message)
+      ContactMessageSerivce.deliver(email: @contact_message.email, message: @contact_message.message )
```
gdpelican: Oops typo.

app/services/contact_message_service.rb
```diff
@@ -1,8 +1,10 @@
 class ContactMessageService
+  def self.deliver(email:, message:)
+    data = { from: { type: "user", email: email }, body: message }
+    headers = {'Authorization' => "Bearer #{ENV['INTERCOM_ACCESS_TOKEN']}",
```
gdpelican:
```
return unless ENV['INTERCOM_ACCESS_TOKEN']
```

gdpelican: We should probably remove `app/models/contact_message.rb` as well. As a side note, this means we won't be storing all of our contact messages in our database, which is probably a good thing anyway, right?

app/services/contact_message_service.rb
```diff
@@ -1,8 +1,10 @@
...
+               content_type: :json,
+               accept: :json}
 
-  def self.create(contact_message:, actor:)
-    contact_message.user = actor
-    ContactMessageMailer.delay(priority: 2).contact_message_email contact_message if contact_message.save
+    RestClient.post("https://api.intercom.io/messages", data.to_json, headers)
```
gdpelican: I'll hold off on my desire to transform this into `Clients::Intercom`, since this is clearly a bandaid until we can get unhooked from Intercom.

I'll note though that the express purpose of a `Clients::Base` subclass is 'a model which sends http requests to a specific api', where we define a base api (`api.intercom.io`), a path (`messages`), some data / headers to send, and some behaviour to do when it responds.

Gemfile.lock
robguthrie: I couldn't get HTTParty to work.. it was after 3 hours that I thew it out. :(

app/services/contact_message_service.rb
robguthrie: oh wow.. I did not know about Clients::Base. Sounds good.

Gemfile.lock
robguthrie: on the plus side, we remove our intercom gem dependancy

app/services/contact_message_service.rb
robguthrie: yea, next pass this will happen

PR #4108, opened 2017-05-24
improve poll full serializer

robguthrie:


gdpelican:
Merged into the poll common card work

PR #4124, opened 2017-05-29
Don't serialize motions with discussions

gdpelican:


gdpelican:
Will revisit this with more robust approach to removing motions soon

PR #4252, opened 2017-08-13
full login specs, token expiry to 30 mins, find verified on login

robguthrie:


app/models/login_token.rb
```diff
@@ -1,7 +1,7 @@
 class LoginToken < ActiveRecord::Base
   belongs_to :user, required: true
   has_secure_token :token
-  EXPIRATION = ENV.fetch('LOGIN_TOKEN_EXPIRATION_MINUTES', 15)
+  EXPIRATION = ENV.fetch('LOGIN_TOKEN_EXPIRATION_MINUTES', 30)
```
gdpelican: Can we also replace the line in the login token email so it says 'This token will expire in 30 minutes' instead of 'this token will expire at a time that's never right?'

robguthrie: yes

PR #4272, opened 2017-08-24
Migrate cv to reaction

gdpelican:


gdpelican:
@robguthrie This is good to go but need to think about how we do the migration properly. Think you could run it on your prod DB to see how long it takes currently?

PR #4286, opened 2017-09-02
Usage reporting

robguthrie:


app/controllers/api/usage_reports_controller.rb
```diff
@@ -0,0 +1,11 @@
+class API::UsageReportsController < ActionController::Base
+  def create
+    UsageReport.create!(params[:usage_report]) if ENV['HOSTED_BY_LOOMIO']
```
gdpelican: This line makes me think that this wants to go into the loomio org plugin.

But actually there's a helper method in ApplicationController called `hosted_by_loomio?` which gets overridden (from 'false' to 'true' by the org plugin, so maybe that's the way to go? Just want to minimize the ways that we check for loomio.org.

```
UsageReport.create(params) if hosted_by_loomio?
```

app/extras/clients/loomio.rb
```diff
@@ -0,0 +1,15 @@
+class Clients::Loomio < Clients::Base
+  def send_usage_report!(usage_report)
+    post "usage_reports", params: { usage_report: usage_report }
```
gdpelican: Yay, great stuff!

db/migrate/20170901005004_usage_reports.rb
```diff
@@ -0,0 +1,17 @@
...
+    create_table :usage_reports do |t|
+      t.integer :groups_count
+      t.integer :users_count
+      t.integer :discussions_count
+      t.integer :polls_count
+      t.integer :comments_count
+      t.integer :stances_count
+      t.integer :reactions_count
+      t.integer :visits_count
+      t.string  :canonical_host
```
gdpelican: I think we're likely going to want to index on canonical_host (so that we can search by host name), or alternatively assign a site_id integer to each unique canonical host for easy lookup / graphs over time etc.

app/services/usage_report_service.rb
```diff
@@ -0,0 +1,15 @@
+class UsageReportService
+  def self.send
+    return if ENV['DISABLE_USAGE_REPORTING']
+    Clients::Loomio.new.send_usage_report!({
+      groups_count: Group.count,
```
gdpelican: Seems like we're missing a field or two here, support_email? reactions_count? Would be good to have the schema match the report that's sent over to start

app/controllers/api/usage_reports_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+    head :ok
+  end
+
+  private
+  def params
+    request.parameters
```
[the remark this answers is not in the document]
robguthrie: We need this to bypass strong_parameters, which I don't want to have to define for this.

PR #4288, opened 2017-09-05
Reactions

gdpelican:


robguthrie:
I feel much better knowing the reactions will meet basic a11y requirements now.

There are a few more things to test out here.. And I think I have energy to make a non english based way to find emoji (classic category tabs and scrolling). But if we make everything else production ready, I won't hold up the train.

gdpelican:
@robguthrie Could you push up your changes for this branch plz?

PR #4291, opened 2017-09-05
display title of poll being voted upon

robguthrie:


gdpelican:
Ack sorry I just sniped this one. I did just the poll type, but happy for this one to go in instead.

robguthrie:
Cool. Yea I think title is critical to improving this.

PR #4300, opened 2017-09-13
make app themeable

robguthrie:


angular/core/components/navbar/navbar.scss
```diff
@@ -1,4 +1,22 @@
 .lmo-navbar {
+  .navbar__logo-bg {
+    background-repeat: no-repeat;
+    background-position: center;
+    background-size: auto 20px;
+  }
+
+  // @media only screen and (min-width: 768px) {
```
gdpelican: I reckon just lose these; we won't miss em.

angular/core/components/poll_page/poll_page_controller.coffee
```diff
@@ -1,13 +1,12 @@
 angular.module('loomioApp').controller 'PollPageController', ($scope, $rootScope, $routeParams, MessageChannelService, Records, $location, ModalService, PollService, PollCommonOutcomeModal, PollCommonEditVoteModal, PollCommonShareModal, Session) ->
-  $rootScope.$broadcast('currentComponent', { page: 'pollPage', skipScroll: true })
```
gdpelican: I can't help but feel there's some reason for this being pre-model load. Could you ensure that this still works properly for, say, loading a private discussion while logged out?

angular/core/css/1_base.scss
```diff
@@ -28,3 +28,7 @@ a {
   color: $link-color;
   text-decoration: none;
 }
+
+*:focus {
+    outline: none;
```
gdpelican: I'm seeing some a11y complaints about this one:

http://www.outlinenone.com/
http://a11yproject.com/posts/never-remove-css-outlines/

angular/core/root_controller.coffee
```diff
@@ -26,7 +26,16 @@ angular.module('loomioApp').controller 'RootController', ($scope, $timeout, $loc
     IntercomService.boot()
     MessageChannelService.subscribe()
 
+  setTitle = (title) ->
+    document.querySelector('title').text = _.trunc(title, 300) + ' | Loomio'
+    Session.pageTitle = title
+
   $scope.$on 'currentComponent', (event, options = {}) ->
+    if options.titleKey
```
gdpelican: a possible alternative:
```
setTitle(options.title or $translate.instant(options.titleKey))
```

angular/core/root_controller.coffee
```diff
@@ -26,7 +26,16 @@ angular.module('loomioApp').controller 'RootController', ($scope, $timeout, $loc
     IntercomService.boot()
     MessageChannelService.subscribe()
 
+  setTitle = (title) ->
+    document.querySelector('title').text = _.trunc(title, 300) + ' | Loomio'
```
gdpelican: I've just thought that a better way to do this would be
```
%title {{ pageTitle }}
```
in application/index.html.haml,

then 
```
$scope.pageTitle = "#{_.trunc(title, 300)} | Loomio"
```

rather than relying on querySelector

gdpelican: Also, there's a bit of discomfort around where our 'grab bag' goes at the moment... sometimes it's AppConfig, other times Session, and I'm not sure which is better (or which means which).

I'd be tempted here to do
```
$rootScope.$broadcast 'pageTitle', title
```
```
# navbar.coffee
$scope.$on 'pageTitle', (_, title) -> $scope.title = title
```

to avoid a global state if we don't need one

app/models/formal_group.rb
```diff
@@ -53,7 +53,7 @@ class FormalGroup < Group
   has_attached_file    :logo,
                        url: "/system/groups/:attachment/:id_partition/:style/:filename",
                        styles: { card: "67x67#", medium: "100x100#" },
-                       default_url: 'default-logo-:style.png'
+                       default_url: AppConfig.theme[:default_group_logo_src]
```
gdpelican: Hm, do we care that we're losing our default image sizing here?
(ie, normally you'd be able to get `default-logo-small`, `default-logo-medium`, etc.)

How big is the default group logo src image size?

public/stylesheets/error_page.css
```diff
@@ -14,7 +14,7 @@ h1 {
 }
 
 #header {
-    background-image: url("/assets/loomio-orange.png");
+    background-image: url("/theme/large_logo.png");
     background-repeat: no-repeat;
     height: 150px;
     text-indent: -999em;
```
gdpelican: I think the 'mascot.gif' is the wrong one to move (I don't believe we use that asset anywhere)... we want 'public/img/mascot.png' instead (which is used in the oauth apps and for the auth modal and slack pages). I'd also be happy to do this without customizing the mascot for now.

app/models/formal_group.rb
robguthrie: Na, we don't care. It's 826 bytes

public/stylesheets/error_page.css
robguthrie: yea, I backtracked on that and hoped you'd spot where I did not clean up properly

robguthrie: so well done

angular/core/components/poll_page/poll_page_controller.coffee
robguthrie: Ok, that test seems to work fine.

PR #4301, opened 2017-09-13
Read at improvements

gdpelican:
Would really love a CR on this, @robguthrie 

Why do this?
People complain about our 'last read' stuff all the time. It's annoying and hard to repro and makes us look bad. I'd like to simplify the logic around it, so that at least when we make tweaks it's easier to understand.

Changes in this branch:
- I've moved 'mark_as_read' to the events controller, where it feels more natural to me. I've also added a 'mark_as_seen' endpoint to discussions, which is called when you view a discussion, but haven't necessarily seen any of its content yet. These two actions were conflated in 'mark_as_read' before
- Call 'mark_as_read' when a thread item becomes visible, and 'mark_as_seen' when the thread page is loaded.
- Serialize out some minimal information about what's changed on the reader from the 'mark_as_read' request. This information is slotted into updating the existing discussion.


angular/core/components/thread_page/activity_card/activity_card.coffee
```diff
@@ -56,7 +56,7 @@ angular.module('loomioApp').directive 'activityCard', ->
 
     $scope.threadItemVisible = (item) ->
       addSequenceId(item.sequenceId)
-      $scope.discussion.markAsRead(item.sequenceId)
+      item.markAsRead()
```
gdpelican: This way we avoid the awkward process of taking an event, getting the discussion, and sending over its sequence id along with the discussion to the API.

angular/core/models/discussion_model.coffee
```diff
@@ -128,12 +128,10 @@ angular.module('loomioApp').factory 'DiscussionModel', (DraftableModel, AppConfi
...
 
-    markAsRead: (sequenceId) ->
-      sequenceId = @lastSequenceId if isNaN(sequenceId)
-
-      if @discussionReaderId? and (_.isNull(@lastReadAt) or @lastReadSequenceId < sequenceId)
-        @remote.patchMember @keyOrId(), 'mark_as_read', sequence_id: sequenceId
-        @update(lastReadAt: moment(), lastReadSequenceId: sequenceId)
+    markAsSeen: ->
+      return unless @discussionReaderId and !@lastReadAt
+      @remote.patchMember @keyOrId(), 'mark_as_seen'
```
gdpelican: We call mark_as_seen if you land on a thread page, you have a discussion reader, and your lastSeenAt is unpopulated. Mark_as_seen doesn't do much more than `update_attribute` on `last_read_at`

angular/core/models/event_model.coffee
```diff
@@ -34,5 +34,10 @@ angular.module('loomioApp').factory 'EventModel', (BaseModel) ->
     model: ->
       @recordStore[@constructor.eventTypeMap[@eventable.type]].find(@eventable.id)
 
+    markAsRead: ->
+      return unless @sequenceId > @discussion().lastReadSequenceId
+      @remote.postMember @id, 'mark_as_read'
```
gdpelican: Only call 'mark_as_read' if the event's sequence id is greater than the existing last read sequence id

app/controllers/api/events_controller.rb
```diff
@@ -28,6 +33,13 @@ def sequence_id_for_comment(collection)
   # we always want to serialize out events in the events controller
   alias :events_to_serialize :resources_to_serialize
 
+  def respond_with_reader(scope: default_scope, serializer: DiscussionReaderSerializer, root: :discussions)
+    render json: Array(DiscussionReader.for_model(resource.discussion, current_user)),
```
gdpelican: Here we serialize out JSON like this:
```
{ discussions: [
  id: #
  last_read_at: (time)
  last_read_sequence_id: # 
]}
```
which will update the discussion on the client with the new reader values. It's one of the only times we respond with a resource that _isn't_ the controller resource, but I think the tradeoff of having an endpoint that we just send an id to (rather than a discussion id and sequence id) is worth it.

app/models/events/new_discussion.rb
```diff
@@ -7,7 +7,12 @@ def self.publish!(discussion)
     create(kind: 'new_discussion',
            user: discussion.author,
            announcement: discussion.make_announcement,
-           eventable: discussion).tap { |e| EventBus.broadcast('new_discussion_event', e) }
+           eventable: discussion,
+           created_at: discussion.created_at).tap { |e| EventBus.broadcast('new_discussion_event', e) }
```
gdpelican: We need to ensure that the `created_at` values for the events match correctly, otherwise we might end up with off-by-millisecond errors when 'reading' content that you just created.

app/serializers/discussion_reader_serializer.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  attributes :id,
+             :discussion_reader_id,
+             :read_items_count,
+             :read_salient_items_count,
+             :last_read_sequence_id,
+             :last_read_at
+
+  def id
+    object.discussion_id
```
gdpelican: Here the 'id' is actually the discussion's id, to ensure that the clientside model is updated correctly.

config/initializers/event_bus.rb
```diff
@@ -63,23 +64,14 @@
...
-  config.listen('discussion_reader_viewed!',
-                'discussion_reader_dismissed!') do |discussion, actor|
-
-    reader_cache = Caches::DiscussionReader.new(user: actor, parents: Array(discussion))
-    collection = ActiveModel::ArraySerializer.new([discussion], each_serializer: MarkedAsRead::DiscussionSerializer, root: 'discussions', scope: { reader_cache: reader_cache } )
-
-    MessageChannelService.publish(collection, to: actor)
+  config.listen('discussion_reader_viewed!', 'discussion_reader_dismissed!') do |reader|
+    MessageChannelService.publish(
+      DiscussionReaderSerializer.new(reader, root: :discussions).as_json,
```
gdpelican: I, for the life of me, could only vaguely follow why we did the removed lines 76-82 that way. Here we're saying, 'broadcast the newly read values to the person who read them', which makes your other tabs / devices auto-update when you read something.

PR #4305, opened 2017-09-15
add site specific features

robguthrie:


robguthrie:
Hi @gdpelican can you see anything that needs to be fixed here?

app/models/concerns/null/user.rb
```diff
@@ -1,6 +1,9 @@
 module Null::User
   # include HasAvatar
   include Null::Object
+  def is_admin
```
gdpelican: This wants to be added to false_methods, rather than in its own method

app/controllers/api/registrations_controller.rb
```diff
@@ -3,7 +3,7 @@ class API::RegistrationsController < Devise::RegistrationsController
 
   def create
     build_resource(sign_up_params)
-    if resource.save
+    if can_create? && resource.save
```
gdpelican: We don't really want to render 422 if can_create? is false. I think a before_action is the way to go here.

```
actor.ability.authorize! :create, resource
```

Then `can_create?` goes into a :create, User ability block, which is the most natural place for it

angular/core/components/sidebar/sidebar.haml
```diff
@@ -42,11 +42,11 @@
         %md_button.sidebar__list-item-button--subgroup{href: "", lmo-href: "{{groupUrl(group)}}", ng-if: "!group.isParent()", ng-class: "{'sidebar__list-item--selected': onPage('groupPage', group.key)}"}
           {{group.name}}
         .sidebar__list-item-padding
-      %md_list_item
+      %md_list_item{ng-if: "canStartGroup()"}
         %md_button.sidebar__list-item-button.sidebar__list-item-button--start-group{href: "", ng-click: "startGroup()", aria-label: "{{ 'sidebar.start_group' | translate }}"}
           %i.sidebar__list-item-icon.material-icons group_add
           %span{translate: "sidebar.start_group"}
-      %md_list_item
+      %md_list_item{ng-if: "canPublicGroups()"}
```
gdpelican: `canViewPublicGroups` would follow the ability naming convention most closely

angular/core/components/sidebar/sidebar.coffee
```diff
@@ -63,6 +65,9 @@ angular.module('loomioApp').directive 'sidebar', ($rootScope, $mdMedia, $mdSiden
     $scope.currentUser = ->
       Session.user()
 
+    $scope.canStartGroup = -> AbilityService.canStartGroups()
+    $scope.canPublicGroups = -> AppConfig.features.public_groups
```
gdpelican: Might as well make this an AbilityService method too, `canViewPublicGroups`

angular/core/components/sidebar/sidebar.coffee
```diff
@@ -47,8 +47,10 @@ angular.module('loomioApp').directive 'sidebar', ($rootScope, $mdMedia, $mdSiden
       InboxService.unreadCount()
 
     $scope.showContactUs = ->
-      # TODO: use loomio_org plugin to determine official site or not
-      AppConfig.baseUrl == 'https://www.loomio.org/'
+      IntercomService.available()
```
gdpelican: I think this changes behaviour a little bit; we have the contact form as a fallback if people have disabled intercom for whatever reason (as a 3rd party script).

Anyway, we should always display contact us, and switch what happens when you click it based on whether Intercom is available or not.

app/extras/app_config.rb
```diff
@@ -30,4 +30,17 @@ def self.theme
...
+     {
+       create_user:   env_boolean('FEATURES_CREATE_USER', true),
+       create_group:  env_boolean('FEATURES_CREATE_GROUP', true),
+       public_groups: env_boolean('FEATURES_PUBLIC_GROUPS', true),
+       help_link:     env_boolean('FEATURES_HELP_LINK', true)
+     }
+   end
+
+   def self.env_boolean(key, default)
+     %w(1 true).include? ENV.fetch(key, default).to_s
```
gdpelican: For some reason %w(1 true) feels not the best to me; it's arbitrary and may or may not follow rules about ENV we've created already or will create.

```
# .env
FEATURES_CREATE_USER=1 # you can create users
# you can create groups because you didn't put anything
FEATURES_PUBLIC_GROUPS=yes # you can't see public groups
FEATURES_HELP_LINK=false # you can't see the help link
```

It also took me a bit to figure out that we're wanting to default to true here so we don't have to change existing installs at all. How would you feel about

```
def self.features
  {
    create_user: !ENV['DISABLE_FEATURE_CREATE_USER'],
    create_group: !ENV['DISABLE_FEATURE_CREATE_GROUP']
  }
end
```

app/controllers/api/registrations_controller.rb
robguthrie: knew you'd have the answer for that. Thanks

angular/core/components/sidebar/sidebar.coffee
robguthrie: Yea.. it's because I noticed we don't support contact by email right now. I'd prefer that too.

PR #4311, opened 2017-09-22
support changing palette with ENV

robguthrie:


app/extras/app_config.rb
```diff
@@ -27,7 +28,9 @@ def self.theme
       primary_palette:                   ENV.fetch('THEME_PRIMARY_PALETTE', 'orange'),
       accent_palette:                    ENV.fetch('THEME_ACCENT_PALETTE', 'cyan'),
       primary_palette_config: JSON.parse(ENV.fetch('THEME_PRIMARY_PALETTE_CONFIG', '{"default": "400"}')),
-      accent_palette_config:  JSON.parse(ENV.fetch('THEME_ACCENT_PALETTE_CONFIG', '{"default": "500"}'))
+      accent_palette_config:  JSON.parse(ENV.fetch('THEME_ACCENT_PALETTE_CONFIG', '{"default": "500"}')),
+      custom_primary_palette: json_parse_or_false('THEME_CUSTOM_PRIMARY_PALETTE'),
```
gdpelican: What's the difference between having a custom primary palette, and defining the palette you want in THEME_PRIMARY_PALETTE_CONFIG? Not sure I see the need for these new ENVs yet

angular/core/config/material.coffee
```diff
@@ -1,6 +1,16 @@
 angular.module('loomioApp').config ($mdThemingProvider, $mdAriaProvider) ->
   theme = window.Loomio.theme
-  $mdThemingProvider.theme('default')
-    .primaryPalette(theme.primary_palette, theme.primary_palette_config)
-    .accentPalette(theme.accent_palette, theme.accent_palette_config);
+
+  if theme.custom_primary_palette
```
gdpelican: I'm pretty skeptical that we need this much logic here. Can we get away with
```
$mdThemingProvider.theme('default')
  .primaryPalette(theme.primary_palette, theme.primary_palette_config)
  .accentPalette(them.accent_palette, theme.accent_palette_config)
```

and ensuring that those theme values are correct based on ENV on the ruby side?

app/extras/app_config.rb
robguthrie: The client does not want any of the material palettes, so I have to generate one (two actually) and pop it in as a new palette.

angular/core/config/material.coffee
robguthrie: yea, we can do that.

PR #4314, opened 2017-09-26
migrate to materialdesignicons.com icons

robguthrie:


robguthrie:
@gdpelican this is ready to go.

PR #4322, opened 2017-10-02
use angular 1.6

robguthrie:
we did not do this earlier because angular-bootstrap held us back


gdpelican:
Sweet. What does angular 1.6 get us? Performance increases?

robguthrie:
they say it has some compile perf fixes and otherwise are pretty unclear about what the perf improvements actually are.. I'm just crossing fingers and hoping it's better.

PR #4325, opened 2017-10-03
[wip] fetch new events on reconnect

robguthrie:
I've tested the fetchAllNew manually and it works well. Need to test the faye stuff and remove the sinful code

### DIFF
--- a/angular/core/services/connection_service.coffee
+++ b/angular/core/services/connection_service.coffee
@@ -1,0 +1,8 @@
+angular.module('loomioApp').factory 'ConnectionService', ->
+  new class ConnectionService
+    onReconnect: (fn) ->
+      PrivatePub.faye (client) ->
+        client.on 'transport:up', fn
+    onDisconnect: (fn) ->
+      PrivatePub.faye (client) ->
+        client.on 'transport:down', fn
--- a/app/controllers/api/events_controller.rb
+++ b/app/controllers/api/events_controller.rb
@@ -9,13 +9,24 @@ class API::EventsController < API::RestfulController
   private
 
   def accessible_records
-    load_and_authorize(:discussion).items.sequenced
+    if params[:event_id_gt]
+      Event.joins(:events_groups).
+        where("events_groups.group_id in (?)", current_user.group_ids).
+        where("events.id > ?", params[:event_id_gt]).order(:id)
+    else
+      load_and_authorize(:discussion).items.sequenced
+    end
   end
 
   def page_collection(collection)
-    collection.where('sequence_id >= ?', sequence_id_for(collection))
-              .includes(:eventable)
-              .limit(params[:per] || default_page_size)
+    if params[:event_id_gt]
+      collection.includes(:eventable)
+                .limit(params[:per] || default_page_size)
+    else
+      collection.where('sequence_id >= ?', sequence_id_for(collection))
+                .includes(:eventable)
+                .limit(params[:per] || default_page_size)
+    end
   end
 
   def default_page_size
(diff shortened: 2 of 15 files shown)


app/controllers/api/events_controller.rb
```diff
@@ -9,13 +9,24 @@ def mark_as_read
...
+    if params[:event_id_gt]
+      Event.joins(:events_groups).
+        where("events_groups.group_id in (?)", current_user.group_ids).
+        where("events.id > ?", params[:event_id_gt]).order(:id)
+    else
+      load_and_authorize(:discussion).items.sequenced
+    end
   end
 
   def page_collection(collection)
```
robguthrie: sorry @gdpelican. please show me how to this better.

app/models/event.rb
```diff
@@ -35,11 +40,21 @@ def trigger!
...
+      related_groups.map {|g| EventsGroup.new(group_id: g.id, event_id: self.id) }
+    )
+  end
 
-  # which communities should know about this event?
-  # Polls override this to look at the communities associated with the poll.
-  def communities
-    Array(eventable&.group&.community)
+  # which groups are privvy to this event?
+  def related_groups
```
robguthrie: also I don't know if we need to include self.discussion here. seems like eventable will always be aware of the groups it belongs to.

group or groups thing could be improved i'm sure.

app/controllers/api/events_controller.rb
gdpelican: This is pointing towards a separate endpoint for me. This new endpoint is more like `events#index`, whereas the existing discussion endpoint is more like `events#for_discussion`

```
def index
  # I'd strongly consider a timestamp here; this feels more like "Give me all the events which happened since I disconnected at <time>"
  instantiate_collection { |collection| collection.where("events.id > ?", params[:event_id]) }
  respond_with_collection
end

def for_discussion
  instantiate_collection do |collection|
    collection.where(discussion: load_and_authorize(:discussion)).sequenced
  end
  respond_with_collection
end

private

def accessible_records
  resource_class.joins(:events_groups)
                .where("events_groups.group_id": current_user.group_ids)
end
```

app/models/event.rb
gdpelican: I think I'd like any of Poll, Discussion, or Group to respond to a `related_groups` method, which returns an array of groups we care about for that model.

```
module HasGuestGroup
  def related_groups
    [guest_group, group].compact
  end
end
```

```
class Group
  def related_groups
    Array(self)
  end
end
```
```
class Membership
  delegate :related_groups, to: :group
end
```

then

```
class Event
  def create_event_groups!
    self.event_groups.import(
      self.eventable.related_groups.map { |group| EventGroup.new(group: group) }
    )
  end
end
```

angular/core/services/connection_service.coffee
```diff
@@ -0,0 +1,8 @@
+angular.module('loomioApp').factory 'ConnectionService', ->
+  new class ConnectionService
```
gdpelican: I don't quite follow the need for this class yet. I think this might belong in MessageChannelService, under the 'subscribe' method

```
new class MessageChannelService
  subscribe: (options = {}) ->
    $http.post('/api/v1/message_channel/subscribe', options).then handleSubscriptions
    PrivatePub.faye (client) ->
      client.on 'transport:up', ->
        FlashService.warning 'You're disconnected!'
      client.on 'transport:down', ->
        FlashService.success 'We're back!'
        Records.events.fetchNew()
```

It might even go into the 'handleSubscriptions' method, ie once we know that we've successfully subscribed to our channels, then we apply our onConnect and onDisconnect methods

app/models/event.rb
robguthrie: Sounds cool.. q: why not just call it groups? rather than related_groups? Seems like that's the natural direction to me.

robguthrie:
thanks for the review + help GD!

PR #4330, opened 2017-10-06
Doc managements

gdpelican:


robguthrie:
resources quick feedback.

need submit on enter for add resource form

suggested helptext for the form::
Share documents, files and links with your group.

dont show see all on context panel, if all resources are already listed?
"see all (1) resources" is ugly text (sorry). Maybe just remove the count and parens?
I added resources to a thread, then navigated back to group, and the see all count was off.

clicking on attached pdf resource from resources page gave 404

looks like you've picked up a bunch of locales json in your diff by accident.

PR #4336, opened 2017-10-10
add markdown parsing to outcome thread item

robguthrie:


gdpelican:
If we do this we need to wrap the textarea in poll_common_outcome_form.haml in an %lmo_textarea like the rest. You may need to disallow uploading attachments on it, but we want the emoji picker and other goodies that come with that directive.

gdpelican:
I've also added the ability here to add attachments to outcomes, because we can't think of a reason for that not to be the case.

PR #4338, opened 2017-10-15
nested comments & accurate mark as read

robguthrie:


robguthrie:
unfinished but:
remove salient concept
remove lastreadsequenceid
track every read sequence id in the minimum set of ranges
need to work on the client and how the sequenceids are sent over next

app/models/discussion_reader.rb
```diff
@@ -53,20 +68,41 @@ def discussion_reader_volume
   end
 
   private
+  def serialize_read_ranges(ranges)
+    # could merge adjancent ranges in the future
+    self.read_sequence_id_ranges = ranges.sort_by{|r| r.first}.map{|r| [r.first,r.last].join(',')}.join(' ')
```
gdpelican: So the data stored here is `1,4 6,10 15,20`, which means you've missed sequence 5 and 11-14?

gdpelican:
```
ranges.sort_by(&:first)
```

gdpelican: What happens if there's a single item 'read'? (and is that even possible?)... it looks like we do `5,5`; do we handle that edge case properly?

app/models/discussion_reader.rb
```diff
@@ -53,20 +68,41 @@ def discussion_reader_volume
...
+  def compute_read_items_count
+    parse_read_ranges.map(&:count).sum
+  end
+
+  def can_merge?(a, b)
+    a.include?(b.first - 1) || a.include?(b.last + 1)
+  end
+
+  def merge(ranges)
+    sorted = ranges.map{|r| [r.first, r.last] }.flatten.sort
```
gdpelican: I might be tempted to make this
```
def merge(*ranges)
  ranges.flatten.map { |r| [r.first, r.last] }.sort
end
```

which makes `merge` an easier method to call
```
merge(mergeable, a, b, c)
```

gdpelican: Also, I feels to me a bit like there may be an object trying to get out here
```
class DiscussionReader
  before_save :consolidate_ranges
  def ranges
    @ranges ||= RangeCollection.new(self.read_sequence_id_ranges)
  end

  def consolidate_ranges
    ranges.reduce.to_s
  end
end

class RangeCollection
  def initialize(rang_str)
    # maybe this takes the string from the database directly and converts it to an array of Ranges internally?
    @ranges = range_str.split(' ').map { |r| Range.new(r.split(',')[0], r.split(',')[1]) }
  end

  def reduce # (from reduce_ranges)
    # returns a new SequenceRangeCollection which has compacted @ranges so that adjacent ranges are merged
  end

  def sorted
    # sort @ranges so they're in order
  end

  def to_s
    # takes the array of ranges and converts them to database-friendly format
    @ranges.map { |r| [r.first, r.last].join(',') }.join(' ')
  end
end
```

app/models/discussion_reader.rb
```diff
@@ -20,19 +22,32 @@ def self.for_model(model, actor = nil)
   end
 
   def update_reader(read_at: nil, volume: nil, participate: false, dismiss: false)
-    viewed!(read_at, persist: false)    if read_at
+    viewed!([], persist: false)         if read_at
```
gdpelican: I think 'read_at' will be useless to us going forward

app/models/discussion_reader.rb
robguthrie: yes that's handled

robguthrie:
@gdpelican this is green now. Last opportunity for a code review tonight (just if you'd like to, I'm feeling happy about, but would appreciate it if you have time), and I'm gonna be aiming for deploy tomorrow morning.

angular/core/components/discussion_form/discussion_form.haml
```diff
@@ -0,0 +1,52 @@
+%md-dialog.lmo-modal__narrow
```
robguthrie: @gdpelican did you remove this file in another commit?

angular/core/components/thread_page/activity_card/activity_card.coffee
```diff
@@ -1,103 +1,63 @@
...
+  scope: {discussion: '=', loading: '='}
   restrict: 'E'
   templateUrl: 'generated/components/thread_page/activity_card/activity_card.html'
-  replace: true
-  controller: ($scope, $location, $mdDialog, $rootScope, $window, $timeout, Records, AppConfig, AbilityService, PaginationService, LoadingService, ModalService) ->
-
+  controller: ($scope) ->
     $scope.$on 'fetchRecordsForPrint', (event, options = {}) ->
-      $scope.loadEvents(per: Number.MAX_SAFE_INTEGER).then ->
+      $scope.tw.loadAll().then ->
```
robguthrie: Rob fix this

gdpelican:
Ack... I'd really rather have more time to review than getting a message at
8pm the night before. Can we wait until tomorrow afternoon?

angular/core/components/thread_page/event_children/event_children.scss
```diff
@@ -0,0 +1,6 @@
+// .thread-item--indent {
```
robguthrie: rob remove this

robguthrie:
yea sure, no worries

app/models/discussion_reader.rb
```diff
@@ -25,23 +25,33 @@ def self.for_model(model, actor = nil)
...
+    save!                               if changed?
   end
 
-  def viewed!(read_at, persist: true)
-    return if self.last_read_at && self.last_read_at > read_at
-    assign_attributes(read_attributes(read_at))
-    EventBus.broadcast('discussion_reader_viewed!', self)
+  def viewed!(ranges = [], persist: true)
+    # byebug
+    # ranges = RangeSet.to_ranges(ranges)
```
gdpelican: Do you want to keep this in to ensure you have a rangeset before proceeding? Good ol' ruby type check guarding, heh. I'm fine either way.

app/models/event.rb
```diff
@@ -8,13 +9,13 @@ class Event < ActiveRecord::Base
 
   scope :sequenced, -> { where.not(sequence_id: nil).order(sequence_id: :asc) }
   scope :chronologically, -> { order(created_at: :asc) }
+  scope :excluding_sequence_ids, -> (ranges) { where RangeSet.to_ranges(ranges).map {|r| "(sequence_id NOT BETWEEN #{r.first} AND #{r.last})"}.join(' AND ') }
```
gdpelican: Looks like we don't use this scope?

app/models/event.rb
```diff
@@ -23,6 +24,10 @@ class Event < ActiveRecord::Base
 
   acts_as_sequenced scope: :discussion_id, column: :sequence_id, skip: lambda {|e| e.discussion.nil? || e.discussion_id.nil? }
 
+  def child_count
```
gdpelican: Shouldn't this be take care of with a default in the database?

```
add_column :events, :child_count, :integer, default: 0, null: false
```
?

app/models/events/discussion_description_edited.rb
```diff
@@ -1,3 +1,4 @@
+# confirm this is no longer used
```
gdpelican: It's not, blow it up :)

app/models/events/discussion_title_edited.rb
```diff
@@ -1,3 +1,4 @@
+# confirm: we don't use this?
```
gdpelican: Same here, this can go.

app/models/events/poll_expired.rb
```diff
@@ -6,6 +6,7 @@ class Events::PollExpired < Event
   def self.publish!(poll)
     create(kind: "poll_expired",
            eventable: poll,
+           parent: poll.created_event,
```
gdpelican: Seems like there's a sensible default of 
```
eventable.created_event
```

for this. In any case I'm simplifying the event creation process a bit in my branch, and will move this stuff over.

app/models/poll.rb
```diff
@@ -113,6 +113,18 @@ def undecided_count
 
   alias_method :user, :author
 
+  def parent_event
+    if discussion
```
gdpelican: discussion&.created_event

app/models/poll.rb
```diff
@@ -113,6 +113,18 @@ def undecided_count
...
+      discussion.created_event
+    else
+      nil
+    end
+  end
+
+  def created_event
+    events.find_by(kind: "poll_created")
```
gdpelican: Symbol here too?

I wonder if there's some sort of future concern in here around objects which have creation events / sequence ids

app/serializers/events/parent_event_serializer.rb
```diff
@@ -0,0 +1,9 @@
+class Events::ParentEventSerializer < Events::BaseSerializer
```
gdpelican: Do we need this serializer?

app/services/event_parent_migrator.rb
```diff
@@ -0,0 +1,33 @@
...
+        parent_event.update_child_count
+      end
+    end
+  end
+
+  def self.assign_stance_parents(group)
+    total_polls = group.polls.count
+    group.polls.find_each do |poll|
+      puts "#{poll.id} of #{total_polls} polls"
+      next unless parent_event = poll.events.find_by(kind: "poll_created")
```
gdpelican: can probably use our new `poll.created_event` and `comment.created_event` methods here, no?

config/initializers/event_bus.rb
```diff
@@ -60,26 +60,19 @@
...
+                'discussion_dismiss',
+                'discussion_mark_as_seen') do |reader|
     MessageChannelService.publish(
       ActiveModel::ArraySerializer.new([reader], each_serializer: DiscussionReaderSerializer, root: :discussions).as_json,
       to: reader.user
     )
   end
 
-  config.listen('discussion_mark_as_seen') do |discussion, actor|
-    MessageChannelService.publish(
```
gdpelican: I don't think you want to lose the broadcast of `mark_as_seen` to the group, otherwise it won't live update for people other than you.

spec/extras/range_set_spec.rb
```diff
@@ -0,0 +1,53 @@
+require_relative '../../app/extras/range_set.rb'
```
gdpelican: Would probably be a real good idea to pull this out to a gem. :)

spec/services/group_service_spec.rb
```diff
@@ -13,11 +13,6 @@
       expect(group.reload.creator).to eq user
     end
 
-    it 'creates a new guest group' do
```
gdpelican: Hm did this break? Wouldn't expect it to..

angular/core/components/lmo_textarea/lmo_textarea.scss
```diff
@@ -3,6 +3,7 @@
 }
 
 .lmo-textarea__helptext {
+  margin-left: 5px;
```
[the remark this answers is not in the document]
robguthrie: this is what it needs to look correct

angular/core/components/document/list/document_list.haml
```diff
@@ -1,6 +1,5 @@
-%section.document-list
+%section.document-list{ng-if: "model.hasDocuments()"}
```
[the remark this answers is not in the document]
robguthrie: I'll remove this.

angular/core/components/thread_card/thread_card.scss
```diff
@@ -0,0 +1,3 @@
+.thread-card {
```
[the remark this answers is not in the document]
robguthrie: the context is part of this card, and has no 'order' in the page now.

angular/core/components/thread_page/add_comment_panel/add_comment_panel.coffee
```diff
@@ -0,0 +1,32 @@
+angular.module('loomioApp').directive 'addCommentPanel', (AbilityService, ModalService, AuthModal, Session, ScrollService, $timeout) ->
+  scope: {eventWindow: '=', parentEvent: '='}
+  restrict: 'E'
+  templateUrl: 'generated/components/thread_page/add_comment_panel/add_comment_panel.html'
+  controller: ($scope) ->
+    $scope.discussion = $scope.eventWindow.discussion
+    $scope.actor = Session.user()
+    $scope.isLoggedIn = AbilityService.isLoggedIn
```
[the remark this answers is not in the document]
robguthrie: same.

angular/core/components/thread_page/add_comment_panel/add_comment_panel.coffee
```diff
@@ -0,0 +1,32 @@
...
+    $scope.close = -> $scope.show = false
+    $scope.isReply = false
+
+    $scope.indent = -> $scope.eventWindow.useNesting && $scope.isReply
+
+    $scope.$on 'replyToEvent', (e, event) ->
+      # if we're in nesting and we're the correct reply OR we're in chronoglogical, always accept parentComment
+      if (!$scope.eventWindow.useNesting) || ($scope.parentEvent.id == event.id)
+        $scope.show = true
+        $timeout ->
```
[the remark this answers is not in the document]
robguthrie: because a digest cycle needs to happen before this will work.. :(

angular/core/components/thread_page/add_comment_panel/add_comment_panel.coffee
```diff
@@ -0,0 +1,32 @@
...
+          $scope.isReply = true
+          $scope.$broadcast 'setParentComment', event.model()
+
+      ScrollService.scrollTo('.add-comment-panel textarea', {bottom: true, offset: 200})
+
+    $scope.$on 'commentSaved', ->
+      if $scope.parentEvent == $scope.discussion.createdEvent()
+        $scope.parentComment = null
+      else
+        $scope.close()
```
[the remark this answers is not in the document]
robguthrie: there is an X button on a reply form which also uses it

angular/core/css/0_lmo_flex.scss
```diff
@@ -2,6 +2,16 @@
   display: flex;
 }
 
+.lmo-flex--row {
```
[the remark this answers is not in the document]
robguthrie: this is an angular material directive. I'd rather not depend on that when a plain css approach will do. As you mention we should do a sweep of these soon.

angular/core/services/ability_service.coffee
```diff
@@ -1,10 +1,10 @@
 angular.module('loomioApp').factory 'AbilityService', (AppConfig, Records, Session) ->
   new class AbilityService
 
-    isLoggedIn: ->
+    isLoggedIn: =>
```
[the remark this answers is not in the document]
robguthrie: np

angular/core/services/base_event_window.coffee
```diff
@@ -0,0 +1,46 @@
...
+    anyPrevious: -> @numTotal() > 0 && @firstLoaded() > 1
+    numPrevious: -> @numTotal() > 0 && @firstLoaded() - 1
+    anyNext:     -> @numTotal() > @lastLoaded()
+    numNext:     -> @numTotal() - @lastLoaded()
+    anyMissing:  -> @numTotal() > @numLoaded()
+    numMissing:  -> @numTotal() - @numLoaded()
+
+    # min and max are the minimum and maximum values permitted in the window
+    setMin: (val) ->
+      @min = val
```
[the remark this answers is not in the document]
robguthrie: nice, thanks

angular/core/services/scroll_service.coffee
```diff
@@ -5,6 +5,8 @@ angular.module('loomioApp').factory 'ScrollService', ($timeout) ->
     $timeout ->
       elem      = document.querySelector(target)
       container = document.querySelector(options.container or '.lmo-main-content')
+      if options.bottom
```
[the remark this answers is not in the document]
robguthrie: We're trying to alight the element to the bottom of the viewport.

app/controllers/api/events_controller.rb
```diff
@@ -1,21 +1,33 @@
 class API::EventsController < API::RestfulController
   include UsesDiscussionReaders
```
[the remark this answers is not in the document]
robguthrie: I think we do. let's check and confirm

app/controllers/api/events_controller.rb
```diff
@@ -1,21 +1,33 @@
...
+
   def accessible_records
-    load_and_authorize(:discussion).items.sequenced
+    records = load_and_authorize(:discussion).items.
+              includes(:user, :discussion,
+                       eventable: [:reactions],
+                       parent: [:user, {eventable: [:reactions]}]).uniq
+
+    # using :from and :comment_id is deprecated
+    # use :min_sequence_id instead of :from, and don't use :comment_id
```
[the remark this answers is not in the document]
robguthrie: yea I went back on this actually. 
I think the idea of a generic notation for min_column_name max_column_name is cool though and makes things seem clearer to me... from gets ambiguous when you're talking about potentially a few different columns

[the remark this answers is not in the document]
robguthrie: I think we want to look at reactions and attachement caches that the serializer can use at will, rahter than includes statements which I don't think actually work all the time anyway.
Before I started the work I saw I big n+1 problem on reactions and attachments, this improves the situation somewhat but I think the right solution is to solve this at the serializer level.

app/models/concerns/events/position.rb
```diff
@@ -0,0 +1,44 @@
...
+    update_counter_cache :parent, :child_count
+  end
+
+  private
+  def set_depth
+    self.depth = parent ? (parent.depth + 1) : 0
+  end
+
+  def refresh_order_value
+    self.position = self.class.select(:position).find(self.id).send(:position)
```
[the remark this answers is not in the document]
robguthrie: cool thanks

app/models/discussion.rb
```diff
@@ -61,8 +44,7 @@ class Discussion < ActiveRecord::Base
 
   has_many :events, -> { includes :user }, as: :eventable, dependent: :destroy
 
-  has_many :items, -> { includes(:user).where(kind: THREAD_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
-  has_many :salient_items, -> { includes(:user).where(kind: SALIENT_ITEM_KINDS).order('created_at ASC') }, class_name: 'Event'
+  has_many :items, -> { includes(:user).order('events.id ASC') }, class_name: 'Event'
```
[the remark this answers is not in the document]
robguthrie: Shall we drop the discussion_id from those events then?

app/models/discussion.rb
```diff
@@ -137,6 +116,18 @@ def body
     self.description
   end
 
+  def ranges
+    RangeSet.parse(self.ranges_string)
+  end
+
+  def first_sequence_id
+    (ranges.first || []).first || 0
```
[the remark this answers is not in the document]
robguthrie: nice, I thought I was being clever with what I did, but this is even cleverer

app/models/discussion_reader.rb
robguthrie: na, we dont need it, which is nice

app/models/event.rb
robguthrie: I'd like to keep it for the future though, sentimental attachement. can I leave it commented for a while?

app/services/event_parent_migrator.rb
robguthrie: done

PR #4340, opened 2017-10-17
Ensure modals with long content display correctly

gdpelican:


gdpelican:
@robguthrie This is ready for another peek.

I've added a scroll indicator here to let users know that there's more content to see on a longer modal now. I've also removed the 90% height restriction for small screens, which should help a bit. We can also make the modal header a bit smaller for tiny screens if we feel we need to.

screen shot 2017-10-19 at 12 22 50 am

robguthrie:
amazing!

PR #4342, opened 2017-10-22
Automatically link URLs starting with "www." (Fixes #4196)

### LINKED ISSUE #4196: Links not automatically linking in discussion context
www.something.com is not auto-linking in discussion context 

http://www.something.com does though

JustusFT:
Marked should now auto-link urls that start with "www."; The url does not need to be prefixed with "http://" or "https://" to be auto-linked.
e.g. www.google.com
Fixes #4196


robguthrie:
Thanks for this PR! James and I are both on Holiday this week, but we really appreciate your efforts and will review this next week.

gdpelican:
Closing this for now (mildly) in favour of #4344 , although I'd like to see if we can find a better third way. Thanks for the effort @JustusFT !

PR #4344, opened 2017-10-26
Automatically link URLs (Fixes #4196)

### LINKED ISSUE #4196: Links not automatically linking in discussion context

perkyguy:
This PR uses the customized Marked `parse` function, already defined in `angular/core/config/marked.coffee`, and modifies the default regular expression for URLs. Previously it was relying on having an `http` or `https` prepending the URL, now it is able to pick up a wide array of URLs, including IPs, with query string, etc. The Marked renderer also is modified to prepend an `http` to found URLs which do not contain the protocol at the beginning of the address. 

There is another PR, https://github.com/loomio/loomio/pull/4342, which aims at doing the same, but it is limited in scope, the regex used here should give a little more functionality.

### DIFF
--- a/angular/core/config/marked.coffee
+++ b/angular/core/config/marked.coffee
@@ -15,12 +15,15 @@ angular.module('loomioApp').config (markedProvider) ->
       _super.heading(emojione.shortnameToImage(text), level, text)
 
     renderer.link      = (href, title, text) ->
+      if marked.InlineLexer.rules.gfm.url.test(href) and !/^http/.test(href)
+        href = 'http://' + href
       _super.link(href, title, text).replace('<a ', '<a rel="noopener noreferrer" target="_blank" ')
 
     renderer
 
   _parse = marked.parse
   marked.parse = (src, opt, callback) ->
+    marked.InlineLexer.rules.gfm.url = /^((https?:\/\/)?([a-z0-9+!*(),;?&=.-]+(:[a-z0-9+!*(),;?&=.-]+)?@)?([a-z0-9\-\.]*)\.(([a-z]{2,4})|([0-9]{1,3}\.([0-9]{1,3})\.([0-9]{1,3})))(:[0-9]{2,5})?(\/([a-z0-9+%-]\.?)+)*\/?(\?[a-z+&$_.-][a-z0-9;:@&%=+/.-]*)?(#[a-z_.-][a-z0-9+$%_.-]*)?)/i
     src = src.replace(/<img[^>]+\>/ig, "")
     src = src.replace(/<script[^>]+\>/ig, "")
     return _parse(src, opt, callback)


robguthrie:
Thanks for this PR! James and I are both on Holiday this week, but we really appreciate your efforts and will review this next week.

gdpelican:
I keep wondering why we need to handle this, and not marked. Is there a way that we can submit this regex change as a PR to marked (possibly with an option flag) and not have to pollute our already-a-bit-mucky marked lexer with another regex?

robguthrie:
Marked hasn't had a commit since 21 January and there are 166 PR's waiting. I wonder if there is an actively maintained fork?

perkyguy:
[Here's a fork](https://github.com/8fold/marked) that I found that started up in September, looks like they are trying to actively maintain it. Or there is [remarked](https://github.com/nlf/remarked)? They both _should_ be drop in replacements?

kxmo:
Hi all,

I've been looking into this and thought I'd share what I've found.

## A solution
[Marked](https://github.com/markedjs/marked) appears to now be actively maintained, and supports links starting with www (using version 0.3.17, the most recent):
```
var markdown = require('marked');
console.log(markdown('http://example.com')); // <p><a href="http://example.com">http://example.com</a></p>
console.log(markdown('http://www.example.com')); // <p><a href="http://www.example.com">http://www.example.com</a></p>
console.log(markdown('www.example.com')); // <p><a href="http://www.example.com">www.example.com</a></p>
console.log(markdown('example.com')); // <p>example.com</p>
```
## Other options
### @perkyguy's links
- The first link is no longer up, and it doesn't seem to be under a different repo on the account.
- The second link hasn't been updated in 4 years, but has no open issues or PRs - this is probably fine because Markdown doesn't change that often. However, it doesn't solve this problem in it's current form: 
```
var markdown2 = require('remarked');
console.log(markdown2('http://example.com')); // <p><a href="http://example.com">http://example.com</a></p>
console.log(markdown2('http://www.example.com')); // <p><a href="http://www.example.com">http://www.example.com</a></p>
console.log(markdown2('www.example.com')); // <p>www.example.com</p>
console.log(markdown2('example.com')); // <p>example.com</p>
```
### @perkyguy's solution
Neither updated `marked` nor `remarked` support IP addresses, but your solution does. While it's a useful addition (and could be added as well as using the new marked) I would expect (without any research) that there are only a small number of users sending or receiving IP addresses if any.

## The problem generally
This problem is easy to solve poorly (link \*.\* and have the browser sort it out) and really hard to solve well. There are [many](http://data.iana.org/TLD/tlds-alpha-by-domain.txt) valid TLDs and it's difficult to know whether a user intended to link something. For example `... think that's a great idea.Bet you could ...` shouldn't be linked but idea.bet is a valid address and should be linked in other circumstances.

I think stopping at www\.* is a good compromise over linking every start of a sentence where a user forgot to type a space. Interestingly Github took the opposite approach, and linked `www.*` (which I escaped) as `http://www/`.

gdpelican:
@kxmo Thanks for the detailed report, very useful.

Is your conclusion that we should update our marked library, which will make the basic case we're looking to solve (ie `www.google.com`) link automatically?

kxmo:
@gdpelican That's right!

I'm looking at doing that now, and I understand the required changes are:

- Update https://github.com/loomio/angular-marked/blob/10e2385bf6cfc73fb7810193fcde5b8f2b13e099/package.json#L10 to version 0.3.17
- Bump patch version in both angular-marked and https://github.com/loomio/loomio/blob/bf384ef412cd2fb476b44f1898d77e9f499ad719/client/package.json#L56
- Add tests to https://github.com/loomio/loomio/blob/master/spec/models/document_spec.rb

Is there anything else that needs to change?

Is the angular-marked version in loomio/client/package.json (1.2.4) supposed to be the same as angular-marked/package.json (1.2.2)?

gdpelican:
Alright, closing this one in favor of https://github.com/loomio/loomio/pull/4582, which should fix the auto-linking URLs, as well as Marked can manage for now.

PR #4352, opened 2017-10-29
turn off Logging for TEST enviroment

skatkov:
Noticed your testsuite execution times and decided to help you out with it a bit. ;-)

This is a very simple trick -- turn off your Logging and get a 6% speed up to testsuite.


gdpelican:
Thanks a bunch! :D

PR #4353, opened 2017-10-29
checkout as orphan branch to make deployment faster

robguthrie:
untested (sorry!) but if the command is correct this will save bandwidth when deploying over mobile connections (which I've been doing recently)


gdpelican:
Didn't know about orphan branches, but if it deploys I'm happy

robguthrie:
I'll test it soon

robguthrie:
tested ok

PR #4354, opened 2017-10-29
Don't pick up login tokens from the current user as pending

gdpelican:


robguthrie:
Looks like a good way to do it.

PR #4359, opened 2017-10-31
Use controller force_ssl rather than config force_ssl 

robguthrie:
The advantage of force_ssl in the application_controller rather than application.rb is we can bypass it for internal network requests (ie: inbound email) which can't go through the ssl layer.

We're also moving to dynamicly compiling assets the first time they're requested, which fixes a problem I was facing with the managed installs and knowing asset paths.

This only affects email (and marketing) assets.


gdpelican:
:+1:

PR #4363, opened 2017-11-01
update moment locale from user.locale

### LINKED ISSUE #4349: {{time}} template is not using momentjs locales
In client translations, some strings use the {{time}} template (closing in, closed, etc.). This time template can be displayed as relative time, instead of the timestamp, but the relative time is always in english (in or ago etc.)

It seems that [momentjs supports a lot of locales](https://momentjs.com/), but Loomio requires "moment" and not "moment-with-locales": for example, in French, Loomio displays "3 years ago" or "in 3 days" instead of "il y a 3 ans" or "dans 3 jours" as it should with locales.

Thanks to @uportalis and @swergas for pointing this out and in #3898.

robguthrie:
fixes #4349


angular/core/components/profile_page/profile_page_controller.coffee
```diff
@@ -2,15 +2,16 @@ angular.module('loomioApp').controller 'ProfilePageController', ($scope, $rootSc
   $rootScope.$broadcast('currentComponent', { titleKey: 'profile_page.profile', page: 'profilePage'})
 
   @showHelpTranslate = => AppConfig.features.help_link
-  
+
   @init = =>
     return unless AbilityService.isLoggedIn()
     @user = Session.user().clone()
-    $translate.use(@user.locale)
```
gdpelican: Should be able to drop $translate from the list of dependencies here

angular/tasks/config/vendor.yml
```diff
@@ -2,7 +2,7 @@
 path: node_modules
 js:
   - bowser/bowser.js
-  - moment/moment.js
+  - moment/min/moment-with-locales.min.js
```
gdpelican: Hm. This adds about 200k to our payload and I'm not convinced it's worth it.

screen shot 2017-11-01 at 3 33 11 pm

robguthrie:
@gdpelican thanks for the suggestion, hows this? Those two polyfils are pretty light.

angular/core/services/session.coffee
```diff
@@ -19,6 +19,13 @@ angular.module('loomioApp').factory 'Session', ($rootScope, $location, $translat
 
     user
 
+  setLocale: (locale) ->
+    $translate.use(locale)
+    lc_locale = locale.toLowerCase()
+    fetch("/client/moment_locales/#{lc_locale}.js").then((resp) -> resp.text()).then (data) ->
+      eval(data)
```
robguthrie: @gdpelican have a think about if you're ok with this eval statement. I think it's safe, but would like you to double check.

PR #4370, opened 2017-11-02
rewrite preferred locale code (needs review)

robguthrie:
we have a huge number of locale files, only a few have more than 50% of the app strings.

define a list of supported locales
define a way to fall back from a specifc to a more general locale, or to a similar locale

eg: 'pt-BR' falls back to 'es', and 'en-US' falls back to 'en'
however zh_TW is supported, so other zh dialects should fall back to zh_TW

let me know what other rspec tests I should write here.

I've also removed lots of duplicate locale files


app/helpers/locales_helper.rb
```diff
@@ -1,74 +1,80 @@
...
-    end
+  def strict_filter(locales)
+    Array(locales).compact.select do |locale|
+      supported_locales.select{|supported_locale| supported_locale == normalize(locale)}.first
+    end.first
   end
 
-  private
+  def fuzzy_filter(locales)
+    Array(locales).compact.map do |locale|
```
gdpelican:
```
def fuzzy_locales_for(locales)
  Array(locales).map { |l| [normalize(l), strip_dialect(l), fallback_for(l)] }.flatten
end

def fuzzy_filter(locales)
  (fuzzy_locales_for(locales) & supported_locales).first
end
```

app/helpers/locales_helper.rb
```diff
@@ -1,74 +1,80 @@
...
   end
 
-  def locales_from_browser_detection
-    parser = HttpAcceptLanguage::Parser.new(request.env["HTTP_ACCEPT_LANGUAGE"])
-    locales = parser.user_preferred_languages +
-              parser.user_preferred_languages.map { |l| l.split('-').first } # to catch locales like fr-fr, or en-nz
-    filter_locales(locales, detectable_locales)
+  def user_selected_locale
+    return nil unless current_user&.is_logged_in?
+    current_user.selected_locale
```
gdpelican: current_user always returns something, so it won't be nil, it'll be a LoggedOutUser
So I think we'd rather just have LoggedOutUser respond to `selected_locale`? (I think it might do this already)

Then we can just called `current_user.selected_locale` instead of having this method

gdpelican: Same for user_detected_locale; in fact it might make sense to just push down a method onto user (and Logged out user like `best_locale`

```
# user.rb
def best_locale
  selected_locale || detected_locale
end

# logged_out_user.rb
nil_methods :best_locale
```

app/helpers/locales_helper.rb
```diff
@@ -1,74 +1,80 @@
...
+    I18n.locale = preferred_locale
     yield if block_given?
     save_detected_locale
   end
 
-  def selectable_locales
-    Loomio::I18n::SELECTABLE_LOCALES
+  def preferred_locale
+    strict_filter(params[:locale]) ||
+    fuzzy_filter(user_selected_locale) ||
```
gdpelican: Hey, do you think we could do something like

```
strict_filter(params[:locale]) ||
fuzzy_filter([
  current_user.selected_locale,
  browser_detected_locales,
  current_user.detected_locale,
  I18n.default_locale
].flatten)
```

so that we only have to iterate through one array?

app/helpers/locales_helper.rb
```diff
@@ -1,74 +1,80 @@
...
   end
 
-  def detectable_locales
-    Loomio::I18n::DETECTABLE_LOCALES
-  end
-
-  def rtl_locales
-    Loomio::I18n::RTL_LOCALES
+  def supported_locales
+    YAML.load_file(Rails.root.join('config', 'locales.yml'))["supported_locales"]
```
gdpelican: I'd recommend throwing this into AppConfig, like this:

```
# app_config.rb
CONFIG_FILE = %w(... locales)
```
then
```
AppConfig.locales['supported_locales']
```

which means we likely don't need this method either.

app/helpers/locales_helper.rb
```diff
@@ -1,74 +1,80 @@
...
-    filter_locales(current_user.selected_locale, selectable_locales)
+  def fallback_for(locale)
+    fallbacks[locale] || fallbacks[strip_dialect(locale)]
   end
 
-  def locales_from_saved_browser_detection
-    return unless current_user&.is_logged_in?
-    filter_locales(current_user.detected_locale, selectable_locales)
+  def fallbacks
+    YAML.load_file(Rails.root.join('config', 'locales.yml'))["fallbacks"]
```
gdpelican: This can get called through AppConfig via

```
AppConfig.locales['fallbacks']
```

robguthrie:
Thanks for the perspective @juliencarnot - I'll stop using it to fallback across languages, and instead just fallback to dialects where there is a translation. ie fallback from zh to zh_TW and nl to nl_TW.

Then we'll make better use of the browser detected locales array anyway.

I'm not feeling the energy to change much more right now, but I would like to link to our transifex page and make a short CTA for translation support on the profile page.

juliencarnot:
@robguthrie sounds good to me! I'll copy my proposal of language switch in an issue, for future evaluation.

PR #4371, opened 2017-11-02
fixes incorrect reference to string in locales

juliencarnot:
This should allow to display the locale string instead of its name.


gdpelican:
Thanks @juliencarnot !

PR #4379, opened 2017-11-03
adds the missing sizes of (optimized) icons and set theme color to the manifest.json

juliencarnot:
just marginal improvements towards progressive web app support...


app/controllers/manifest_controller.rb
```diff
@@ -9,11 +9,31 @@ def show
       orientation: 'portrait',
       start_url: '/dashboard',
       background_color: '#ffffff',
-      theme_color: '#ffffff',
+      theme_color: '#AED9EB',
       icons: [{
-        src:   [root_url.chomp('/'), AppConfig.theme[:icon_src]].join(''),
```
gdpelican: Since this is a white-labelled app (it has customizable branding based on which instance we're running), we'll want to keep the reference to the `AppConfig.theme` values in here... and likely expand a bit so that the theme color / background color looks at the current theme as well.

juliencarnot: Right, thanks for the explanation. Should be better now...

PR #4380, opened 2017-11-03
add seen by count to discussion context

robguthrie:
Adds a "Seen by " to the metadata line on discussion context panel.

Issues:
I'd like to only update discussion readers count on reader create... how?

And in dev mode, creating a discussion ends up with Seen by 3, including Loomio Bot and Jennifer Grey.. just sometimes. WTF?

image

### DIFF
--- a/angular/core/components/thread_page/context_panel/context_panel.haml
+++ b/angular/core/components/thread_page/context_panel/context_panel.haml
@@ -45,9 +45,9 @@
     %span.nowrap.context-panel__discussion-privacy.context-panel__discussion-privacy--public{ng-show: "!discussion.private"}>
       %i.mdi.mdi-earth>
       %span{translate: "common.privacy.public"}>
-    %span{ng-show: "discussion.discussionReadersCount > 0"}>
+    %span{ng-show: "discussion.seenByCount > 0"}>
       %span{aria-hidden: "true"}>·
-      %span.context-panel__seen_by_count{ translate: "thread_context.seen_by_count", translate-value-count: "{{discussion.discussionReadersCount}}"}
+      %span.context-panel__seen_by_count{ translate: "thread_context.seen_by_count", translate-value-count: "{{discussion.seenByCount}}"}
     %outlet{name: "after-thread-title", model: "discussion"}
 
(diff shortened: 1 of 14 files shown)


robguthrie:
Also remove specific font size from timeago, because it looks better when it just inherits.

gdpelican:
Which route(s) show seen by 3, as opposed to the expected seen by 1?

robguthrie:
@gdpelican setup_discussion

gdpelican:
My hunch is that readers are being created when an email is being sent out,
rather than when it's read for the first time. I think you want to
constrict to where last_read_at != nil

gdpelican:
Seems like a good enough reason to keep last_read_at around past nested
comments.

gdpelican:
Also you mmmay want to consider live updating that seen by number through a memo or something. 

Would be cool to see your post get acknowledged in real time, plus there's the weird thing of '3 people have read a discussion. I visit it for the first time. Because I 'read' it after I visit the page, those 3 people don't include me. When I reload the page, it'll say '4 people have read it', which will include me.'

gdpelican:
For the only on create thing, it looks like `update_counter_cache` passes options through to each of the callbacks

```
def update_counter_cache(association, column_name, options = {})
  method = # make a callback method
  after_create method, options
  after_update method, options
  after_destroy method, options
end
```

So something like

```
update_counter_cache :discussion, :discussion_readers_count, if: :update_cache?
def update_cache?
  new_record? || destroyed?
end
```
might work.

gdpelican:
https://github.com/cedric/custom_counter_cache/pull/23

I think best might be to point to my branch for now; it may take him a bit to respond and/or merge. I have a hunch it's a really good idea to go back and apply `except: :update` for all the counters we can; it slows down a lot of our update actions.

robguthrie:
Thanks for the help, @gdpelican. The live update should just work, as anything that references a discussion should include it in it's serialisation.  I'll finish it up tomorrow.

gdpelican:
Hm, looks like we serialize discussion reader updates to the user, but not
to the discussion's channel (event_bus.rb#70). So I guess you'd see your
own come in, but I still think it'd be nice to see that number live update
as more people read it in real time.

robguthrie:
oh right! I'm not sure I mind about that experience, given that it will be updated when other comments etc come though, and when you visit the page, I think I'm ok with it not updating on new reads for now.

Issue #4381, opened 2017-11-03
Muting explanation is not displayed but loads images on dashboard page


juliencarnot:
The dashboard page loads [loomio/public/img/mute-small.png](https://github.com/loomio/loomio/blob/master/public/img/mute-small.png) 4.36 KB  and [loomio/public/img/mute.png](https://github.com/loomio/loomio/blob/master/public/img/mute.png) 20.7 KB, both on desktop and mobile.

These screen captures look outdated and I couldn't find where and how they are displayed, if they are. From the following code, it looks like it should be showed on the muted threads page if the user has no muted threads, but it doesn't:

![capture du 2017-11-03 12-05-25](https://user-images.githubusercontent.com/3848493/32370979-56075776-c08f-11e7-8eeb-6db79005b55b.png)

[/angular/core/components/dashboard_page/dashboard_page.haml](https://github.com/loomio/loomio/blob/master/angular/core/components/dashboard_page/dashboard_page.haml) line 21:
`.dashboard-page__explain-mute{ng-show: "dashboardPage.filter == 'show_muted' && !dashboardPage.userHasMuted()"}
          %p
            %strong{translate: "dashboard_page.explain_mute.title"}
          %p{translate: "dashboard_page.explain_mute.explanation_html"}
          %p{translate: "dashboard_page.explain_mute.instructions"}
          .dashboard-page__mute-image--large{ng-if: "dashboardPage.showLargeImage()"}
            %img{src: "/img/mute.png"}
          .dashboard-page__mute-image--small{ng-if: "!dashboardPage.showLargeImage()"}
            %img{src: "/img/mute-small.png"}
          %p{translate: "dashboard_page.explain_mute.see_muted_html"}`

Can this code and resources be safely removed?

robguthrie:
if we change ng-show to ng-if then it wont load the resources when they're not used. I don't think.

When you first mute a thread, we display a modal, are those images used then?

I think I'm ok with removal though. @gdpelican what's your take?

juliencarnot:
You were right, it displays only before the user has used the feature, which is a good feature:
![nevermuted](https://user-images.githubusercontent.com/3848493/32403996-62bea6aa-c147-11e7-89d0-6274920dff4c.png)

I'll PR the ng-show to ng-if change and updated/optimized captures.

juliencarnot:
[...]

PR #4385, opened 2017-11-04
replaces ng-show by ng-if to remove from DOM when test is false, opti…

juliencarnot:
…mizes captures
See #4381


gdpelican:
Sssseems good to me!

PR #4388, opened 2017-11-04
Add anonymous voting

gdpelican:
I threw this together in about 2 hours and am not certain how I feel about it yet... but the experience so far is really great.

From Patrick's point of view:
screen shot 2017-11-05 at 2 01 20 am

From Jenn's point of view:
screen shot 2017-11-05 at 2 02 15 am

Option on the form:
screen shot 2017-11-05 at 2 06 23 am

Things to do:
- Write tests around when to not serialize out stance participants
- Update emails to reflect anonymous votes
- Ensure that notifications (ie 'X voted on your proposal') don't leak participant info (or maybe make those options exclusive?)

NB that I've excluded the anonymous option for time poll, because I couldn't imagine that ever making sense.


juliencarnot:
I added my thoughts about this on:
https://www.loomio.org/d/gUbpokjN/a-secret-vote

PR #4394, opened 2017-11-06
404 on guest groups

gdpelican:
..so that we don't get lots of errors about loading things for them.

We'll also want to figure out _why_ people are visiting guest groups, and fix that UI hiccup as well.


robguthrie:
LGTM! thanks!

Issue #4339, opened 2017-10-16
ranked choice display bug


robguthrie:
![image](https://user-images.githubusercontent.com/486367/31597017-93621e62-b2a2-11e7-8604-9e4408eea977.png)

I can circle back and fix soon if you'd rather not

gdpelican:
I see an easy fix for truncating the long titles, but a quick look didn't yield a way to break them into multiple lines the way we do for dot votes / polls. Which is the right fix here?

robguthrie:
I think we want them to all work the same way. so I'd +1 investigating how to break lines.

PR #4400, opened 2017-11-08
removes white-space:nowrap to allow long option wrapping

juliencarnot:
should fix https://github.com/loomio/loomio/issues/4339

### DIFF
--- a/angular/core/components/poll/common/ranked_choice_chart/poll_common_ranked_choice_chart.coffee
+++ b/angular/core/components/poll/common/ranked_choice_chart/poll_common_ranked_choice_chart.coffee
@@ -31,4 +31,4 @@ angular.module('loomioApp').directive 'pollCommonRankedChoiceChart', (AppConfig,
 
     $scope.styleData = (option) ->
       'background-image': backgroundImageFor(option)
-      'width': percentageFor(option)
+      'background-size': percentageFor(option)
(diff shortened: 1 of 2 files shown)


robguthrie:
Thanks @juliencarnot. 
I would love to just merge this but I don't know what the effects are. 

Have you tested in some browsers, if so, which ones? I love screenshots if you're able to snap one.

We "officially" support ie11 (just barely), chrome, firefox and safari, from mobile (eg: iphone4 size) to desktop sizes. Usually we just resize the desktop browser window to test how things look.

Are you able to give some feedback on how this works? Sorry to be a drag

juliencarnot:
Sure, I tested in Firefox and Chromium:
before:
before

after:
after

I don't have access to IE/Edge, but I'm pretty confident the text will wrap as it does for other polls!

juliencarnot:
hum, hadn't notice the mess in the last column, will look into it and try to find a solution

juliencarnot:
I think this is a better solution:
afterafter

background-size is supported by IE 11 and all modern browsers, it allows the text to use 100% of the space and not only of the bar.

robguthrie:
thanks for sorting this out... really helpful. Yes, background-size has been the successful method for other bar charts.

gdpelican:
Thanks heaps @juliencarnot :)

PR #4407, opened 2017-11-12
update versions and add a few details for installing on linux

juliencarnot:
updates versions in doc and helping to check for updates, and adds instructions for linux psql


gdpelican:
This is perfect, thank you!

PR #4414, opened 2017-11-13
dev doc: removes unnecessary npm install, adds yarn

juliencarnot:
After the [details](https://www.loomio.org/d/iZxd3o9f/common-errors-on-update-from-upstream) from @robguthrie


robguthrie:
Good improvement thank you.
Not sure why it's conflicting rn.
Also a minor thing, I don't think try `gulp compile` again adds much, one could just try gulp dev again.

gdpelican:
Thanks again for your work @juliencarnot !

juliencarnot:
@robguthrie `gulp dev` still stalls after the compile-fast step on my install, but `gulp compile` works...

gdpelican:
Hmmmm... I ddon't think gulp dev is stalling, rather it's running a watch
which will update your files automatically when you change them. I wouldn't
expect `gulp dev` to complete unless you manually kill it.

juliencarnot:
That makes a lot of sense! Thanks for the explanation!

robguthrie:
ah hah! now we all understand! :)

PR #4421, opened 2017-11-19
allow middle/scrollwheel click and right click open in new tab on sidebar and dropdown buttons that open a page

### LINKED ISSUE #3943: Make it possible to right-click open new tab in sidebar buttons
A user has reported that they can't right-click to choose 'open in new tab' on group names in the sidebar, and I've confirmed it's so.

juliencarnot:
In sidebar, navbar and user-dropdown, replaces href attributes with ng-href because lmo-href alone doesn't allow middle/scrollwheel click and right click open.
Buttons opening modals and logout are untouched, middle/scrollwheel click has no effect on them, no open in new tab if right clicked. Cleaned-up the commits and merged with current master.

This should fix #3943

### DIFF
--- a/client/angular/components/navbar/navbar.haml
+++ b/client/angular/components/navbar/navbar.haml
@@ -6,7 +6,7 @@
           %i.mdi.mdi-menu
 
       .navbar__middle.lmo-flex.lmo-flex__horizontal-center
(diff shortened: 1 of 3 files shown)


gdpelican:
This is a nice solution, thanks @juliencarnot (and again sorry for the delay)

Issue #3977, opened 2017-03-10
'cover photo' feature is in need of some love and care...- no?


zblace:
There is almost no control of display and positioning of cover pics!  :-(

Also it would be awesome if they do not post-process it and distort otherwise clear .png graphics uploaded into that blurry mess we see now on display :-/

gdpelican:
PRs are more than welcome in this area, but it's not on our immediate roadmap (we're focusing on the experience _outside_ of a Loomio group at the moment)

zblace:
What is experience outside of the group in practice?

gdpelican:
Sorry for not responding to this earlier, @zblace .

We've been working on a whole bunch of new decision types (finding a meeting time, single/multiple choice polling, dot voting, and yes/no questions), all of which will be able to be created outside of a group (because we've found it's too hard to get started if you need to make a group, start a discussion, and _then_ make a decision. We're decisionmaking software, you should be able to make a decision instantly with this thing.)

The other piece of this is a bunch of work around sharing polls in places where users are... putting a shareable link into an email chain, posting into slack and facebook, embedding polls in webpages using JS, etc. You can check out `loomio.org/p/new` to try out the new flows.

In response to the cover photo thing, I agree that it's not ideal, and would love for someone to step up and either write these features in (as you note, users expect _way_ more control over the media you put up on your pages than when we first wrote this a couple years ago), or throw in a library that does this better than we've coded it.... but for now I have to leave it because we're stretched real thin on dev resourcing.

zblace:
​Thank you for elaborating on priorities.
I agree that is good choice and direction.
Looking forwrad to ​see you move from only
in-site decision making into sherable applets/services.

best Z

Issue #3790, opened 2016-10-17
Resolution for group cover photos


rdbartlett:
When you visit the group cover upload modal, you see the recommended size is 1400x320px. If you look in `app/models/group.rb` you'll see we call this size "largedesktop". However, when you visit the group page, we're serving the "desktop" size, which is 970x200px. 

This results in cover images having poor resolution on high density displays, and being cropped unexpectedly (the specified aspect ratio is 0.2286, but Loomio is serving images at 0.2062).

At least, the prompt should provide accurate information. 

Ideally, we would serve high resolution images to high resolution displays, and smaller files to small screens. (E.g. my 3 year old laptop display is 2600px wide.)

Does it make sense to switch out `cover_url_desktop` for `cover_url_large_desktop` as a first step, and then look at a more intelligent solution further down the line?

If there are groups with cover photos that lack a large_desktop size, I think it is easy to [generate the missing sizes](https://github.com/thoughtbot/paperclip/wiki/Thumbnail-Generation#generatingregenerating-your-thumbnails).

What do you think is the best way forward folks? I'd like to resolve this before we start telling everyone to upload new cover photos.

gdpelican:
Hm.

- I'm not worried about generating sizes
- I'm very worried about making people download big images on mobile (the existing image is already waay too big for that.)
- I have a hunch that our current solution was a compromise for that
- It's impossible to inline @media queries, which means we can't dynamically change the image being displayed based on window sizing that way.
- I _think_ we want this: https://css-tricks.com/responsive-images-youre-just-changing-resolutions-use-srcset/, because we can serialize out all of the possible images (they're just urls, not actual image data, so it's easy), throw them into a srcset, and let the browser decide what size is best.

However, I'm a pretty unenthusiastic about doing that work right this second; how important do you feel it is?

gdpelican:
Also, srcset _just_ got released the IE bleeding edge, so it won't work for IE10; sigh.

rdbartlett:
I registered this ticket after setting up a real group and feeling embarrassed about the lofi image, with the wrong aspect ratio. 

I'm keen for us to make some improvement here, as we're putting a lot more emphasis on cover photos as part of the activation work.

Regarding browser compatibility, if you use `src` and `srcset`, I'm guessing IE will fall back to whatever you define in src though right?

robguthrie:
I think we can provide the best resolution image for the viewport width without using unsupported html features.

[...]

PR #4430, opened 2017-11-23
keep aspect ratio of group banner and show all of it in all screen sizes

juliencarnot:
uses background-size: contain and 22.85% padding to maintain aspect ratio and show all of the banner image on all screens. 

This should fix aspect ratio problems in #3790 and #3977 .

Tested on Firefox and Chromium.

See results with following banner image:
default_group_cover

Before:
group page mobile scroll top master
After:
group page mobile scroll top pr

Before:
group page scroll top master
After:
group page scroll top pr

Before:

group page scroll top no sidebar master

After:

group page scroll top no sidebar pr

### DIFF
--- a/angular/core/components/explore_page/explore_page.scss
+++ b/angular/core/components/explore_page/explore_page.scss
@@ -45,8 +45,9 @@
 }
 
 .explore-page__group-cover {
-  height: 100px;
-  background-size: cover;
+  height: 85px;
+  background-size: contain;
+  background-repeat: no-repeat
 }
 
 .explore-page__group-details {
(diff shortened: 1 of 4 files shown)


robguthrie:
hah! Amazing!!

Thanks for the care and support @juliencarnot. Work like this really helps to make things smooth and nice (both in the UX for users and the feeling for the work for me).

robguthrie:
Has this been tested on IE 11? If you don't have a way to do that let me know.

juliencarnot:
I don't have a Windows box at hand to test it. From what I've read, the css technique I used should be widely supported, but it would be great if you can check that!

PR #4431, opened 2017-11-24
Revert "keep aspect ratio of group banner and show all of it in all screen sizes"

robguthrie:
Reverts loomio/loomio#4430

We're seeing a couple of minor visual problems:
screen shot 2017-11-24 at 2 47 00 pm

### DIFF
--- a/angular/core/components/explore_page/explore_page.scss
+++ b/angular/core/components/explore_page/explore_page.scss
@@ -45,9 +45,8 @@
 }
 
 .explore-page__group-cover {
-  height: 85px;
-  background-size: contain;
-  background-repeat: no-repeat
+  height: 100px;
+  background-size: cover;
 }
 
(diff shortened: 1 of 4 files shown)


robguthrie:
@juliencarnot FYI. Seems like there is something up here.

juliencarnot:
Hum, sorry about that, it's strange. I ran into the problem padding-top on pages with no banners before, but fixed it by setting the padding only when there is a group cover to display (see addClass in root_controller.coffee) to lmo-group-background and not to lmo-main-background.

What browser were you using? Did you force reload in the browser, no local cached elements? I can't reproduce it in either Firefox and Chromium, and don't get how it could show on the recent threads page:
capture du 2017-11-24 08-33-31

I also tried with a user belonging to only 1 group, but it displays well for me too...

PR #4432, opened 2017-11-25
adds language support for pt_BR, it, cs

juliencarnot:
These languages are all being actively translated and have more than 33% completion on https://www.transifex.com/rdbartlett/loomio-1/languages/. It will help translators to improve translations in context and will be a good incentive to continue the efforts.


gdpelican:
Perfect, thank you!

Issue #4425, opened 2017-11-20
Pasting image into comment makes it show up twice


mhjb:
![image](https://user-images.githubusercontent.com/1380991/33042140-ac676560-cea5-11e7-86df-ecc55c2e927e.png)

BenDawes:
https://github.com/loomio/loomio/pull/4438

mhjb:
I think this is sorted now

PR #4438, opened 2017-11-28
Fix Duplicate Attachment on Paste

BenDawes:
Thanks to lines 8/9:
```

    $scope.$watch 'files', ->
      $scope.upload($scope.files) 

```
We don't need to call upload, since setting $scope.files will trigger upload anyway.


BenDawes:
Raised in https://github.com/loomio/loomio/issues/4425

gdpelican:
Great catch; I'd actually just nabbed this in the upcoming documents branch, but more than happy to merge this now. Thanks!

PR #4440, opened 2017-11-29
Attempt to fix discussion deserialization errors

gdpelican:


gdpelican:
Think this will do the trick @robguthrie ?

app/models/discussion.rb
```diff
@@ -95,10 +95,14 @@ def thread_item_created!
...
     update_sequence_info!
-    RecalculateReadItemsCountsJob.perform_later(discussion)
-    true
+    update_reader_items_count!
   end
 
+  def update_reader_items_count!
+    discussion_readers.map(&:sync_read_items_count!)
+  end
+  handle_asynchronously :update_reader_items_count!, priority: 0
```
robguthrie: default priority is 10, maybe 5? emails should be 100 or so.

app/models/discussion_reader.rb
```diff
@@ -67,6 +67,10 @@ def discussion_reader_volume
     self.class.volumes.invert[self[:volume]]
   end
 
+  def sync_read_items_count!
+    update_attribute(:read_items_count, calculate_read_items_count)
```
robguthrie: I can't see any place in the server side where we use this anymore.

The client calculates it's own read items count pretty quickly anyway. I think we can drop the job ( and column later, with the other columns I'm yet to drop).

app/models/discussion_reader.rb
```diff
@@ -88,15 +92,12 @@ def unread_ranges
...
 
   def read_ranges_string
-    if self[:read_ranges_string] == nil
+    self[:read_ranges_string] ||= begin
       if last_read_sequence_id == 0
         ""
       else
-        first = (discussion.first_sequence_id == 0) ? 1 : discussion.first_sequence_id
-        "#{first}-#{self.last_read_sequence_id}"
+        "#{[discussion.first_sequence_id, 1].max}-#{last_read_sequence_id}"
```
robguthrie: yep, looks correct to me.

robguthrie:
thank you!

PR #4449, opened 2017-12-05
Lazy create created_events

robguthrie:


app/services/create_missing_event_service.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    poll.stances.order(:created_at).each(&:created_event)
+    # ensure there is a poll closed event?
+    poll.outcomes.order(:created_at).each(&:created_event)
+
+    poll.discussion.items.last.send(:reorder)
+
+    # TODO ensure poll closed and expired events are sorted
+
+    Events::StanceCreated.find(res.ids.first.to_i)
```
gdpelican: Don't need this!

PR #4452, opened 2017-12-07
help users understand we convert timezones correctly

robguthrie:
@gdpelican I think we need a translations for timezone/place_name, so we don't just shove english timezones everywhere. 

Any idea on the best way to do this?

Seems similar to translations for language name too


gdpelican:
Hm, I imagine we want to allow translations in different languages for each time zone name; English speakers will want 'Prague' whereas Czech / German / etc speakers will want 'Praha' for example, or like Copenhagen vs København

I guess just use the 'canonical' time zone name as a key, and allow for different language translations the same way we do for other string?

robguthrie:
hey I don't think this was ready for merge? We need to translate the timezone names

robguthrie:
never mind/all good I've fixed it for now.

PR #4454, opened 2017-12-08
A first sniff at ngVue

gdpelican:
There's some things not to like (primarily for me, not being able to take advantage of *.vue files just yet, and needing to separate out css files from coffee), but this is a working implementation where all of our %timeago directives are powered by Vue, and that's pretty neat.

### DIFF
--- a/angular/boot/dependencies.coffee
+++ b/angular/boot/dependencies.coffee
@@ -21,5 +21,6 @@ angular.module('loomioApp', [
   'angulartics.google.tagmanager',
   'vcRecaptcha',
   'ngAnimate',
-  'angular-sortable-view'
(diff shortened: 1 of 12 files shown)


robguthrie:
you sly dog!!!!

gdpelican:
I really only want Vue if I can get the single-file syntax, so I'll probably try to make that happen with this soon.

Just... like, how cool would this be?

```
<script lang="coffee">
  Vue.component =
    props: ['timestamp']
    created: -> # do some stuff
</script>

<template lang="haml">
  .vue-timeago{title: "{ this.title }"}
</template>

<style lang="scss">
  .vue-timeago {
     background-color: $background-color;
  }
</style>
```

PR #4457, opened 2017-12-11
just in time parent event lookup

robguthrie:


app/models/concerns/events/discussion_parent.rb
```diff
@@ -0,0 +1,7 @@
+module Events::DiscussionParent
+  include Events::LookupMissingParent
+
+  def self.lookup_parent(eventable)
+    eventable.discussion&.created_at
```
gdpelican: Should this be returning an event instead of a date time?

app/models/concerns/events/lookup_missing_parent.rb
```diff
@@ -0,0 +1,9 @@
+module Events::LookupMissingParent
+  def self.lookup_parent(comment)
```
gdpelican: Looking at this in PR form, it seems like a lot of concepts that get placed in a lot of places. I wonder if we want to just write one slightly gnarly method that we can remove when we don't want it anymore? This would be similar to, say, our `polymorphic_url` method, which takes a model and returns a url for it, no matter what it is. The internals are a little hairy, but it works and doesn't need changing.

```
# event.rb, no need to touch other files
def self.lookup_parent(eventable)
  case eventable
  when Comment then (eventable.first_ancestor || eventable.discussion).created_event
  when Discussion then eventable.discussion.created_event
  when Poll, Stance, Outcome then eventable.poll.created_event
  when PaperTrail::Version then lookup_parent(eventable.item)
end
```

PR #4463, opened 2017-12-12
use polyfill for closest on ie11

robguthrie:
error:
https://errbit.loomio.io/apps/57215524657272000e000001/problems/5a2f83b929e6e20007dce940

polyfill source:
https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

converted via js2coffee


gdpelican:
Looks like this would be affecting mentio on IE11

robguthrie:
yes, that's where the error is coming from

PR #4464, opened 2017-12-12
fix for parentEvent undefined exception

robguthrie:


gdpelican:
I think this points to needing to call the discussions#show route on
loading this directive, so we know there's a created evert (the server will
soft create if one doesn't exist)

I think this will hide the activity card when no created event exists,
which fixes the error but not the experience?

robguthrie:
I was under the impression we always did call discussions#show (ie: full discussion serializer) and this would just wait until that response had come back before showing it.

PR #4474, opened 2017-12-13
Calculate full name

gdpelican:


robguthrie:
hahah. oh man. good times

PR #4475, opened 2017-12-13
Close threads

gdpelican:


gdpelican:
Okay @robguthrie this is ready for a look

robguthrie:
image

The baseline of Open threads 0 Closed is off.

I would recommend swapping the search icon and the closed threads labels anyway:
image

And I would also suggest just one font size for everything in the title bar.

Maybe don't display "0 closed" if there is 0 closed.

Those closed badge and the model are excellent. Reopen worked well. Search worked well

PR #4476, opened 2017-12-13
Fix event removed error

gdpelican:


angular/core/services/ahoy_service.coffee
```diff
@@ -1,7 +1,8 @@
-angular.module('loomioApp').factory 'AhoyService', ($rootScope, $window) ->
+angular.module('loomioApp').factory 'AhoyService', ($rootScope, $window, AppConfig) ->
   new class AhoyService
     init: ->
       return unless ahoy?
+      return if AppConfig.environment == 'development'
```
robguthrie: I'm a little concerned about this, it means dev behaves different to prod. What are the reasons for this change?

gdpelican: Ahoy makes it really hard for me to track the network panel, as most of the requests in there are event requests that don't have anything that I'm interested in. Happy to put it in an ENV that I can turn off.

robguthrie:
better!

Issue #4445, opened 2017-12-01
Screwy numbers on Check poll


rdbartlett:
"15 out of *blank*" and "16/28": which one to believe?

![image](https://user-images.githubusercontent.com/970124/33484757-d37adcda-d68a-11e7-8056-626b0f6eaab5.png)

gdpelican:
Thanks. This seems like something a contributor could do a really good job with if we don't get to it first.

BenDawes:
I have a fix for the missing value after 'out of' @gdpelican . However I'm not sure what we want the correct value to be. It's my belief that in the above example, 16 people HAVE responded, and 15 of them HAVE responded with a Tick (the other 1 with a Cross).

The current (buggy) code for the string after 'out of', if it were correct, would in the above example put '15 out of 28'. However, is there a strong case that it should be '15 out of 16', since 15 out of 28 might give the erroneous impression that that is the final result.

Shall we go with (X out of Y) where X is number of Tick responses and Y is total responses so far, or the current idea of Y is total members involved in the poll?

gdpelican:
Hm, it's a good question... the original intent for this poll was to have _only_ a yes response (like 'I did it'), but we got plenty of feedback that people wanted to be able to respond in the negative ('sorry, but I can't make it')

My gut says that we want "<number of participants> out of <number of poll members>", to match the percentage below, and leave the chart visual to convey how many yes / no responses there are (so, 16 / 28 in the example above)

BenDawes:
So suppose there are 30 members in my group, 5 respond Yes, 10 respond No, and 15 haven't responded yet. What should it display?

1) 5 out of 30 (50% of members have stated their position, 15/30). 
2) 15 out of 30 (50% of members have stated their position, 15/30). 
3) 5 out of 15 (50% of members have stated their position, 15/30).

and should the background of the results (the progress bar) be:

a) 1/6 complete ('Yes'-votes over total members asked)
b) 1/3 complete ('Yes'-votes over total responses received)
c) 1/2 complete (Responses received over members asked)

?

gdpelican:
In that example, I'd say #2. The background should already be set up to have 3 'bars' in it, one for 'yes', one for 'no', and one for 'undecided' (although at present 'no' and 'undecided' are fairly similar shades of grey)

BenDawes:
Okay, cool, forgot about the 3 bars.

[...]

PR #4488, opened 2018-01-03
Fix Check count results page (#4445)

BenDawes:
Root cause of bug in #4445: membersCount() returns an integer for the total members, whereas in the fallback of goal() it was treated as if it returned a list (of members)

Per discussion in #4445, this PR also updates the numerator of the check poll results to be the total responses, not the Yes responses.

### DIFF
--- a/angular/core/components/poll/count/chart_panel/poll_count_chart_panel.haml
+++ b/angular/core/components/poll/count/chart_panel/poll_count_chart_panel.haml
@@ -6,5 +6,5 @@
       .poll-count-chart-panel__no{ng-style: "{'flex-basis': percentComplete(1), 'background-color': colors[1]}"}
       .poll-count-chart-panel__yes{ng-style: "{'flex-basis': percentComplete(0), 'background-color': colors[0]}"}
     .poll-count-chart-panel__data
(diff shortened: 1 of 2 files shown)


angular/core/components/poll/count/chart_panel/poll_count_chart_panel.haml
```diff
@@ -6,5 +6,5 @@
       .poll-count-chart-panel__no{ng-style: "{'flex-basis': percentComplete(1), 'background-color': colors[1]}"}
       .poll-count-chart-panel__yes{ng-style: "{'flex-basis': percentComplete(0), 'background-color': colors[0]}"}
     .poll-count-chart-panel__data
-      .poll-count-chart-panel__numerator {{poll.stanceCounts[0]}}
+      .poll-count-chart-panel__numerator {{poll.stanceCounts[0] + poll.stanceCounts[1]}}
```
gdpelican: I believe this could be poll.stancesCount?

BenDawes: Happy to change it to that, yes, wasn't sure if we wanted to incorporate any future stances we may add such as acknowledged but undecided; will change now.

gdpelican:
Thanks @BenDawes !

PR #4489, opened 2018-01-04
fix inconsistent naming: use group instead of groupName for all templates

### LINKED ISSUE #4487: Translation inconsistencies
Hi, I've found some inconsistencies and "bugs" in the translation files (notably the french ones)
In particular the `groupName` variable, which seems to have changed to `group` on some places in the code but have not been updated in translation files, which leaded to missing group name in some messages.

This issue is mostly about fixing this bug but also querying about "unifying" the variable names used in translation messages, as the `groupName` variable seems to be still used in some places despite being changed to `group` in other places.

ariasuni:
fix #4487

### DIFF
--- a/angular/core/components/move_thread_form/move_thread_form.coffee
+++ b/angular/core/components/move_thread_form/move_thread_form.coffee
@@ -19,6 +19,6 @@ angular.module('loomioApp').factory 'MoveThreadForm', ->
 
     $scope.moveThread = ->
       if $scope.discussion.private and $scope.targetGroup.privacyIsOpen()
-        $scope.submit() if confirm($translate.instant('move_thread_form.confirm_change_to_private_thread', groupName: $scope.targetGroup.name))
+        $scope.submit() if confirm($translate.instant('move_thread_form.confirm_change_to_private_thread', group: $scope.targetGroup.name))
       else
         $scope.submit()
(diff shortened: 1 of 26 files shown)


gdpelican:
Thanks for the PR @ariasuni.

Could you go through and ensure that all of the translation files have this
update, so it doesn't break other languages?

ariasuni:
I wanted to do that in the first place, but then it will be overwritten when updating your strings from your translation system, right?

robguthrie:
actually, we can push changes back into transifex. so a pr of that is welcome.

ariasuni:
So I edited my PR, does it look good to you?

ariasuni:
It’s been a month, did anyone have a look at it?

robguthrie:
Hi @ariasuni. Thanks for taking the time to try to improve Loomio, and sorry for the slow reply.

As you can see there is now a significant merge conflict with your branch so it is unlikely we can merge it without some significant time investment.

We were hesitant to merge the branch before the merge conflict happened because it would require a lot of manual uploading of files to transifex, our translation UI service - being a manual process this is time consuming and error prone.. We felt that renaming variables did not provide enough value to us to warrant the extra time and risk of errors, so we did not proceed with the support work required to accept your pull request.

Sorry if this is a disappointing experience for you, it is for us too, actually. We'd love to be able to just click merge on this, but we couldn't justify the work required in this case.

I hope that you may feel compelled to contribute again, if you'd like to tell me a bit about your level of dev experience I can see what tasks there are that might suit you and provide the project with significant value.

Warm regards,
Robert

ariasuni:
No problem, that sort of things happens ^^

Issue #4391, opened 2017-11-05
Loomio believes I reside in Athens though I've never specified my location


wfdd:
![location mov](https://user-images.githubusercontent.com/8489551/32414500-483841fe-c220-11e7-9186-24cbefaaf76b.gif)

gdpelican:
Yes, we detect location based on browser information that's sent over on initial request (and have done so for a long time). Location can be overridden with anything you like if you'd like this information to not be visible.

wfdd:
I think it's actually some consolation that Loomio believes I live in a place I've not visited in nearly two decades if the developers here think it acceptable to leak people's location.

robguthrie:
Hrrmm I wonder what others think.

I'm open to not revealing user location without permission... seems kind of invasive in some situations.

The location is guessed from IP address and it sucks pretty bad. Which made me think it would be "ok"

We _could_ ask the user's browser for their location when they click the location field, using html5 geolocation: https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/Using_geolocation

So one scenario is that we don't reveal the guessed location to users (we use it for basic analytics). We offer to geolocate users from the profile page, and allow users to override the result string.

juliencarnot:
I'd say reset/leave it blank if the user did not choose to disclose a location in the text input of the profile page. Using html5 only would give results that are ranging from unusable if you're on a desktop with an ISP not disclosing ip address locations but a network node 300km away, or creeping exact with Chrome/Google maps' streetview SSID geolocation or from a mobile.

I noticed that in a group that's supposed to be local, virtually all users appear to be from cities far away, sometimes in different European countries (of course, they're not). That's "ok" from a privacy point of view, but useless for users and for the platform. If you really need this data for analytics, I hope you don't trust the reports... 

In my view,  potential geolocation uses for Loomio, such as finding a nearby group containing a keyword (this would imply that groups can have a location), or setting up a subgroup for geographical region should remain on a user opt-in base, and will require storing structured geodata and support approximation for privacy, so I'd leave that out of the scope until there's a real case for it.

juliencarnot:
By the way, I just noticed these info on my account in the JSON bootData: `"location":"La Pesse, France","city":"Frankfurt am Main","region":"Hesse","country":"Germany"`

[...]

PR #4497, opened 2018-01-16
displays location and lastSeen on group members page and fixes minor glitches

juliencarnot:
Using Loomio in a large group that doesn't meet IRL, I find that it would be useful to know where members are (if they choose to enter a location, as discussed in #4391) and if they've been active on loomio recently (this can be useful for coordinators to gauge their group engagement too, see [this comment](https://www.loomio.org/d/laZLgn2I/time-poll/13)).

image

This PR also hides inaccurate detected locations from GUI per #4391, adds page title to group page header and removes unecessary parenthesis from users panel in locales. Sorry for this all-in-one commit :-(


gdpelican:
Cool, this all looks good to me, thanks @juliencarnot !

PR #4507, opened 2018-01-21
hides domain from email address displayed in pending invitations for better privacy

### LINKED ISSUE #593: Invited users' email addresses are exposed to other group members
Given a public group exists
And all members can invite other users to the group
Then all group members can see the invited users' email addresses

This is okay for medium-sized groups, but will become a privacy issue when we scale. We'll need to think of an alternative here. Perhaps completely hiding the invited users list from everyone except admins? It's quite an easy solution that might just do the trick.

juliencarnot:
fixes #593, by displaying only the email address prefix so that other members have to guess the domain instead of seeing the full email address, in order to protect invited users from spam.


robguthrie:
Hi Julien, thanks for your efforts.

We've had no complaints from users about this behaviour.

This patch will still send full email addresses across the wire to the client, so it's not really effective if a spammer wanted to get the addresses. (We'd need to change the serializer to keep the address within the server) 

Even if the above were not true, a large proportion of addresses can be guessed with the approach of hiding the domain.

So, I'm sorry to say that I don't think I can approve this suggested change as it feels like security though a tiny bit of obscurity, for a problem we're not hearing about from the users.

I reckon we need to think about what the right solution is.. It could be that we ask for a name when inviting people, then do not show the email at all.

PR #4515, opened 2018-02-05
Announcements

gdpelican:


app/models/announcement.rb
```diff
@@ -41,7 +41,7 @@ def announce_and_invite!
   end
 
   def users_to_announce
-    @users_to_announce ||= User.where(id: announcees.pluck(&:user_ids).flatten.uniq)
+    @users_to_announce ||= User.where(id: announcees.pluck(:user_ids).flatten.uniq)
```
gdpelican: This bug was a reallll ripper @robguthrie

gdpelican:
Moved to the `announcements` branch

PR #4524, opened 2018-02-12
Actioncable

robguthrie:


client/shared/helpers/user.coffee
```diff
@@ -31,40 +31,32 @@ module.exports =
...
+    _.each Session.user().groups(), (group) ->
+      AppConfig.cable.subscriptions.create { channel: "GroupChannel", group_id: group.id },
+        received: (data) ->
+          console.log("received ", data)
+          if data.memo?
+            switch data.memo.kind
+              when 'comment_destroyed'
+                if comment = Records.comments.find(memo.data.comment_id)
+                  comment.destroy()
+              when 'comment_updated'
```
robguthrie: I think we just want to create a comment_updated event and publish it the same way as the other events.

client/shared/helpers/user.coffee
```diff
@@ -31,40 +31,32 @@ module.exports =
...
+                Records.comments.import(memo.data.comment)
+                Records.import(memo.data)
+              when 'comment_unliked'
```
robguthrie: not used

app/services/message_channel_service.rb
```diff
@@ -1,25 +1,5 @@
...
-    else
-      PrivatePub.publish_to(to.message_channel, data)
-    end
-  end
-
-  def self.ensure_valid_channel(model)
-    return unless model
-    raise UnknownChannelError.new unless model.respond_to?(:message_channel)
-    Rails.application.secrets.faye_url.present?
+    ActionCable.server.broadcast to.message_channel, data if to.respond_to?(:message_channel)
```
robguthrie: we might need to pluralise this, to `message_channels.each do ` if we follow that things have a groups method.

PR #4554, opened 2018-02-22
Poll exporter

Space-Ed:


app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,121 @@
...
+
+
+  def to_csv (opts={})
+    CSV.generate do |csv|
+
+        meta_table.keys.each {|key|
+          csv<<[key.to_s.humanize, meta_table[key]]
+        }
+        csv << []
+        stance_matrix.each {|row|csv<<row}
```
gdpelican: I'd have a look to see if you could do `csv.append` or `csv.push` or something instead of putting csv inside the block each time. This would clean up a lot if it looked like

```
CSV.generate do |csv|
  csv.append meta_table.map { |key, value| [t(:"poll_exporter.#{key}"), value] }
  csv.append nil # add an empty row
  csv.append stance_matrix
end
```

app/views/poll_mailer/common/_summary.html.haml
```diff
@@ -1,8 +1,13 @@
 .poll-mailer-common-summary
-  - if @info.poll.discussion
+  - if @info.poll.discussion && @info.action_name != :export
     = render "poll_mailer/common/discussion_info", discussion: @info.poll.discussion
 
   %h1= link_to @info.poll.title, @info.links[:target], class: 'poll-mailer__poll-title'
+  -if @info.action_name == :export
+    %p
+    Also available in
```
gdpelican: Let's translate this as well, `export.poll.also_available` maybe

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,121 @@
...
+      outcome_statement: outcome&.statement,
+      poll_url: poll_url(@poll)
+    }.compact
+  end
+
+  def stance_matrix
+    rows = []
+    rows << ["Participant", @poll.poll_options.map(&:display_name), "Reason"].flatten
+
+    totals = [0]*@poll.poll_options.size
```
gdpelican: No totals anymore looks like

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  def poll_info (current_user)
+     PollEmailInfo.new(recipient: current_user, event: @poll.created_event, action_name: :export)
+  end
+
+  def meta_table
+    outcome = @poll.current_outcome
+    engagement = nil
+    I18n.with_locale(:en) {
```
gdpelican: No need for this with_locale; all of the requests are wrapped in a with_locale call for the current user, so this is just forcing :en on folks who may not want it.

```
I18n.t(:"export.poll.percent_voted",
  num: @poll.stances_count,
  denom: @poll.members.count,
  percent: "#{100 * @poll.stances_count / @poll.members.count}%"
)
```

I'd also be tempted to write an `engagement` method on poll:
```
# poll.rb
def engagement
  @engagement ||= "#{100 * stances_count / members.count}%"
end
```

app/controllers/dev/main_controller.rb
```diff
@@ -161,7 +168,7 @@ def setup_logged_out_group_member
   def setup_logged_out_member_of_multiple_groups
     patrick
     create_group
-    create_another_group
```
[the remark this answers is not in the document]
Space-Ed: I'm not sure exactly, I'll put it back

app/controllers/polls_controller.rb
```diff
@@ -1,5 +1,19 @@
 class PollsController < ApplicationController
   include UsesMetadata
+  include LoadAndAuthorize
+  include EmailHelper
+
+  helper :email
+
+  def export
```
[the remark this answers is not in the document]
Space-Ed: Presuming that they will stay well aligned to the same exporter interface, I'd perhaps want to think about what the future developments may be before making this a mixin.

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,121 @@
...
+      rows << []
+    end
+
+    options
+  end
+
+  def stance_list
+    rows = []
+    rows << ["created_at", "author ", "is latest ", @poll.poll_options.map(&:display_name)].flatten
+    totals = [0]*@poll.poll_options.size
```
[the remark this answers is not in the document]
Space-Ed: Ah yes agreed this style is less likely to duplicate object references.

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    ## for each participant show the
+    @poll.stances.sort_by(&:created_at).each do |stance|
+      user = stance.participant
+      row = [stance.created_at, user.name, stance.latest]
+
+      @poll.poll_options.each do |poll_option|
+        # find the value for this stance_choice for poll option in this stance
+        stance_choice = stance.stance_choices.find {|choice| choice.poll_option.id == poll_option.id}
```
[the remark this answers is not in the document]
Space-Ed: Noted, subtle but important distinction!

app/controllers/polls_controller.rb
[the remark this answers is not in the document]
gdpelican: Yep cool happy to wait on this one, perhaps when/if we write a DiscussionExporter

The third time you write something usually feels like the right time to abstract it probly :)

app/extras/poll_exporter.rb
Space-Ed: csv seems to not have an append method, this was a workaround, hoping for a better way

config/locales/server.en.yml
```diff
@@ -29,6 +29,24 @@ en:
       cannot_add_options: Cannot add options to this poll
       cannot_remove_options: Cannot remove options from this poll
       must_have_options: Must have at least one poll option
+    export:
+      percent_voted: “%{num}/%{denom} (%{percent}) members have stated their position”
+      title: "title"
+      author: "author"
+      created_at: "created_at"
```
gdpelican: I think we want the translation strings in here to be as human-readable as possible (ie, 'Created at', rather than 'created_at'), since translators will be looking at this and may be confused by the underscore formatting (What's "outcome_created_at" in Spanish?).
```
# en.yml
poll:
  export:
    created_at: "Created at"
```
so that someone can translate to
```
# es.yml
poll:
  export:
    created_at: "Empezado"
```

My ideal thing here would be
- Create a hash where the keys are underscored symbols
```
{ poll_url: "loomio.org/p/key" }
```
- Translate that in the html template and as we're passing it into the csv
```
- keys.each do |key|
  %h2= t(:"poll.export.#{key}")
```
```
hash.each do |key, value|
  csv << [I18n.t(:"poll.export.#{key}"), value]
end
```

app/models/ability/poll.rb
```diff
@@ -48,5 +48,10 @@ def initialize(user)
     can :close, ::Poll do |poll|
       poll.active? && (user_is_author_of?(poll) || user_is_admin_of?(poll.group_id))
     end
+
+    can :export, ::Poll do |poll|
+      #TODO: review priveledges before merge 
+      can?(:show, poll)
```
gdpelican: I'm feeling like we want to limit this to admins and open it up if people complain. We can simply add `:export` to the list of permissions above with :update, :share, :remind, etc

```
can [:update, :share, :remind, :export], ::Poll do
```

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,118 @@
...
+      end
+      rows << []
+    end
+
+    options
+  end
+
+  def stance_list
+    rows = []
+    rows << ["created_at", "author ", "is_latest", @poll.poll_options.map(&:display_name)].flatten.map{ |name_label| label(name_label) }
```
gdpelican: display_name is correct here, and will be translated properly from when the poll is created.

(Ie, if someone makes a poll and they speak spanish, the poll option's name will be stored as 'Acuerdo' rather than 'Agree')

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  def user_reasons
+    options = {}
+
+    @poll.poll_options.each do |option|
+      rows = []
+      options[option.display_name + "(#{option.total_score})"] = rows
+
+      if option.stances.empty?
+        rows << ["No stances for this poll option"]
```
gdpelican: We'll want to grab this translation too.

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,118 @@
+class PollExporter
+
+  include Routing
+
+  def initialize (poll)
+    @poll = poll
+  end
+
+  def file_name
+    "#{@poll.title.parameterize}-export.csv"
```
gdpelican:
```
"#{@poll.title.parameterize}-#{I18n.t(:"common.action.export").downcase}"
```
or, if you're feeling cheeky,
```
[@poll.title, I18n.t(:"common.action.export")].join('-').parameterize
```

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,118 @@
...
+      outcome_author: outcome&.author,
+      outcome_created_at: outcome&.created_at,
+      outcome_statement: outcome&.statement,
+      poll_url: poll_url(@poll)
+    }.compact
+  end
+
+  def stance_matrix
+    rows = []
+    rows << [label("Participant"), @poll.poll_options.map(&:display_name), label("Reason")].flatten
```
gdpelican: Yep so I think this would be
```
label(:participant)
```
which would go to
```
I18n.t("poll.export.participant")
```
which would get translated to "Participant" (or "Partícipe", etc.)

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    totals = (1..@poll.poll_options.size).map { 0 }
+
+    ## for each participant show the
```
gdpelican: show the ?? :)

app/extras/poll_exporter.rb
```diff
@@ -0,0 +1,118 @@
...
+      outcome_statement: outcome&.statement,
+      poll_url: poll_url(@poll)
+    }.compact
+  end
+
+  def stance_matrix
+    rows = []
+    rows << [label("Participant"), @poll.poll_options.map(&:display_name), label("Reason")].flatten
+
+    totals = (1..@poll.poll_options.size).map { 0 }
```
gdpelican: We don't use totals in this method, I don't think?

app/models/ability/poll.rb
gdpelican: Part of my reasoning for this is that :show is _very_ relaxed for some polls (since we have polls that anyone can participate in), meaning basically anyone could download the result of the poll.

I'm not really opposed to this (it is accessible, after all), but the only people who have asked for this are people who are coordinating a group, so I want to give this to them first.

app/extras/poll_exporter.rb
Space-Ed: This method is never used, so I'll remove it

app/models/ability/poll.rb
Space-Ed: ok, i'll make it poll authors and admins for now

config/locales/server.en.yml
Space-Ed: I'm not too sure about this, I think I would like to provide the translation in the common table. Then I know what goes out will have translation and use the same key. 

I guess what you're also saying is that translation should be only done by replacing strings with I18n.t calls, not having intervening helper functions.

gdpelican:
Alright, I'm ready to merge if these tests pass. Here's some minor cleanup I did FYI @Space-Ed: 

- Account for anonymous polls by using `stance.participant_for_client` rather than `stance.participant` (the former anonymizes voters when the poll is anonymous)
- Account for standalone polls by add a nil check on `poll.group.full_name`
- Account for polls with no members by ensuring 'total' is > 0 (it's a tough case to imagine IRL, but I accidentally wrote a failing test for it)
- Taking out the ' ' in method signatures (`def method(param)` is idiomatic ruby, as opposed to `def method (param)`
- Move `export_matrix` to the poll_mailer folder (where the other views are), and rename to `stance_table`

This is a good-looking piece of work, thanks! 👍 Looking forward to hearing some user feedback on it.

PR #4557, opened 2018-02-27
add skylight gem so we can see perf

robguthrie:
We've been granted free [Skylight](https://www.skylight.io/) usage. Hooray.


robguthrie:
FYI, the env is setup on loomio-production, so should be ready to go. (gonna deploy first thing tomorrow)

PR #4574, opened 2018-03-09
ensure that comment replies are pointing to the right comment.

robguthrie:
There's a bug that when you reply to a nested comment, the comment's parent is actually the "surface" comment, this fixes that.

### DIFF
--- a/client/angular/components/thread_page/event_children/event_children.coffee
+++ b/client/angular/components/thread_page/event_children/event_children.coffee
@@ -15,7 +15,7 @@ angular.module('loomioApp').directive 'eventChildren', ->
       initialSequenceId: $scope.parentEventWindow.initialSequenceId
       per:               $scope.parentEventWindow.per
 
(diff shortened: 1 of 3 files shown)


client/angular/components/thread_page/event_children/event_children.coffee
```diff
@@ -15,7 +15,7 @@ angular.module('loomioApp').directive 'eventChildren', ->
       initialSequenceId: $scope.parentEventWindow.initialSequenceId
       per:               $scope.parentEventWindow.per
 
-    EventBus.listen $scope, 'replyToEvent', (e, event) ->
+    EventBus.listen $scope, 'replyToEvent', (e, event, comment) ->
```
gdpelican: Could just drop this since js doesn't enforce arg length

gdpelican:
Waaay simpler than I thought it'd be, good stuff.

PR #4601, opened 2018-03-22
Rewrite installation instructions

kxmo:
I've recently setup Loomio using the old instructions and saw room for some improvements:
- Rearranged some of the instructions - general instructions in setup_environment.md and Loomio specific in quicksetup.
- Added some instructions that were missing
- Removed unneeded steps or those that could become out of date
I have tested these instructions on Ubuntu 14.04. I don't have access to a Mac but it would be good if someone could test them there.


docs/en/development_handbook/quickstart.md
```diff
@@ -33,16 +64,22 @@ _NB: Take note of the email and password generated at the end of this task; you'
 
 ## Build the frontend client
 ```
-cd client && gulp compile
+gulp compile
 ```
 ## Launch the server
 ```
+cd angular
```
robguthrie: Are you sure you intended to write angular? I think you want to cd to client here.

robguthrie:
Thanks for this! Looks pretty good.

docs/en/development_handbook/quickstart.md
kxmo: If the user keeps the same terminal instance open they're already in client (line 54).
I think gulp changes the working directory anyway so this step may not be needed

gdpelican:
This is amazing, thanks @kxmo !

PR #4606, opened 2018-03-28
UX upgrade for proposals / checks

gdpelican:
screen shot 2018-03-28 at 11 43 42 pm
screen shot 2018-03-28 at 11 43 56 pm
screen shot 2018-03-28 at 11 44 08 pm
screen shot 2018-03-28 at 11 44 22 pm
screen shot 2018-03-28 at 11 44 27 pm
screen shot 2018-03-28 at 11 44 41 pm

PR #4612, opened 2018-03-30
limits button width to a fourth of container even on medium screens

### LINKED ISSUE #4607: vote buttons bigger than the proposal container

juliencarnot:
Looks like PR #4606 min-width reduction was not enough.

fixes #4607 and avoids cut-off text in longer locales...

image


gdpelican:
Thanks for the prompt here @juliencarnot , appreciate it.

I wasn't happy with the 22% magic number here, and also think that we should be operating on `max-width` rather than `min-width` here (since we're basically trying to say 'don't ever get bigger than a quarter the width of the container), but otherwise it looks sweet and I'll merge and deploy :)

PR #4613, opened 2018-03-30
add "new decision" button to group page

### LINKED ISSUE #4573: "New Decision" button
On a group's main profile page, it would be great to have a NEW DECISION button to start a decision. Currently a user has to click on Decisions on the main sidebar, and then the NEW DECISION button is available. It could be placed in the top right corner of "Current Decisions" window, as shown below.

new decision

Thanks!

JosiMcClellan:
Closes #4573 "New Decision" button

The picture in the issue shows the button in the card's corner with no padding.  We thought it was best to keep the padding consistent with the rest of the site instead, but we're happy to change it.


JosiMcClellan:
I see there's some testing to work out before a PR.

PR #4616, opened 2018-04-03
Use current proposal / check styling for poll

gdpelican:


gdpelican:
screen shot 2018-04-04 at 8 55 59 am
screen shot 2018-04-04 at 8 56 12 am

robguthrie:
you're on 🔥 ! Such a nice improvement

robguthrie:
In the poll image, what is the state of the grey backgrounded option?

gdpelican:
That's on hover

PR #4632, opened 2018-04-25
export/import group data via json file

robguthrie:


gdpelican:
@robguthrie Does this need additional work to be merged? It looks like the existing code doesn't really do anything other than memberships...

robguthrie:
Please leave it for now. I modified it for memberships to repair the damage
last week. I'll pick it up again soon.

It's not super fast but it works well enough to get jobs done.

gdpelican:
Would be nice to get this merged in, @robguthrie ; anything I can do to support?

robguthrie:
Hi @gdpelican. This version does not gobble memory when exporting large groups.. we stream data straight into the file with one json record per line. Ruby process stays around 100 mb rather than growing to 8gb.

We may want to think about how to use the .import method, maybe by doing a group by table then import after dropping/ignoring existing records.

Anyway, I'm looking to complete this over the next day or so. Hello.

robguthrie:
@gdpelican the export and import of our biggest groups can happen in 5 minutes or so each now, without crazy memory usage.

I'm thinking burndown lists now.

- Consider who can export a group and it's subgroups. If a group admin cannot access a subgroup, do they get to export it?
- Self service export - create background job and upload the file somewhere. I'd like to reuse our own attachments system for this somehow.
- excluded fields. We're going to exclude some tokens and keys right? Which ones should we exclude?

robguthrie:
Ok so we're now excluding user confidential fields.

So I think now we enable users to export json for any group they belong to. 
They select the parent group then get told they'll be emailed when the file is available. They then get an email with a link to their json for the group. It will contain the top level group and all the subgroups they belong to in a single json file.

I'm thinking that we create an attachment owned by the user with no attachable, and add the exported json to it, then add a delete delayed job scheduled for a week later.

robguthrie:
@gdpelican this is in final stages now. Just going to finish the email styling. CR appreciated

app/controllers/api/groups_controller.rb
```diff
@@ -37,6 +37,11 @@ def upload_photo
     respond_with_resource
   end
 
+  def export
+    service.export(group: load_and_authorize(:group), actor: current_user)
```
gdpelican: Pedantic, but you should be able to do
```
service.export(group: load_resource, actor: current_user)
```
because it should be the service's job to authorize whether you can export or not.

app/models/formal_group.rb
```diff
@@ -48,7 +48,7 @@ class FormalGroup < Group
   belongs_to :default_group_cover
 
   has_many :subgroups,
-           -> { where(archived_at: nil).order(:name) },
```
gdpelican: I agree with this, but wonder if we need to check the `/g/:key/subgroups` query to make sure it's still returning the same thing?

app/models/guest_group.rb
```diff
@@ -5,6 +5,14 @@ def id_and_subgroup_ids
     Array(id)
   end
 
+  def subgroups
+    Group.none
+  end
+
+  def documents
+    Document.none
```
gdpelican: Hmmmm I wonder if we want to move the `has_many :documents` line from FormalGroup to Group instead of doing this.

app/models/membership.rb
```diff
@@ -89,6 +89,6 @@ def discussion_readers
   private
 
   def set_volume
-    self.volume = user.default_membership_volume if group.is_formal_group?
+    self.volume = user.default_membership_volume if id.nil? && group.is_formal_group?
```
gdpelican: Oh hmm what is this change needed for?

I think more idiomatic is `self.new_record?`

app/serializers/discussion_reader_serializer.rb
```diff
@@ -33,10 +33,6 @@ def discussion_reader_id
     object.id
   end
 
-  def discussion_reader_volume
```
gdpelican: This gets used in `discussion_model.coffee`, I wonder if we need to account for that?

gdpelican: Also, wonder if volume changes are not strictly necessary for group export

app/jobs/group_export_job.rb
```diff
@@ -0,0 +1,11 @@
+class GroupExportJob < ActiveJob::Base
+  queue_as :low_priority
+
+  def perform(group, actor)
+    groups = actor.groups.where(id: group.all_groups)
+    filename = GroupExportService.export_filename_for(group)
+    GroupExportService.export(groups, filename)
```
gdpelican: I would have expected
```
filename = GroupExportService.export(group)
```
and then call group.all_groups from within the service

client/angular/components/group_page/group_actions_dropdown/group_actions_dropdown.coffee
```diff
@@ -11,6 +11,18 @@ angular.module('loomioApp').directive 'groupActionsDropdown', ->
   replace: true
   controller: ['$scope', ($scope) ->
 
+    $scope.canExportData = ->
+      Session.user().isMemberOf($scope.group)
+
+    $scope.openGroupExportModal = ->
+      ModalService.open 'ConfirmModal', confirm: ->
+        submit: -> Records.groups.export($scope.group.id)
```
gdpelican: We tend towards actions on the model, so
```
# group_model.coffee
export: =>
  @remote.post(@id, 'export')
```
```
submit: $scope.group.export
```

spec/controllers/api/groups_controller_spec.rb
```diff
@@ -12,6 +12,11 @@
     sign_in user
   end
 
+  describe 'export' do
+    it 'kicks off a group export job'
```
gdpelican: Would be good to write these, even if it's just a check for 200 and 403. Maybe check to ensure Document.count is incremented too.

app/services/group_export_service.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def self.puts_record(record, file, ids)
+    table = record.class.table_name
+    return if ids[table].include?(record.id)
+    ids[table] << record.id
+    file.puts({table: table, record: record.as_json(JSON_PARAMS[table])}.to_json)
+  end
+
+  def self.import(filename)
+    tables = File.open(filename, 'r').map { |line| JSON.parse(line)['table'] }.uniq
```
gdpelican: There's some stuff that smells a bit here, but I haven't got a better solution off the top of my head given that (I'm assuming) we don't want to port all of the file into memory. I think json parsing each line of the file N+1 times is going to hurt us, and wonder if there's a way to avoid it.

I wouldn't be opposed to, say, iterating through each line of the file, parsing it, and then,
if the table matches the current table, `klass.new` and append it to an array of records
if the table doesn't match, `klass.import`, flush the existing array, and start a new one with the new table

Then we just iterate through the file once, parsing each line onece, and still maintain support for weirdo input like a stray group in a middle of a big run of discussions or whatnot.

gdpelican: PS I believe import will silently fail imports with ids that exist already, which would be the same behaviour as this would exhibit

app/services/group_export_service.rb
```diff
@@ -0,0 +1,68 @@
...
+          group.send(relation).find_each(batch_size: 20000) do |record|
+            puts_record(record, file, ids)
+          end
+        end
+      end
+    end
+  end
+
+  def self.export_filename_for(group)
+    "tmp/#{DateTime.now.strftime("%Y-%m-%d_%H-%M-%S")}_#{group.name.parameterize}.json"
```
gdpelican: A note that the group.name.parameterize will fail for guest groups, which we should be able to export just fine.

app/controllers/api/groups_controller.rb
robguthrie: Kind of related. One thing that might be relevant is anonymous polls. 
A data export would allow people to see who voted what. 

If it were not for anonymous polls then I'd say that anyone who is a member should be able to export the group data.

If there are anonymous polls, I don't know if anyone should be able to download the data. It's a weird situation. What are your thoughts?

app/jobs/group_export_job.rb
robguthrie: The plan was that anyone in a group can export the data. Meaning they should only be able to export data for groups they belong to. So I want to pass in only groups the user belongs to.

app/models/guest_group.rb
robguthrie: done

app/models/membership.rb
robguthrie: I've working to remove all unnecessary N+1 queries. Reading volume triggers a query in a couple of places (checking default membership volume and group volume) and it also pollutes the export because it returns a value that isn't actually what the column actually contains.

So that's why I've moved to using a method that clearly says it's giving a computed value for volume rather than overloading the simple accessor. I like it more and it hasn't really been a problem to change.

app/services/group_export_service.rb
robguthrie: import totally fails if some of the ids already exist. so the check is necessary.

Thanks for helping to try and improve, but i think I'd like to call it as good enough for now. It's the best solution (fastest export for large groups by miles) after quite a few attempts and I'd like to move on.

client/angular/components/group_page/group_actions_dropdown/group_actions_dropdown.coffee
robguthrie: done, thanks

robguthrie:
@gdpelican my show stopper issue with this is anonymous polls. What do you think we should do? 

Some options that come to mind:
- Do nothing and have a backdoor where users can see who said what
- Mangle data on anonymous polls
- Don't include anonymous polls in export


Any ideas?

PR #4638, opened 2018-05-01
Better versions

Space-Ed:
Ready for code review and design input, not implemented for polls or groups

### DIFF
--- a/app/controllers/api/versions_controller.rb
+++ b/app/controllers/api/versions_controller.rb
@@ -1,7 +1,22 @@
 class API::VersionsController < API::RestfulController
   def show
     versions = model.versions
-    self.resource = versions[params[:index].to_i|| versions.count-1]
+    index = params[:index].to_i
+
+    versions[index].object_changes = versions.first(index+1).reduce({}) do |changes, version|
+      # for each field, nil values become the latest value we changed to
+      keys = model.class.always_versioned_fields;
+
+      always_changes = keys.map do |k|
+        existing = changes.dig(k, 1)
+        [k, version.object_changes[k.to_s] || [existing, existing]]
+      end.to_h
+
+      version.object_changes.merge always_changes
+    end
+    
+    self.resource = versions[index]
+
     respond_with_resource
   end
 
--- a/app/controllers/api/versions_controller.rb
+++ b/app/controllers/api/versions_controller.rb
@@ -1,4 +1,9 @@
 class API::VersionsController < API::RestfulController
+  def show
+    versions = model.versions
+    self.resource = versions[params[:index].to_i|| versions.count-1]
+    respond_with_resource
+  end
 
   private
 
@@ -6,7 +11,16 @@ class API::VersionsController < API::RestfulController
     PaperTrail::Version
   end
 
+  def model
+    load_and_authorize(:group, optional:true)||
+    load_and_authorize(:discussion, optional:true)||
+    load_and_authorize(:comment, optional:true) ||
+    load_and_authorize(:poll, optional:false)
+  end
+
+
   def accessible_records
+    # used to create an index of all available records
     records = load_and_authorize(params[:model]).versions.order(created_at: :desc)
     records = records.where(event: 'update') if params[:model] == 'discussion'
     records
--- a/client/angular/components/text_diff/text_diff.coffee
+++ b/client/angular/components/text_diff/text_diff.coffee
@@ -6,6 +6,16 @@ angular.module('loomioApp').directive 'textDiff', ->
   templateUrl: 'generated/components/text_diff/text_diff.html'
   controller: ['$scope', ($scope) ->
     $scope.diff = ->
-      diff = new DiffMatchPatch()
-      diff.diff_prettyHtml diff.diff_main($scope.before, $scope.after)
+      differ = new DiffMatchPatch()
+      diff = differ.diff_main($scope.before||"", $scope.after||"")
+      differ.diff_cleanupSemantic(diff)
+      $scope.prettyHtml(diff)
+
+    $scope.prettyHtml = (diff) ->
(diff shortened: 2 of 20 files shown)


client/angular/components/text_diff/text_diff.coffee
```diff
@@ -0,0 +1,21 @@
...
+
+angular.module('loomioApp').directive 'textDiff', ->
+  scope: {before: '=', after: '='}
+  restrict: 'E'
+  templateUrl: 'generated/components/text_diff/text_diff.html'
+  controller: ['$scope', ($scope) ->
+    $scope.diff = ->
+      differ = new DiffMatchPatch()
+      diff = differ.diff_main($scope.before||"", $scope.after||"")
+      differ.diff_cleanupSemantic(diff)
```
gdpelican: I'd love to pull this out into a shared library so we can use it in several places (since it doesn't have to do with angular)

Something like

```
# shared/helpers/text.coffee
DiffMatchPatch = require 'diff-match-patch'
module.exports = ->
  applyDiff: (before, after) ->
    differ = new DiffMatchPatch()
    ... (other stuff to make the diff great!)
    diff
```

Then the controller code turns simply into
```
{ applyDiff } = require 'shared/helpers/text.coffee'
$scope.diff = ->
  applyDiff($scope.before, $scope.after)
```

client/angular/components/thread_page/context_panel/context_panel.coffee
```diff
@@ -64,6 +63,11 @@ angular.module('loomioApp').directive 'contextPanel', ['$rootScope', ($rootScope
       icon: 'mdi-pin-off'
       canPerform: -> AbilityService.canUnpinThread($scope.discussion)
       perform:    -> ThreadService.unpin($scope.discussion)
+    ,
+      name: 'show_history',
+      icon: 'mdi-clock'
+      canPerform: -> $scope.discussion.edited()
```
gdpelican: If we're not using edited() anywhere else I'd be happy to for this to simply be
```
$scope.discussion.versionsCount > 1
```
and to get rid of the `edited()` method

client/angular/components/thread_page/context_panel/context_panel.haml
```diff
@@ -4,7 +4,7 @@
       %i.mdi.mdi-pin{ng-switch-when: "pinned"}
 
     .context-panel__h1.lmo-flex__grow{in-view: "showLintel(!$inview)"}
-      %h1.lmo-h1.context-panel__heading{ng-if: "!discussion.translation"}
+      %h1.lmo-h1.context-panel__heading{ng-if: "!discussion.translaion"}
```
gdpelican: Reckon we don't need this one :)

client/angular/components/thread_page/revision_history_modal/revision_history_content.coffee
```diff
@@ -0,0 +1,19 @@
+Records           = require 'shared/services/records.coffee'
+
+angular.module('loomioApp').directive 'revisionHistoryContent', ->
+  scope:{ model:'=', version:'='}
+  restrict:'E'
+  templateUrl: 'generated/components/thread_page/revision_history_modal/revision_history_content.html'
+  controller:['$scope', ($scope) ->
+
+    $scope.diffType = (field_name) ->
```
gdpelican: Looks like we no longer use `diffType()` ?

client/angular/components/thread_page/revision_history_modal/revision_history_modal.coffee
```diff
@@ -1,45 +1,42 @@
...
+
+    $scope.isNewest = ->
+      $scope.currentIndex == latestIndex
+
+    $scope.setNextRevision = ->
+      if !$scope.isNewest()
+        $scope.currentIndex += 1;
+        $scope.getVersion()
+
+    $scope.setPreviousRevision = ->
```
gdpelican: I'd love to cut down on these methods some. It seems like we could have something like

```
$scope.displayVersion = (index) ->
  $scope.currentIndex = index
  $scope.getVersion()
```
then, in the haml:
```
.md-button{ng-if: "currentIndex > 0", ng-click: "displayVersion(0)"} # display the first version
.md-button{ng-if: "currentIndex > 0", ng-click: "displayVersion(currentIndex-1)"} # display the previous version
.md-button{ng-if: "currentIndex < maxIndex", ng-click: "displayVersion(currentIndex+1)"} # display the next version
.md-button{ng-if: "currentIndex < maxIndex", ng-click: "displayVersion(maxIndex)"} # display the last version
```

spec/controllers/api/versions_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      discussion.group.add_admin! user
+      sign_in user
+      discussion.update(title: "Ready Player One")
+      discussion.update(private: true)
+      discussion.update(title: "Ready Player Two")
+      comment.update(body: "Ready Player One")
+
+    end
+
+    it '(zero) should return the original version' do
```
gdpelican: These tests are fantastic!!

app/controllers/api/versions_controller.rb
```diff
@@ -1,12 +1,41 @@
 class API::VersionsController < API::RestfulController
+  def show
```
gdpelican: Cool. This is the most complex controller method in the app at the moment; I'd love to see it
a) slim down some, and
b) move into a service (`VersionService#compose` maybe? Which takes a model and index, and returns a version to return

```
def show
  # (service defaults to the VersionService constant)
  self.resource = service.compose(versions: model.versions.first(index+1))
  respond_with_resource
end
```
then, in the version service:
```
class VersionService
  def self.compose(versions:)
    index = versions.count
    # the existing version reduce / merge code
    # returns a version with composed object_changes
  end
end
```

PR #4647, opened 2018-05-06
Re-add new decision button on group page

gdpelican:
Since we've removed the option to set a poll's group from the share form, there's no no way to start a poll within a group, which we've just had a paying customer complain about. I'd like to add this back in as a fix to fill the gap for now.

screen shot 2018-05-06 at 2 33 00 pm


gdpelican:
Open to moving it to match the invite button, but think move poll should go through the design process.

PR #4651, opened 2018-05-07
Backfill discussion members

gdpelican:


robguthrie:
Looks good! I wonder if it will blow up somehow, and how to tell if that happens, but seems like it's good enough to try it out.

db/migrate/legacy/backfill_discussion_members_job.rb
```diff
@@ -0,0 +1,24 @@
...
+  def run_for_timeframe(since: 10.years.ago, till: 1.year.from_now)
+    Discussion
+      .includes(:items)
+      .within(since, till, :last_activity_at)
+      .where.not(author_id: User.helper_bot.id)
+      .find_each(batch_size: 250) do |d|
+      to_import = d.items
+                   .where.not(user_id: d.guest_group.member_ids)
+                   .map do |item|
+        Membership.new(group_id: d.guest_group_id, user_id: item.user_id, accepted_at: item.created_at)
```
gdpelican: Hmm, I have to unique this by user somehow, otherwise it's going to try to import multiple memberships for the same user.

robguthrie: Also I wonder if you want to copy volume over as is

gdpelican: 'copy volume over' = copy from, the discussion reader?

I'm about 70% sure the discussion reader takes precedence over the guest group membership at the moment, so I think it's actually more correct to leave those volumes as nil until we get more clear on what to do with volume on guest group memberships

gdpelican:
Want to have a gander at this at your leisure @robguthrie ? I got the wrong prod DB last night (woof), so may try a dry run again tonight.

robguthrie:
Running this now. Seems good to go but I'll report back when it's finished.

PR #4653, opened 2018-05-07
Remove need for .coffee extensions with browserify requires

gdpelican:
By passing extension options to browserify, we can avoid the need to specify the extension all over the client.

This'll come in handy when it comes time to transpile the shared files into js so that react native can read them easily. Also why write `.coffee` 950 times when you can just write it three times instead?

Fingers crossed on a green test suite on this one :D


robguthrie:
Nice one! Looks green enough to me

PR #4655, opened 2018-05-07
Support SSL for SMTP connector

codders:
When using Amazon Simple Email Service (SES), handshake with the SMTP server fails with:

> Job GroupMailer.invitation_created (id=2224) FAILED (0 prior attempts) with Net::ReadTimeout: Net::ReadTimeout

because of issues with the SSL handshake. Passing `ssl: true` to the ActionMailer config fixes this. This PR adds provides the option to pass 'SMTP_USE_SSL' from the application environment to enable the flag in ActionMailer


gdpelican:
Sounds good to me, thanks for this!

PR #4656, opened 2018-05-08
Fix stance icon for

robguthrie:
We've been getting errors for too long about missed yesterday emails raising error on this.

https://sentry.loomio.io/sentry/loomio-6y/issues/4111/

This just moves this little method into a helper so the thing does not blow up. Sorry to add methods to the email helper but it seems like the easiest way to fix this and I'm on a small mission to reduce our error rate.


robguthrie:
whoops should have cut from master. re-doing

PR #4660, opened 2018-05-09
Explain impact of membership removal

robguthrie:


robguthrie:
@gdpelican I think this is good to merge

PR #4662, opened 2018-05-09
Stable

robguthrie:
fix groups not saving on discuss.digital


gdpelican:
Yeah, super happy with this, it's the 'back of the napkin' fix I had in mind when I glanced at it earlier today.

PR #4663, opened 2018-05-09
catch recaptcha missing exceptions

robguthrie:
@gdpelican how might I make an error show up on the form?

Either way, this is an improvement over current situation. Tested locally

### DIFF
--- a/app/controllers/api/registrations_controller.rb
+++ b/app/controllers/api/registrations_controller.rb
@@ -38,7 +38,6 @@ class API::RegistrationsController < Devise::RegistrationsController
   end
 
   def configure_permitted_parameters
-    raise User::RecaptchaMissingError.new if ENV['RECAPTCHA_APP_KEY'] && (!params[:recaptcha] || !pending_membership_is_present?)
     devise_parameter_sanitizer.permit(:sign_up) do |u|
       u.permit(:name, :email, :recaptcha)
     end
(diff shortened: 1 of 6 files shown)


gdpelican:
Hm, we should be doing this differently; we already have a validation on the user called `ensure_recaptcha`, so `require(:recaptcha)` should be unnecessary.

What happens if you take out all of this code, including the `require(:recaptcha)` line?

From there, we'll want to put a validation error like
```
%validation_errors{subject: "user", field: "recaptcha"}
```
which should show up just fine if the client recaptcha fails.

gdpelican:
Oh, plus a check in `ensure_recaptcha` which just returns if RECAPTCHA_KEY isn't set.

robguthrie:
Ok.. thanks for that guidance, have done that.

It's not displaying the validation message on failure, which I this is something about the auth_service and registration_model, I think if you have a look you'll probably understand why.

gdpelican:
I'm hesitant about merging / deploying this at the moment because it currently prevents the creation of helper bot / Patrick Swayze in dev if you have RECAPTCHA_KEY set, which is not the thing we want. Should work fine in prod, but also potentially not (and the whole test suite runs without RECAPTCHA_KEY set, so we don't really know from that)

Not super confident on the right fix (I'm tempted to make a UserService#create and put the check there?), but would like to fix it before more.

robguthrie:
no I don't think they overlap much, and we may keep the recaptcha as is
too, when trying to cut work down.

PR #4670, opened 2018-05-10
ensure outcome author is emailed

### LINKED ISSUE #4495: Time poll didn't send calendar invites to email participants
I'd emailed three people a shareable link to a threadless time picker poll (XdExwbG3). They all voted. I closed the poll, and received a calendar invite myself, but none of the other three participants did. It made for a great deal of irl meeting time confusion. 

I really want to introduce people to Loomio by using time pickers instead of Doodle, but I feel like I can't really quite trust it at the moment :'(

robguthrie:
I guess we can remove the announcement column on events after this?

fixes #4495

### DIFF
--- a/app/models/events/outcome_created.rb
+++ b/app/models/events/outcome_created.rb
@@ -22,6 +22,6 @@ class Events::OutcomeCreated < Event
   end
 
   def notify_author?
-    announcement && poll.author_receives_outcome
+    poll.author_receives_outcome
   end
 end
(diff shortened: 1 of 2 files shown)


robguthrie:
@gdpelican where do you think we should put a test for this?

gdpelican:
I put one into event_spec :)

robguthrie:
actually, later on, I wondered if this should have been on outcome_announced rather than created. So if you send announcements, you get one.

PR #4672, opened 2018-05-11
More proper diffing of markdown

gdpelican:
The current revision history functionality really doesn't do a great job with markdown:

This extracts the text diffing algorithm from Discourse and gem-ifys it so we can use it for ourself, harharhar. This is a server side rendering solution, which I'm keen to expand for use across all version fields, not just title and description.

screen shot 2018-05-11 at 5 22 20 pm
screen shot 2018-05-11 at 5 22 30 pm
screen shot 2018-05-11 at 5 22 45 pm
screen shot 2018-05-11 at 5 23 51 pm


robguthrie:
Cool. I'm sorry this isnt the place for it, but, I'll shoot anyway. Is it possible to, now that this work has been done, link directly to the diff from the "thing edited" item in a thread.

So when someone edits a proposal, you get a link to "view changes". I think that would really make this great work shine, because that's when people most want to see the changes.

Forgive me for the tangential comment, and also if that feature already exists and I'm tripping.

gdpelican:
Yeah, totally a valid feature request. I'm scrambling a bit with this one because I'm not happy with the current non-markdown experience, but am happy to put that functionality in as a follow-up; it won't be too tough.

robguthrie:
My personal take is: It's a feature of markdown that it looks reasonable without processing, and that the nature of edits are that sometimes people are doing really weird things, or fixing broken markdown in the first place. There isn't a lot of people looking at a diff view, and when they do, they're not looking to understand the presentation, they're looking for changes, which is what a diff does so well.

I think it's actually the wrong thing to do (to render markdown when comparing a diff) but that's just my two cents - but I'm not surprised it's technically very challenging.

I remain supportive of you trying something challenging, but don't think you need to feel like it's a must have.

gdpelican:
I had the same thought, but I'd invite you to look at the current results
of the diff match patch engine, which are less than acceptable and not
easily fixable. I don't think this is challenging at all; someone else has
already written all of the hard bits.

robguthrie:
Nice one. Full support!

gdpelican:
Oh, heh, it looks like we took the edit thread items out; does the ability to link them directly to a version mean we should put them back in again?

robguthrie:
Na, we don't need it on thread edited, but it would be really good for poll edited.

PR #4673, opened 2018-05-12
dont track clicks in emails, we dont look anyway

robguthrie:


fauno:
this also means loomio email won't be treated as phishing by antispam :)

#4022

PR #4387, opened 2017-11-04
Accessibilityimprovements

juliencarnot:
see #4386 and please review, I'm just beginning to understand angularjs!

PR #4675, opened 2018-05-13
Accessibility improvements

gdpelican:
Forked from #4387


gdpelican:
Thanks for this work @juliencarnot and sorry it took so long to come around on.

PR #4680, opened 2018-05-14
Allow starting discussions from slack (in addition to polls)

gdpelican:


gdpelican:
@robguthrie Can I get an amen or a language pass?

PR #4682, opened 2018-05-15
Fix add to homescreen

gdpelican:


robguthrie:
@gdpelican I tested this on FF but can't repro a 30 second hang. Any tips?

PR #4685, opened 2018-05-15
show sidebar on medium size desktop screens

robguthrie:


app/serializers/group_serializer.rb
```diff
@@ -74,7 +74,7 @@ def include_logo_url_medium?
   def cover_urls
     {
       small:      cover_photo.url(:desktop),
-      medium:     cover_photo.url(:desktop),
+      medium:     cover_photo.url(:largedesktop),
```
robguthrie: on my retina screen I was being served a 960px file, which looked bad. need more pixels

robguthrie:
turns out my window size was 1279, which ment the sidebar was behaving really crazy. this makes it behave correctly in those just less than cases

PR #4687, opened 2018-05-17
Mention users from any group you belong to

robguthrie:
just started working, need feedback and fix specs


app/controllers/api/users_controller.rb
```diff
@@ -0,0 +1,19 @@
+class API::UsersController < API::RestfulController
+  def index
```
gdpelican: I don't think you want to make a new controller here... we have the ProfileController which acts as our UsersController at this point, which `Records.users` points to... easier to stick with that I believe.

gdpelican: Then we don't have to do all these gymnastics around moving apiEndpoint around; if anything we should consider renaming ProfileController to UsersController, which is what it always wanted to be anyway.

app/models/concerns/events/notify/mentions.rb
```diff
@@ -6,7 +6,11 @@ def trigger!
 
   # send event notifications
   def notify_mentions!
-    mention_recipients.each { |mentionee| Events::UserMentioned.publish!(eventable, user, mentionee) }
+    AnnouncementService.create(
+      model: eventable,
+      actor: user,
+      params: {kind: "user_mentioned", recipients: {user_ids: mention_recipients.map(&:id)}}
```
gdpelican: Should be pluck(:id), right?

app/controllers/api/users_controller.rb
gdpelican: Adding `profile#index` instead of `users#index` cuts this PR down to about 10 files, instead of 21

app/controllers/api/users_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+                        first:       "#{params[:q]}%",
+                        subsequent:  "% #{params[:q]}%")
+    end
+    respond_with_collection serializer: Simple::UserSerializer
+  end
+
+  private
+  def accessible_records
+    resource_class.active.verified.distinct.
+    joins(:memberships).where('memberships.group_id': current_user.group_ids).
```
gdpelican: This is duplicate logic of has_mentions#mentioned_users, which makes me think it wants to be a scope on User

```
# user.rb
scope :mentionable_by, (user) -> {
  active.verified.distinct.joins(:memberships).where("memberships.group_id": user.group_ids).where.not(id: user.id)
}
```

app/controllers/api/users_controller.rb
robguthrie: ok thanks will do. I think the move of apiendpoint is technically right, but agree it does not need to be included here.

app/controllers/api/users_controller.rb
robguthrie: nice

robguthrie:
@gdpelican thanks for the help, it's looking much better now.

app/controllers/api/profile_controller.rb
```diff
@@ -5,6 +5,14 @@ def show
     respond_with_resource serializer: UserSerializer
   end
 
+  def mentionable_users
+    @users = resource_class.mentionable_by(current_user).search_for(params[:q])
```
gdpelican: I'd argue that you want to run this through `instantiate_collection` so that it pages; no point in returning more than 30 results here; in fact we could probably get away with first 5/10.

```
instantiate_collection do
  collection.mentionable_by(current_user).search_for(params[:q])
end
respond_with_collection serializer: Simple::UserSerializer
```

what trouble are you running into with respond_with_collection?

gdpelican: Should be something like

```
def mentionable_users
  instantiate_collection { |collection| collection.search_for(params[:q]) }
  respond_with_collection serializer: Simple::UserSerializer, root: :users
end

def accessible_records
  resource_class.mentionable_by(current_user)
end
```

robguthrie: When I use respond_with_collection it always returns current_user, I couldn't see how that was being done.

gdpelican: Oh right, it's because 'resource' is always defined in the profile controller (simplified version:)

```
def resource
  current_user 
end

def resources_to_serialize
  Array (resource || collection)
end
```

I'm not certain why I picked resource first; it's probably arbitrary, but for now I'd override `resources_to_serialize` to be `collection || resource`, and take a look later at seeing if we could update it in restful_controller

app/extras/queries/users_to_mention_query.rb
```diff
@@ -1,6 +1,6 @@
 class Queries::UsersToMentionQuery
   def self.for(model)
-    model.mentioned_group_members
+    model.mentioned_users
          .where.not(id: model.group.members.mentioned_in(model)) # avoid re-mentioning users when editing
```
gdpelican: Have to be careful here, I think this wants to be `model.members`, so that you can't remention people in guest groups as well. Also have to make sure that stance, comment, etc. respond correctly to 'members'

app/controllers/api/profile_controller.rb
robguthrie: I initially did see that, try it, but it did not work.... and now I don't know why, because now it does. Anyway, thanks!

robguthrie:
@gdpelican please have a look at my last commit. I've changed tack and now subclass announcement_created. I did this for a couple of reasons, but the main one was that user.mentioned_in would be broken without it.

app/models/events/user_mentioned.rb
```diff
@@ -1,18 +1,14 @@
-class Events::UserMentioned < Event
-  include Events::Notify::InApp
-  include Events::Notify::ByEmail
-
-  def self.publish!(model, actor, mentioned_user)
-    super model, user: actor, custom_fields: { mentioned_user_id: mentioned_user.id }
+class Events::UserMentioned < Events::AnnouncementCreated
```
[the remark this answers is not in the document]
robguthrie: Cool, Nice idea.

mentioned_in looks for notifications hanging off previously created user_mentioned events, so we have a few problems with them being kind: announcement_created, which can be solved by them being user_mentioned.

- switching tracks half way, editing previous comments will re-mention because of the change to looking for announcement_created from user_mentioned in mentioned_in
- we have a database that at a certain point in history, has a completely different method of mentioning, which may haunt us.
- desired behaviour for mentioning vs announcing is different. if you've already announced a poll, then you mention someone in it later, unless they have different kinds, the announcement will look like a mention to mentioned_in, so they wont receive the mention, even though you could have re announced it to them.

Plus I think it is nice to prove that this code is really decoupled and we can use it for more than one event kind.

robguthrie:
@gdpelican I think this is ready to go.  Do you agree?

app/models/events/announcement_created.rb
```diff
@@ -12,6 +12,10 @@ def memberships
     @memberships ||= Membership.where(id: custom_fields['membership_ids'])
   end
 
+  def members
+    User.where(id: memberships.pluck(:user_id))
```
gdpelican: Can probably move this stray method back into notification_recipients

client/shared/helpers/listen.coffee
```diff
@@ -7,9 +7,9 @@ EventBus = require 'shared/services/event_bus'
 module.exports =
   listenForMentions: ($scope, model) ->
     $scope.unmentionableIds = [model.authorId, Session.user().id]
```
gdpelican: Hm, I wonder if we could get rid of this client-side check now; we should be filtering out these users on the server I reckon. Then this turns into

```
$scope.fetchByNameFragment = (q) ->
  Records.users.fetchMentionable(q).then (response) ->
    $scope.mentionables = Records.users.find(_.pluck(response.users, 'id'))
```

client/shared/interfaces/user_records_interface.coffee
```diff
@@ -1,8 +1,15 @@
 BaseRecordsInterface = require 'shared/record_store/base_records_interface'
 UserModel            = require 'shared/models/user_model'
+RestfulClient        = require 'shared/record_store/restful_client'
```
gdpelican: Don't need this requires anymore

client/shared/record_store/base_records_interface.coffee
```diff
@@ -4,22 +4,26 @@ utils         = require './utils'
...
         @collection.ensureUniqueIndex(name)
 
-      @remote = new RestfulClient(@model.apiEndPoint or @model.plural)
+      @remote = new RestfulClient(@apiEndPoint or @model.plural)
 
     onInterfaceAdded: ->
 
-    setRemoteCallbacks: (callbacks) ->
-      _.merge @remote, _.pick(callbacks, ['onPrepare', 'onSuccess', 'onUploadSuccess', 'onFailure', 'onCleanup'])
+    setRemoteCallbacks: (callbacks, remote = @remote) ->
```
gdpelican: Don't think we need this change anymore (and I'd be hesitant to allow passing in of remotes here, seems dangerous)

client/shared/record_store/base_records_interface.coffee
```diff
@@ -4,22 +4,26 @@ utils         = require './utils'
 module.exports =
   class BaseRecordsInterface
     model: 'undefinedModel'
+    collectionName: null
```
gdpelican: I don't think we're using this option anymore? Something would seem quite off to me if we were making a db table that wasn't the plural of the model.

gdpelican:
I'm ready to merge this, but one minor follow-on improvement which might be nice:
@mentioning someone adds them to the guest group just fine, but they don't appear immediately in the members card. Maybe we need to add `Events::LiveUpdate` to the UserMentioned event?

PR #4688, opened 2018-05-18
Display emails for pending invites by inviter

gdpelican:


gdpelican:
Care to spot check me on this one @robguthrie ?

robguthrie:
Looks good to me. Interesting choice to pass email through the name field. Does that mean email arrives twice?

Also I notice that I used simple::userserializer in mention freely, which does not look much simpler than this one. should I use normal user serializer over there?

PR #4690, opened 2018-05-19
Groupless threads (WIP)

gdpelican:
This allows you to create a thread without a group, invite people into its guest group, and request membership to a thread if it's membership_on_request.

I haven't put much UX thinking into this yet, nor tested through many full workflows, but it's a good start.

This includes the mention-freely branch, because mentions need to work slightly differently with standalone discussions.


robguthrie:
Nice! I think major UX work can come afterwards, there is a pattern mostly already here that this can fit within.

PR #4691, opened 2018-05-19
Widen discussion / poll text areas, remove maxheight on lmo textarea

gdpelican:


gdpelican:
Thanks again for the prompt here @juliencarnot ; we'll give this a go to see if we can improve the writing experience a bit.

PR #4692, opened 2018-05-20
retry create if event not unique

robguthrie:
we do this successfully in discussion readers, so I wonder if we want to do it here too for situations like: https://sentry.loomio.io/sentry/loomio-6y/issues/4389/


robguthrie:
two alternatives: 
- limited retrys: http://blog.mirthlab.com/2012/05/25/cleanly-retrying-blocks-of-code-after-an-exception-in-ruby/
- create a new sequence for each discussion so this does not happen in the first place (not sure how much migration work that is)

Personally, I think we could get away with this for now.

PR #4704, opened 2018-06-04
change missed_yesterday to catch_up and enable weekly frequency

robguthrie:
as requested by a customer who wanted weekly frequency for their private host


app/jobs/send_weekly_catch_up_email_job.rb
```diff
@@ -0,0 +1,13 @@
+class SendWeeklyCatchUpEmailJob < ActiveJob::Base
+  def perform
+    zones = User.pluck('DISTINCT time_zone').select do |zone|
+      Time.find_zone(zone) &&
```
gdpelican: Here's my suggestion re: duplicate jobs, but whatever you'd like is good

```
def perform(period)
  settings = settings_for(period)
  zones = User.pluck('DISTINCT time_zone').select { |zone| Time.find_zone(zone) }.select(settings[:include])

  User.email_catch_up.where(time_zone: zones).find_each do |user|
    UserMailer.delay(priority: 30).catch_up(user, settings[:period], settings[:name])
  end
end
  
private

def settings_for(period)
  case period.to_sym
  when :weekly
    {
      period: 1.week.ago,
      include: Proc.new do |zone|
        DateTime.now.in_time_zone(zone).wday == 1 &&
        DateTime.now.in_time_zone(zone).hour == 6
      end
    }
  when :daily
    {
      period: 24.hours.ago,
      include: Proc.new { |zone| DateTime.now.in_time_zone(zone).hour == 6 }
    }
  end.merge(name: period)
end
```
```
SendCatchUpJob.perform_later(:weekly)
```

gdpelican: The thrust being that you pass a single string into the job, which constructs a 'settings' object, which affects the behaviour of the job pretty granularly.

robguthrie: Thanks for this. I like it but I don't think I want to use it in this case. I think the next step is per user configuration for the granularity of the catchup, and that will remove the two jobs thing anyway.

PR #4709, opened 2018-06-05
fix locales with dialects and moment locales

### LINKED ISSUE #4569: Portuguese isn't appearing in app even when language is selected
Can confirm by setting locale to `pt_BR`; the app is 60% translated, but all of the strings remain in English. I can conform that they appear to come over the wire from `/api/v1/translations?lang=pt_BR` correctly.

I believe it's a broken translation from a recent locale update, or _perhaps_ we're handling the slightly-odd locale string (pt_BR) incorrectly when parsing the results from the server.

robguthrie:
fixes #4569

### DIFF
--- a/client/shared/interfaces/locale_records_interface.coffee
+++ b/client/shared/interfaces/locale_records_interface.coffee
@@ -14,6 +14,10 @@ module.exports = class LocaleRecordsInterface extends BaseRecordsInterface
       onSuccess: (response) ->
         if response.ok
           response.text().then (data) ->
-            eval(data)
+            # some, not all (why?), moment.js locale files have return
+            # statements on the end of them, but they are not wrapped in
+            # functions which raises an error.
+            # So we stip the return statement with this regex.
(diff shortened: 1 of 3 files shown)


client/shared/interfaces/locale_records_interface.coffee
```diff
@@ -14,6 +14,10 @@ module.exports = class LocaleRecordsInterface extends BaseRecordsInterface
       onSuccess: (response) ->
         if response.ok
           response.text().then (data) ->
-            eval(data)
+            # some, not all (why?), moment.js locale files have return
+            # statements on the end of them, but they are not wrapped in
+            # functions which raises an error.
+            # So we stip the return statement with this regex.
+            eval data.replace(/return.+\n\n$/, "\n")
```
gdpelican: I'd suggest doing this in client/tasks/shared.coffee, so we just have to do it once on build, rather than on every request

PR #4717, opened 2018-06-10
Delete own user account

robguthrie:


app/services/user_service.rb
```diff
@@ -1,4 +1,13 @@
 class UserService
+  def self.destroy(user:)
+    user.deactivate!
+    zombie = User.create(name: I18n.t(:'user.deleted_user'),
+                         email: "deleted-user-#{rand(10**10)}@example.com")
```
gdpelican: I'd reach for `SecureRandom.uuid` here first I think, I believe that's the most common 'give me something random that won't collide' mechanism in the codebase

client/angular/pages/profile_page/profile_page_controller.coffee
```diff
@@ -44,6 +44,18 @@ $controller = ($scope, $rootScope) ->
         fragment: 'deactivate_user'
       submit: -> ModalService.open 'ConfirmModal', confirm: confirm
 
+  @deleteUser = ->
+    if AbilityService.canDeactivateUser()
```
gdpelican: We'll want to clear this up some and perhaps move it to shared, but I'm happy with this for now. The code I'd want to write:

```
{ deleteConfirmation, deactivateConfirmation } = require 'shared/helpers/user'

@deactivateUser = ->
  ModalService.open 'ConfirmModal', confirm: deactivateConfirmation

@deleteUser = ->
  ModalService.open 'ConfirmModal', confirm: deleteConfirmation
```

client/shared/interfaces/user_records_interface.coffee
```diff
@@ -22,6 +22,9 @@ module.exports = class UserRecordsInterface extends BaseRecordsInterface
   deactivate: (user) =>
     @remote.post 'deactivate', user.serialize()
 
+  destroy: (user) =>
```
gdpelican: Seems like you don't need the user to destroy here, meaning we can avoid the slightly awkward
```
Records.users.destroy(Session.user()) # could just be Records.users.destroy()
```

app/services/user_service.rb
```diff
@@ -1,4 +1,13 @@
 class UserService
+  def self.destroy(user:)
+    user.deactivate!
+    zombie = User.create(name: I18n.t(:'user.deleted_user'),
```
gdpelican: Should the zombie be marked as deactivated?

app/services/user_service.rb
```diff
@@ -1,4 +1,13 @@
...
+                         email: "deleted-user-#{rand(10**10)}@example.com")
+    MigrateUserService.migrate!(source: user, destination: zombie)
+    user.reload.destroy
+    zombie
```
gdpelican: I don't believe we need to return the zombie; if we wanted to serialize it out we'd have to do
```
Events::UserDestroyed.publish!(user, zombie)
```
and have the zombie be the eventable for the event that's created.

(But I don't think we need that.)

gdpelican: Also, we should chuck a
```
EventBus.broadcast 'user_destroy', user, zombie
```
on there for completeness.

PR #4718, opened 2018-06-10
display rank for ranked_choice stance choices in catch up email

robguthrie:


app/views/poll_mailer/common/_poll_option.html.haml
```diff
@@ -4,6 +4,7 @@
 - show_link  = local_assigns[:link]
 - stance_choice = local_assigns.fetch(:stance_choice, nil)
 - show_score = poll.has_variable_score && stance_choice
+- rank_or_score = (poll.poll_type == 'ranked_choice') ? stance_choice.rank : stance_choice.score
```
robguthrie: @gdpelican I though about querying the poll template here.. or adding a rank_or_score method to stance_choice. Your thoughts?

gdpelican: Hrm, right. I guess I'd consider this first:
```
# stance_choice.rb
def score
  if poll.ranked_score # <- new poll template option
    poll.minimum_stance_choices - self[:score] +1
  else
    self[:score]
  end
end
```
and then if that seemed too crazy or broke stuff then doing `rank_or_score`
```
def rank
  poll.minimum_stance_choices - self.score +1 if poll.ranked_score
end

def rank_or_score
  rank || score
end
```

robguthrie: I'm against overloading simple accessor methods nowdays, particularly with something that could hit the database. rank_or_score seems right to me. thanks

PR #4719, opened 2018-06-10
dont use used invitations in emails

robguthrie:
I think this is a good idea. isn't it?

### DIFF
--- a/app/models/poll_email_info.rb
+++ b/app/models/poll_email_info.rb
@@ -84,7 +84,7 @@ class PollEmailInfo
   private
 
   def membership
-    @membership ||= poll.guest_group.memberships.find_by(user: recipient)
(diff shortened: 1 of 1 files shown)


app/models/poll_email_info.rb
```diff
@@ -84,7 +84,7 @@ def utm_hash(args = {})
   private
 
   def membership
-    @membership ||= poll.guest_group.memberships.find_by(user: recipient)
+    @membership ||= poll.guest_group.memberships.pending.find_by(user: recipient)
```
gdpelican: To my eye the only place we use poll_email_info#membership is in `target_url`; if they click on a link to a used membership it'll fly through `InvitationAlreadyUsed` and land on `membership.target_model` anyway, so I'm not sure if this introduces a change in behaviour.

robguthrie:
yea.. never mind

PR #4725, opened 2018-06-13
Thread comment responses by email

gdpelican:


gdpelican:
Have any thoughts on this one @robguthrie ?

robguthrie:
yea, I think I'm ok with this. Let's give it a go. Thanks.

robguthrie:
@gdpelican I think this broke reply to comment within the app!!

robguthrie:
Update, I think it's related to drafts.

I'm getting: {"errors":{"parent":["Needs to have same discussion id"]}}

On some threads where I already have a draft saved

PR #4727, opened 2018-06-18
allow self promote to coordinator when no coordinators

robguthrie:
rather than this big task of ensuring there are coordinators, and checks and rules.. let's just allow groups to have no coordinators, and if that is the case, a member can self promote to coordinator.


robguthrie:
@gdpelican any thoughts/comments?

gdpelican:
Looks (retroactively) good to me :+1:

Issue #4734, opened 2018-06-24
Unable to login with password if unverified user accounts exist


robguthrie:
Just had a report from a user that they could not login with a password, only with login token. I've tested this and it turns out that devise is not scoped to only select from email_verified accounts. Woha. how did we get away with that for so long?

PR #4735, opened 2018-06-24
fix for cannot login by password when unverified emails exist

robguthrie:
a fix for #4734 

@gdpelican can you double check this? I only want to apply it for sign_in's initiated by the user, so if this breaks signing_in an unverified user just after creation then it's no good.


gdpelican:
I don't think I have any idea whether this will break unverified sign in, but IIRC we have some automated testing around that path, right?

robguthrie:
Cool. I'm satisfied by spec/controllers/api/stances_controller_spec.rb:330

PR #4773, opened 2018-08-22
allow up to 500 chars in a stance statement.

robguthrie:
We've had a few people say the stance limit is too low recently. Going to give this a go and see what happens. In the future we make make stance limit more configurable, but it would be great if we don't have to.


gdpelican:
I think you need to change the client side validation too;
`poll_common_stance_form` maybe?

gdpelican:
poll_common_stance_reason.haml

robguthrie:
thanks, will do!

PR #4906, opened 2018-09-24
Fixed link syntax

briced:


robguthrie:
Thanks for contributing @briced. We've removed the documents from the repo. It all lives in the loomio/loomio-help repo now.

PR #4910, opened 2018-09-26
disable email login feature

robguthrie:


robguthrie:
@gdpelican when you get a chance (and feel like it), can you advise on how you would disable account registration and session creation via email/password  based on this flag? No rush for now

client/angular/components/auth/provider_form/auth_provider_form.haml
```diff
@@ -2,7 +2,6 @@
   .auth-provider-form__providers
     %button.md-button.md-raised{type: "button", class: "{{provider.name}}", ng-click: "select(provider)", ng-repeat: "provider in providers"}
       %i.mdi.mdi-24px{ng-if: "provider.name != 'saml'", class: "mdi-{{provider.name}}"}
-      %i.mdi.mdi-24px.mdi-microsoft{ng-if: "provider.name == 'saml'"}
       %span{translate: "auth_form.continue_with_provider", translate-value-provider: "{{capitalize(provider.name)}}"}>
       %span
-    %p.md-subhead.auth-email-form__or-enter-email{translate: "auth_form.or_enter_your_email"}
```
gdpelican: I guess this should probably be moved to `auth_email_form`

robguthrie: Would be nice. However, we only want to show it if there are login providers. So either email form needs to check for providers, or providers form needs to check for email... At this stage the CSS and spacing makes this the way to go.

PR #4913, opened 2018-09-29
update lodash

robguthrie:


client/shared/services/session.coffee
```diff
@@ -45,7 +45,7 @@ setDefaultParams = (params) ->
     Records[endpoint].remote.defaultParams = _.pick params, _.identity
```
lheckemann: This should become `pickBy`

gregorykan: thank you @lheckemann

PR #5018, opened 2018-11-22
allow configuring SMTP SSL verification

lheckemann:
Not verifying the SMTP server's certs is not a great idea, allowing a MITM to obtain auth credentials.


robguthrie:
Cool. Thanks! I've just added a fallback

lheckemann:
Thanks. The travis failure seems to be unrelated and caused by chargify as far as I can tell.

PR #5024, opened 2018-11-27
fix: add compilation step of client application

Pindar:
This missing step causes several issues on the client like not showing emojis and not properly working service worker.

See also https://www.loomio.org/d/Na2A3ifI/emojis-are-missing-or-not-loaded

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ WORKDIR /loomio/client
 RUN npm install -g yarn
 RUN yarn
 RUN npm rebuild node-sass
+RUN node_modules/gulp/bin/gulp.js compile
 WORKDIR /loomio
 
 EXPOSE 3000


s2-simond:
ref https://github.com/loomio/loomio-deploy/issues/35

robguthrie:
Sorry this is not the correct solution. 

The client is built on execute of docker_start.sh, which is necessary because the client includes ENV variables that are only available once the container is running.

Pindar:
I will check it. My docker settings are `resources: {"requests": {"cpu": "100m", "memory": "400Mi"}, "limits": {"memory": "1Gi"}}`. Is it too less?

Pindar:
@robguthrie thank you, it seems you found it. I change my values to `resources: {"requests": {"cpu": "100m", "memory": "800Mi"}, "limits": {"memory": "2Gi"}}` and it worked. I keep monitoring and will update later once more.

s2-simond:
Maybe you can add some debugging echo output to https://github.com/loomio/loomio/blob/master/docker_start.sh#L7 so it's visible in the logs that the compiling happens. At the moment it's unclear and I always have to exec into the container. Just saying.

Pindar:
Maybe you can add some debugging echo output to https://github.com/loomio/loomio/blob/master/docker_start.sh#L7 so it's visible in the logs that the compiling happens. At the moment it's unclear and I always have to exec into the container. Just saying.

PR #5029, opened 2018-11-28
chore: add exit option to bash

Pindar:
Exit immediately when one command fails on startup. This should make it more easy to catch memory issues or any other kind of configuration problems on startup.


robguthrie:
Thanks!

PR #5046, opened 2018-12-04
Delay deletes

robguthrie:


robguthrie:
@gregorykan can you please deploy to loomiotest and manually check the delete behaviour is acceptable? we may need a flash notice that delete has been queued.

PR #5146, opened 2019-01-15
[WIP] Webhook integration for Microsoft Teams

gdpelican:
Loomio <> Microsoft Teams!

Watch the [explainer video](https://www.useloom.com/share/33b703f90bc94e0bb33066f5baf6f79b) for a demo!

-------
captura de pantalla 2019-01-15 a la s 7 03 41 p m
captura de pantalla 2019-01-15 a la s 7 03 44 p m
captura de pantalla 2019-01-15 a la s 7 04 06 p m
------
captura de pantalla 2019-01-15 a la s 8 00 37 p m
captura de pantalla 2019-01-15 a la s 7 00 52 p m
captura de pantalla 2019-01-15 a la s 6 57 20 p m


app/helpers/pretty_url_helper.rb
```diff
@@ -49,4 +49,11 @@ def polymorphic_title(model)
     end
   end
 
+  def user_avatar_url(user)
+    [
+      root_url(default_url_options.merge(trailing_slash: false)),
```
gdpelican: Note to self: need to account for remote avatar urls

app/extras/clients/base.rb
```diff
@@ -107,11 +107,12 @@ def post_content!(event)
   end
 
   def serialized_event(event)
-    begin
-      "#{self.class.name.demodulize}::#{event.kind.classify}Serializer".constantize
-    rescue NameError
-      "#{self.class.name.demodulize}::BaseSerializer".constantize
-    end.new(event, root: false).as_json
+    serializer = [
+      "#{self.class.name.demodulize}::#{event.kind.classify}Serializer",
```
gdpelican: This is a imo slightly better approach to this, which cascades through specific to general class names. So, if we're doing a `poll_expired` event, we start with
```
Microsoft::PollExpiredSerializer
```
then, if we can't find that, we look for the eventable type
```
Microsoft::PollSerializer
```
and, failing those, the base serializer
```
Microsoft::BaseSerializer
```

This allows us to be super flexible in specifying the behaviour for each event, without a bunch of empty classes that don't do much.

gdpelican:
^^ @robguthrie This one is ready for a look at your leisure. :)

robguthrie:
Great stuff.

It's not going to be great to have a card per possible integration, so how about this design?

a single "Integrations" card, like this:
image

and then the new integration modal might look like this:
image

Here's the figma link if you want to play with it.
https://www.figma.com/file/IPHrPHaJAv6SpfRFxEGGW2Ho/Untitled?node-id=1%3A14

robguthrie:
Also, have you added support for comments and votes too?

gdpelican:
Have you watched the explainer video! I cover some of this stuff there.

robguthrie:
Whoops! no! Will watch soon.

robguthrie:
Thanks for the great video, which I've now watched. (Sorry!)

Please put the help into the loomio/loomio_help, not into the marketing, it won't work with the new marketing site going live at the end of the month.

We've been asked to enable votes and comments, but I hear you on the issue with that, so please make it configurable with checkboxes + labels for the event kinds - that would be fantastic.

Thanks for the great and prompt work. Love it.

gdpelican:
Cool, I'll push on that over the next couple days.

Maybe we can scope the 'integrations' card as part of the Telegram work when/if that eventuates? That feels like it would take care of the UI duplication just before it starts to get uncomfortable.

gdpelican:
Alright, the granular event sourcing looks like it pretty much works; here's a screenshot:
captura de pantalla 2019-01-17 a la s 9 39 42 p m

I should probably put in the ability to edit the integration now (to change the event kinds that respond, rather than needing to destroy and recreate it to change the events), but won't have time to do that this week.

robguthrie:
Wowzers. That's great!

Do you think this could be beta tested on a private host as is?

gdpelican:
Should be able to, yep. Have a go at flicking the 'FEATURES_ENABLED_MICROSOFT_TEAMS' env on to try it out.

app/extras/clients/base.rb
robguthrie: This is nice, but I think that I'd rather throw an exception if we don't have the correct serializer for the event.

robguthrie:
I've merged this in, thanks @gdpelican. Any changes can be on a new branch.

robguthrie:
uh.. I've made a mess here, realised I don't want to deal with new slack code on master just yet.

PR #5201, opened 2019-02-05
Add standalone boolean to method call

simonv3:
It was missing from the example

```ruby
    def use_static_asset(path, filename, standalone: false)
      @static_assets.add StaticAsset.new(path_prefix(path), filename, standalone)
    end
```


simonv3:
^ feel like the tests are failing with something that isn't my fault :P

simonv3:
After comments on #5202 it sounds like `use_static_asset` shouldn't be used anymore.

PR #5202, opened 2019-02-05
Add reference to application.css in head

simonv3:
I couldn't find where the application.scss.erb file was being loaded in the front-end, so it's very possible that I'm missing something here, but to be able to load assets in plugins [as described here](https://github.com/loomio/loomio/blob/master/lib/plugins/README.md#add-assets-to-the-asset-pipeline) the [application.scss.erb](https://github.com/loomio/loomio/blob/master/app/assets/stylesheets/application.scss.erb#L1) loads things from [the repository in plugins](https://github.com/loomio/loomio/blob/master/lib/plugins/repository.rb#L43), but if that file isn't loaded, than the assets aren't loaded either. 

So, maybe something changed in how this should be done, but this seemed like a fairly straightforward fix to me. 

One alternative that I thought of doing was creating a component with just a `.scss` file, but that seemed like overkill.

### DIFF
--- a/app/views/application/index.html.haml
+++ b/app/views/application/index.html.haml
@@ -18,6 +18,8 @@
     %link{rel: "apple-touch-icon", href: AppConfig.theme[:icon_src]}
 
     %link{rel: "stylesheet", href: client_asset_path(:"angular.min.css")}
+    %link{rel: "stylesheet", href: asset_path(:"application.css")}
+
     %link{rel: 'manifest',   href: '/manifest.json'}
 
     = action_cable_meta_tag


robguthrie:
Does this fix a problem for you? 

I would have thought that styles from plugins end up in the client stylesheet file. 

We have 2 asset sets, one from rails and one from the client. We've almost reduced the rails stuff to nothing - just the most basic layout for error pages etc.  

So I'm kind of surprised to see this is necessary. Maybe there is another command to add your styles to the client stylesheet bundle?

PR #5231, opened 2019-02-21
Add score poll title for decisions card

gdpelican:
Looks like the decision tools card is missing a score translation!

Repro: visit https://www.loomio.org/p/new on loomio.org:
captura de pantalla 2019-02-21 a la s 16 43 58


robguthrie:
Thanks!

PR #5312, opened 2019-04-09
Vue cli

robguthrie:


buoyantair:
Looks like we're shifting from Angular to Vue? :O

robguthrie:
Ya, we're quite far along now. Angular 1 and Angular Material are sundowned projects and we felt that we needed to be using a thriving library and toolkit to take Loomio forward. 

Once this is feature complete we'll return to improving the project functionality - we cannot wait.

PR #5373, opened 2019-04-18
make becoming a coordinator more obvious

robguthrie:


robguthrie:
@gregorykan note we need to port this to the vue client when we get a moment.

PR #5374, opened 2019-04-18
groups can set their own handles

robguthrie:


robguthrie:
need vue form adjustment
also what are valid handle chars?

robguthrie:
if parent changes handle, update subgroups. use parameterize for default

robguthrie:
integrated into #5391

PR #5382, opened 2019-04-18
Duplicate property

MatheusRV:
Listing the same property twice in one object literal is redundant and may indicate a copy-paste mistake.

In ECMAScript 2015 and above, as well as ECMAScript 5 non-strict mode, an object literal may define the same property multiple times, with later definitions overwriting earlier ones. If all definitions assign the same value to the property, this will not to lead to problems at runtime, but it makes the code harder to read and maintain.


robguthrie:
Thanks!

PR #5485, opened 2019-05-24
Replace discriminate with setting inheritance_column

mvz:
This should work, right?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,6 @@ gem 'premailer-rails'
 gem 'griddler', github: 'loomio/griddler'
 gem "griddler-mailin", github: 'loomio/griddler-mailin'
 gem 'activerecord-import', '1.0.1'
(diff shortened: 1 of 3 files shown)


robguthrie:
Yea, I believe it should. Try using
 
```
self.inheritance_column= :identity_type
```

and we'll see if the test suite passes

mvz:
This is as far as I got. The big difference between `discriminator` and Rails' basic STI is that Rails stores the class name (e.g., `Slack` or `Identities::Slack`, and `discriminator` handles storing just `slack` with a lower-case `s`.

robguthrie:
What's the motivation for removing discriminator?

mvz:
I didn't understand how it was different from regular Rails STI. It still seems too tiny to be a separate gem, but I guess at least I understand what it does.

PR #5773, opened 2019-08-15
Feature/choose photo on sign up

robguthrie:


robguthrie:
@gregorykan can you pick this up this morning? I think adopting the welcome carousel approach would be good. Make a user_setup_wizard with a "contrags you're a new user" page (which may include the orientation video), then another page for button for setting profile image.

PR #5781, opened 2019-08-20
Feature/new org flow

gregorykan:


config/locales/client.en.yml
```diff
@@ -747,13 +762,13 @@ en:
     permissions_explaination: 'There are two levels of membership: admin and member. Admins have all permissions. Select member permissions:'
     aria_label: 'Group form'
     new_organisation:
-    start_group_heading: 'Start a group'
-    start_subgroup_heading: 'Start a subgroup'
-    edit_group_heading: 'Edit group settings'
-    edit_subgroup_heading: 'Edit subgroup settings'
-    group_name: 'Group Name'
```
robguthrie: I think keep group name as group name, rather than name?

gregorykan: done

PR #5876, opened 2019-09-14
[WIP] Self service saml

gdpelican:


robguthrie:
Thanks for your work on this James. We've gone ahead with a different implementation (deploying tomorrow probably) but this was a useful reference.

PR #5885, opened 2019-09-17
chore(doorkeeper): get rid of all doorkeeper and doorkeeper-related s…

gregorykan:
…tuff


patcon:
Seems that doorkeeper/oauth2 apps were what allowed the `access_token` API usage to be a thing, no?

re:
https://github.com/loomio/loomio-oauth-sample/blob/master/README.md
https://help.loomio.org/en/dev_manual/using_the_api/#using-the-public-api

Since doorkeeper is removed, does that mean that the API is no longer usable as documented? If so, should I create a ticket for updating the docs to clarify that?

(I've spent the last 4+ hours trying to get the API working, including digging around in DB for auth tokens etc, and would really like to save the next person the time if the docs happen to be obsolete :) )

robguthrie:
We were gettting a lot of serious security warnings. Without contact with anyone actually using the API, I felt I could not upgrade the system successfully.

If you want to reintroduce doorkeeper, and give me feedback on the API and docks, I'll support you with that. 

We've never had any feedback on this before, so please feel welcome to speak up and let me know what you need.

patcon:
Thanks a ton for being open to it Rob :)

The use-case would be a relatively small one, so I'll try to make time to investigate how simple it would be to bring back

patcon:
If I instead added simple per-user API token auth, would that also be acceptable? I suspect that would have smaller footprint...?

robguthrie:
I think that would be fine. Is this for your own server or for loomio.org? If your own server, then you could just populate the api_token yourself via console and not worry about UI or anything.

Then if you:
- create an api_token column on users table
- insert some lookup logic into a `def current_user` method in api/restful_controller (see api/polls_controller for an example)

that should do the trick, I suggest you pass user_id and api_token so we don't need to index on api_token, which makes this even lighter

Is that enough to get you started?

patcon:
Rob, you are wonderful. Thank you. This sounds like enough to experiment with our own instance, and see if it creates any value :) If it really transforms how we work, we'll no doubt work to upstream the ability later. Thank you!

PR #5933, opened 2019-10-03
lessons learned from memperf

robguthrie:


robguthrie:
that time-ago widget was bringing the whole app to it's knees

PR #6031, opened 2019-10-30
Fix typo in proposal outcome panel

davideg:
Hello :wave:
Noticed this typo and figured I'd make a quick fix :wink: rather than just reporting it.
image


robguthrie:
Thank you!

PR #6121, opened 2019-12-01
Fix/group files panel

gregorykan:


db/migrate/20191201225942_ensure_group_ids_for_active_storage_attachments.rb
```diff
@@ -0,0 +1,7 @@
+class EnsureGroupIdsForActiveStorageAttachments < ActiveRecord::Migration[5.2]
+  def change
+    ActiveStorage::Attachment.where(group_id: nil).find_each do |attachment|
+      attachment.update(group_id: attachment.record.group.id) if attachment.record && attachment.record.group
```
robguthrie: lets call attachment.update_attachments! instead

app/serializers/attachment_serializer.rb
```diff
@@ -6,6 +6,7 @@ class AttachmentSerializer < ActiveModel::Serializer
   has_one :author, serializer: UserSerializer, root: :users
 
   def author
+    # TODO: if object.record is a group, there is no author
```
robguthrie: remove comment as we've aliased to creator

PR #6142, opened 2019-12-09
Sanitize group description text from  HTML or Markdown

### LINKED ISSUE #6140: Raw HTML and Markdown strings in explore page group descriptions
Possible duplicate of #3033, but on the current explore page it looks like multiple group descriptions are displaying as raw HTML or Markdown strings:

Screen Shot 2019-12-09 at 9 30 32 AM

Screen Shot 2019-12-09 at 9 30 39 AM

It looks like this was initially fixed in #3801, but might have been lost in later changes. If you want to address it the same way as in that PR I'm interested in putting in a PR that checks `discussion_format` and sanitizes the input based on whether it's HTML or Markdown

pjsier:
Closes #6140. Cleans up group description from HTML or Markdown based on `description_format` field. Loofah is added to the Gemfile, but it's already a sub-dependency so this isn't really adding anything new.

I'm using Loofah instead of [`strip_tags`](https://apidock.com/rails/ActionView/Helpers/SanitizeHelper/strip_tags) because its `to_text` method adds whitespace between tags for clarity. With `strip_tags` this HTML:

```html
<p>Line one</p><p>Line two</p>
``` 

becomes "LineoneLinetwo", while Loofah will return "Line one Line two". Let me know if there's anything I should fix!

### DIFF
--- a/vue/src/components/explore/page.vue
+++ b/vue/src/components/explore/page.vue
@@ -6,6 +6,7 @@ import UrlFor    from '@/mixins/url_for'
 import _truncate from 'lodash/truncate'
 import _map      from 'lodash/map'
 import _sortBy   from 'lodash/sortBy'
+import marked    from 'marked'
 
 import { debounce } from 'lodash'
 
@@ -45,7 +46,12 @@ export default
       { 'background-image': "url(#{group.coverUrl('small')})" }
 
     groupDescription: (group) ->
-      _truncate group.description, {length: 100} if group.description
+      description = group.description || ''
+      if group.descriptionFormat == 'md'
+        description = marked(description)
+      parser = new DOMParser()
+      doc = parser.parseFromString(description, 'text/html')
+      _truncate doc.body.textContent, {length: 100}
   computed:
     showMessage: ->
       !@searching &&


pjsier:
Not sure why some plugin tests are failing here, but it looks like they're unrelated

robguthrie:
Wonderful. However, I'm hesitant to add another gem, is it possible to also
strip tags with nokigiri?

pjsier:
Makes sense, thanks for the suggestion! Looks like there's a straightforward way in nokogiri so I've updated that here

robguthrie:
Oh! I've just realised that this is going to sanitise all group
descriptions throughout the app (ie: on the group page), rendering the
formatting pointless.

Otherwise it's a decent stab at the implementation.

Sorry to not have read the PR sooner, but I can't accept this, the process
needs to happen in the explore_page.vue component on the client.

Given that, I would recommend the following approach:

Have a look at the suggestedTitle method in event_model.coffee. In there is
some code that strips html with the browser by rendering it to a dom node
that is not part of the page.

Start with that, then once that's working for html descriptions, you can
add the markdown rendering code from vue/src/marked.coffee which can render
markdown to html.

Then put those steps in the other order, so if the description is markdown,
render it to an html string, and then, either way you have an html string,
so you can then strip the tags and render it as a normal string in the
template.

Again, I'm sorry for not paying attention from the beginning, I hope you're
enjoying the learning process anyway.

Rob


If the text is markdown then render it to html

pjsier:
No problem, thanks for the review! It's been interesting seeing how things are set up. This update should implement it the way you're describing now

robguthrie:
Hey this is great. Really pleased to have your help and happy to merge this. Thank you!

PR #6151, opened 2019-12-10
per group saml authentication provider

robguthrie:


gregorykan:
https://www.portalguard.com/blog/2016/06/20/saml-single-logout-need-to-know/

PR #6215, opened 2020-01-18
Saml provider upgrade #3

robguthrie:


robguthrie:
this has been pulled apart and applied more sanely

Issue #6260, opened 2020-01-30
[feature request] Edit source in markdown editor


EliotBerriot:
## What is the issue I'm facing?

The new UI of Loomio is shipped with a new WYSIWYG markdown editor.

While I tend to dislike WYSIWYG editors because they sometimes have an unexpected behaviour when you want to perform advanced (or not so advanced) text edition, I understand this is mostly a matter of personal preference, so this is not what I want to discuss here.

However, being a heavy Loomio user for more than a year, there is one crucial things I was able to do with the previous editor I cannot do now. As part of my workflow, I do a lot of collaborative edition on CodiMD (an online markdown editor). When the writing/review process is done, I used to copy-paste the markdown text in a Loomio post for publication and voilà!

This is not working anymore, as shown on the video below. My understanding is that the new editor interprets the markup as it's copy-pasted, yielding unpredictable results, and removing most of the structure (e.g titles and tables).

The only workaround I found was to copy the source as HTML and paste it in Loomio's `contenteditable` tag in the editor, which is slow, error prone, and inaccessible to most people.

## Demo of the issue

![loomio-copy-paste](https://user-images.githubusercontent.com/1970915/73435780-946bfc80-4349-11ea-88f3-f95587eae368.gif)

## What is my suggested solution?

I'd like to have a "see/edit source" button in the editor toolbar. Cliking on this button would switch the editor in a different mode, and let you edit the markdown source directly, without any rendering.

This would solve the specific issue  I have (copy-pasting markdown content from a third-party source), but also help with other annoying issues, when one of the the WYSIWYG editor bugs prevent you from doing something (editing a link, or code block, quote, etc).

It would also give an easy way to copy a message or a part of a message from Loomio to somewhere else.

This is not strictly required, but it would be a nice bonus if the editor could remember what mode it's in, so I don't have to switch it for every new comment or thread. I really don't need the preview.

Thank you for your consideration, let me know if I can clarify anything.

robguthrie:
I'm happy to provide a markdown mode for the editor. The wysiwyg mode is actually HTML, and supports features markdown does not, but conversion between the two formats is possible, and we can indeed add a MD button that gives you such an experience.

This just comes down to prioritization

robguthrie:
@EliotBerriot I've made a start on this feature.. i'll keep pushing it along when I have the motivation

robguthrie:
done!

PR #6264, opened 2020-02-01
support switching to markdown and back

robguthrie:
not ready, but works.

- this imports the showdown converter on the fly, so it does not increase the bundle size
- we should save the user perference for html or markdown when they switch
- we will likely want to support @mention autocomplete and file attachments
- it seems that converting html to markdown does not quite get images right, they need arrow brakets removed, need to look into it.

Screen Recording 2020-02-01 at 9 05 34 PM

Starts to address #6260


EliotBerriot:
Thank you for working on this @robguthrie !

robguthrie:
@EliotBerriot I really got stuck into it in the end. I've deployed it now and I'd love your initial feedback and any bug reports.

Try mentioning, attaching files with the button, dragging and dropping, and pasting attachments too please.

EliotBerriot:
Just deployed it on https://governance.funkwhale.audio, and seems to work as expected!

I've tried the following with success:

- Mentionning a username
- Attaching files with the button
- Drag-n-dropping a picture in the textarea
- Copy pasting a screenshot in the textarea

The switch to/from markdown works as expected, and is saved as well.

Thank you @robguthrie for your hard work on this, I really appreciate it. I'll keep you updated if I find anything unusual during my daily usage.

PR #6281, opened 2020-02-05
add a general chat thread on group startup

gregorykan:


robguthrie:
John is working on some new content for this.

PR #6316, opened 2020-02-18
Self merge user accounts

gregorykan:


app/controllers/merge_users_controller.rb
```diff
@@ -0,0 +1,27 @@
+class MergeUsersController < ApplicationController
+  layout 'basic'
+
+  def confirm
+    @source_user = User.find_by!(id: params[:source_id])
+    @target_user = User.find_by!(id: params[:target_id])
+    @hash = params[:hash]
+    if UserService.validate_account_merge_hash(source_user: @source_user, target_user: @target_user, hash: @hash)
```
robguthrie: target and source cannot be same

vue/src/components/profile/page.vue
```diff
@@ -52,10 +60,36 @@ export default
...
+            submit: sendVerification
+            text:
+              title:    'merge_accounts.modal.title'
+              raw_helptext: @$t('merge_accounts.modal.helptext', sourceEmail: @originalUser.email, targetEmail: @user.email)
+              submit:   'merge_accounts.modal.submit'
+              flash:    'merge_accounts.modal.flash'
+
+    checkEmailExistence: ->
+      return if @originalUser.email == @user.email
+      Records.users.checkEmailExistence(@user.email).then (res) =>
```
robguthrie: isDisabled always == emailExists.. and rather than data can be a computed

app/services/user_service.rb
```diff
@@ -89,4 +89,24 @@ def self.save_experience(user:, actor:, params:)
     user.experienced!(params[:experience], !params[:remove_experience])
     EventBus.broadcast('user_save_experience', user, actor, params)
   end
+
+  def self.send_merge_verification_email(user:, actor:, target_email:)
```
robguthrie: only actor, no user since they are always the same

vue/src/components/profile/page.vue
```diff
@@ -31,13 +34,18 @@ export default
...
-        collections: ['users']
-        query: =>
-          @originalUser = Session.user()
-          @user = @originalUser.clone()
-          Session.updateLocale(@user.locale)
+      @originalUser = Session.user()
+      @user = @originalUser.clone()
+      Session.updateLocale(@user.locale)
+
+      # @watchRecords
```
robguthrie: delete this

PR #6436, opened 2020-04-03
Add support for using nextcloud as identity provider

AlbanBedel:
Loomio must be registered in nextcloud as oauth 2.0 client using https://loomio.example.com/nextcloud/authorize as redirection URL.

In loomio the NEXTCLOUD_HOST environment variable must point to the nextcloud instance, for example https://nextcloud.example.com. NEXTCLOUD_APP_KEY and NEXTCLOUD_APP_SECRET must be set to the client identifier and secret set by nextcloud.


robguthrie:
I've added what you wrote to the loomio-deploy readme. Thank you.

PR #6450, opened 2020-04-07
Sync german client language file; add missing translations

Sefux:
This PR includes some missing german translations for the new vuejs client. It was not clear to me if the old translation workflow is still in use? The translation guide could not be found, because the link on CONTRIBUTING.md is broken.

### DIFF
--- a/config/locales/client.de.yml
+++ b/config/locales/client.de.yml
@@ -22,6 +22,22 @@ de:
     yearly: Jährlich
   current_plan_button:
     upgrade: Upgrade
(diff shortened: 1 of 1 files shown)


robguthrie:
Thanks for doing this. Yes, our main translation workflow is still with Transifex - it's actually teated as the master source for translations, so after I merge this I'll reverse them back into Transifex. 

It's often better to use that service than doing it manually... however you seem to have managed well.

I've updated the CONTRIBUTING.md file just now too.

Thanks again.

PR #6527, opened 2020-05-04
documents can be deleted

gregorykan:
- attachments and documents list have same style
- documents can be deleted from group page
- clicking delete icon for attachment opens group settings modal, from which attachment can be deleted

### DIFF
--- a/vue/src/components/document/list.vue
+++ b/vue/src/components/document/list.vue
@@ -66,7 +66,7 @@ section.document-list
         v-btn.ml-2(v-if="canDelete" icon :aria-label="$t('common.action.delete')" @click='deleteDocument(document)')
           v-icon(size="medium") mdi-delete
       v-layout.attachment-list__item-details
-        v-icon {{ `mdi-${document.icon}` }}
+        v-icon.mr-2 {{ `mdi-${document.icon}` }}
         a.document-list__title(:href='document.url' target='_blank') {{ document.title }}
         v-btn.ml-2(v-if="canDelete" icon :aria-label="$t('common.action.delete')" @click='deleteDocument(document)')
           v-icon(size="medium") mdi-delete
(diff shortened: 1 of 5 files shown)


vue/src/components/document/list.vue
```diff
@@ -34,56 +34,60 @@ export default
...
       EventBus.$emit 'initializeDocument', doc, $mdMenu
 
+    deleteDocument: (document) ->
+      EventBus.$emit 'openModal',
+        component: 'ConfirmModal'
+        props:
+          confirm:
+            submit: document.destroy
+            text:
+              title:    'attachments.confirm.title'
```
robguthrie, line 44: This looks great, except we already have strings for this stuff, lets not add any new ones. I'm trying to reduce the number of strings we have to manage in loomio.

comment_form.attachments.remove_attachment
poll_common_delete_modal.question
poll_common_delete_modal.success

should get you through.

Thank you very much!

vue/src/components/thread/attachment_list.vue
```diff
@@ -1,25 +1,40 @@
...
       img.attachment-list__preview(:src="attachment.preview_url")
     .attachment-list__item-details
-      v-icon(v-if="attachment.icon") {{ `mdi-${attachment.icon}` }}
+      v-icon.mr-2(v-if="attachment.icon") {{ `mdi-${attachment.icon}` }}
       div
         a(:href="attachment.download_url")
           span {{ attachment.filename }}
         space
         span.lmo-grey-on-white ({{ prettifyBytes(attachment.byte_size) }})
+        v-btn.ml-2(v-if="group && canDelete" icon :aria-label="$t('common.action.delete')" @click='deleteDocument')
```
robguthrie, line 35: why do we need a group for this? If this could work wherever there are document_lists, and did not need a group modal, that would be better

gregorykan: ah yes i know what i was trying to do. will remove, unnecessary

PR #6545, opened 2020-05-13
Tweak smtp config to omit auth config when not setting SMTP_AUTH

piemonkey:
Previously not setting the SMTP_AUTH env var still inserted the
:authentication arg into the config making it invalid.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -103,16 +103,26 @@ module Loomio
 
     if ENV['SMTP_SERVER']
       config.action_mailer.delivery_method = :smtp
-      config.action_mailer.smtp_settings = {
-        address: ENV['SMTP_SERVER'],
-        port: ENV['SMTP_PORT'],
-        authentication: ENV['SMTP_AUTH'],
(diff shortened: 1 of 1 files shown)


robguthrie:
Hi @piemonkey. Sorrry for the delay in reviewing this. Thanks for contributing.

However, after looking at this, I don't see why it is necessary - Any values that are not set will be removed, because the hash is `compact`ed.

So if you don't want to have authentication, you can just not set those variables.

Am I missing something here?

piemonkey:
Hmm, that's a good point, I'm not that familiar with Ruby so didn't spot that. I've just tested and it looks like this does indeed work, but each of the env vars `SMTP_AUTH`, `SMTP_USERNAME` and `SMTP_PASSWORD` need to be removed and they can't just be set to an empty string. My version behaves more as a user might expect but I won't be offended if you choose not to take it. It'd be nice to add this sort of thing to some documentation though.

robguthrie:
Right, just so you know, blank strings are considered truthy values in Ruby, so just delete or comment out the lines.

PR #6570, opened 2020-05-21
fix anonymous

robguthrie:
prevent anonymous polls from being deanonymized
remove user_ids from stances after close (they're there while poll is open so people can vote and change vote easily)
remove feature to deanonymize after close, to be replaced with "hide results while open"

TODO: database migration to remove user_id from stances and their created_events for existing anonymous polls and enable export of anonymous polls that are closed


app/services/poll_service.rb
```diff
@@ -81,6 +81,11 @@ def self.expire_lapsed_polls
 
   def self.do_closing_work(poll:)
     poll.update(closed_at: Time.now) unless poll.closed_at.present?
+    if poll.anonymous
+      poll.stances.update_all(participant_id: nil)
+      Event.where(kind: "stance_created", eventable_id: poll.stance_ids).update_all(user_id: nil)
```
robguthrie, line 86: can remove this line

app/views/poll_mailer/common/render_or_rescue.html.erb
```diff
@@ -0,0 +1,8 @@
+<%
```
robguthrie, line 1: take this out?

PR #6601, opened 2020-06-01
fix thread title disappearing after translated

gregorykan:
…d format column; in these situations, fall back to html

### DIFF
--- a/vue/src/components/common/translation.vue
+++ b/vue/src/components/common/translation.vue
@@ -8,7 +8,13 @@ export default
     isHtml: -> @format == 'html'
     translated: ->
       @model.translation[@field]
-    format: -> @model[@field+"Format"]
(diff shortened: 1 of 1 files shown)


robguthrie:
Hi @gregorykan.. title is not formatted &does not support any formatting. I'm wondering why it's going through this.

gregorykan:
> Hi @gregorykan.. title is not formatted &does not support any formatting. I'm wondering why it's going through this.

we're talking about the translation component, here being used in context-panel:

```
    span(v-if='!discussion.translation.title') {{discussion.title}}
    span(v-if='discussion.translation.title')
      translation(:model='discussion', field='title')
```

gregorykan:
oops didn't tag you @robguthrie

PR #6656, opened 2020-06-16
Discard polls

gregorykan:
- [x] poll page?
- [x] poll scopes
- [x] discard poll on thread page with stances?
- [x] bring back destroy action (in poll service but don't expose), make discard its own distinct action in client
- [x] update all stances, all stance created_events to have discussion_id nil
- [x] update poll created_event child count to 0 manually
- [x] send back poll created_event in response
- [x] broadcast poll created_event via group channel
- [x] change translation for confirm modal


robguthrie:
check scopes

PR #6755, opened 2020-07-15
Dont delete discussions

gregorykan:


app/services/discussion_service.rb
```diff
@@ -38,7 +38,6 @@ def self.announce(discussion:, actor:, params:)
   def self.destroy(discussion:, actor:)
```
robguthrie, line 38: new service action self.discard
route DELETE discard
and use base_model discard action within client.
client's discussion_service discard

PR #6854, opened 2020-08-13
Realtime collaborative edting and automatic save drafts

robguthrie:


app/extras/model_locator.rb
```diff
@@ -5,10 +5,14 @@ def locate
     if model.to_sym == :user
       resource_class.verified.find_by(username: params[:id] || params[:username]) || resource_class.friendly.find(params[:id] || params[:user_id])
     elsif model.to_sym == :group
-      resource_class.where.not(handle: nil).find_by(handle: params[:id]) || key_or_id && resource_class.friendly.find(key_or_id)
+      (model_key && resource_class.find_by(key: model_key)) ||
+      (model_id && resource_class.find_by(id: model_id)) ||
+      resource_class.where.not(handle: nil).find_by(handle: params[:id])
     elsif resource_class.respond_to?(:friendly)
+      key_or_id = model_key ? model_key : model_id
```
gregorykan, line 12: these last 2 conditional branches aren't quite right. pretty sure the remaining failing specs are due to this. the changes here were my attempt to fix it.

previously they still referenced `key_or_id` which no longer exists.

app/extras/model_locator.rb
```diff
@@ -5,10 +5,14 @@ def locate
...
       resource_class.friendly.find key_or_id
     else
+      key_or_id = model_key ? model_key : model_id
```
gregorykan, line 15: same as above

PR #6909, opened 2020-09-08
track get_started events from frontpage

gregorykan:


robguthrie:
uhnmm lets chat about this, I think it's probably ok but lets discuss a long term, pollution free approach to metrics tomorrow

PR #7155, opened 2020-12-07
Fix member permissions

robguthrie:
poll/members
  convert to user based √
  change stances controller to work with user_id, poll_id √
  change front end to work with users rather than stances √
  fix notifications count when inviting everyone in thread to poll √
  test
    group invite √
    discussion invite √
    poll invite √
    outcome invite
    discusion, poll edit

poll/reminder_form

  move edit proposal, close proposal to dropdown menu √

  poll/reminder_form
    add people, audiences √
    enter a message √
    click to see sent notifications (update message about opened) √

  new event poll_reminder √
    new mailer √
    add to server, client √
    copy from poll edited/poll announced √
    does not have discussion id - not in thread √
    e2e ..

  we dont want guests/emails in reminder autocomplete? √

anonymous poll fixes
  disable reopen anonymous poll √
  poll votes panel, only show cast stances..

hide results until you've voted √


robguthrie:
fixes #7100

PR #7249, opened 2021-01-14
Allow 'none' smtp authentication

msander:
To use smtp without authentication, we need to set "authentication" to nil. 
I copied the solution from other projects:
https://github.com/huginn/huginn/pull/1942/files
https://github.com/tootsuite/mastodon/pull/1597/files

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -106,7 +106,7 @@ module Loomio
       config.action_mailer.smtp_settings = {
         address: ENV['SMTP_SERVER'],
         port: ENV['SMTP_PORT'],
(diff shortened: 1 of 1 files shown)


robguthrie:
so setting the var to 'none' results in a nil value for the setting. Why not just not set the config var? (so it's nil to begin with)

msander:
You are right, of course. Please excuse my quick conclusion, my mistake was that I had set SMTP_USE_SSL=0.

robguthrie:
No worries at all. Thanks for your attempt to contribute!

PR #7745, opened 2021-08-01
date polls : display results before details

ladefoule:
Displaying the results of the poll before the voting details in case there are many votes.

sondage-date-le-rim


robguthrie:
Hi @ladefoule - thanks for the PR. 

I think it's taken me a minute to understand what you've done here... but I agree it's sensible.

However, I was expecting that your screenshot would have the totals in view.. am I missing something?

robguthrie:
I've implemented this UI change in the next release. Thanks for bringing it to my attention

PR #7873, opened 2021-09-28
Fix discussion_id type in API documentation

mashton:


robguthrie:
Oh whoops! 

Thank you for taking the time to fix this, @mashton. So nice to have people helping look after things.

PR #8364, opened 2022-02-01
fix bug where a broadcast stance would make guests uninvited.

robguthrie:


robguthrie:
this will require an update to the channels server

PR #8464, opened 2022-03-16
importing gives records new ids, then updates references

robguthrie:
the breakthrough in migration: now we can support data migration from any server to any server!


robguthrie:
GroupExportService 
now export provides a list of attachments and urls to fetch them. 
and import will rewrite the ids of the incoming data. much better.. means we can export/import from any server to any other server, like you might expect.

Issue #8579, opened 2022-05-17
Avoid relying on record :id being ordered chronologically in development


asdofindia:
So, the nasty bug that was fixed in [v2.13.0](https://github.com/loomio/loomio/releases/tag/v2.13.0) - [we](https://codema.in/) were affected by that.

We exported data from loomio.org in 2020 March with the high IDs - and those were imported to a fresh database with id sequences at low numbers.

We didn't reach collisions in normal use, but only when trying to run migrations. So we avoided migrations for a long time. But eventually we figured out a workaround - 

For each table with serial id, do:

```
> select id from table order by id desc limit 1;

id
---
44056

```

to find the highest ID (which would have come from the import). And then reset the sequence to a value above that

```
> select setval('table_id_seq', 44057, false);
```

This must work, but if there is any assumption being made in the codebase that ids are ordered chronologically, it will cause problems.

So, I request loomio developers to avoid using ids for ordering chronologically (or any similar logic).

robguthrie:
Thanks for letting me know. I'm sorry you had to work around the issues with that process. I'm so pleased it's fixed now.

I am also running a couple of servers which have the same problem/situation.

Yes, I'll keep this in mind. If you find any places where this is a problem, please let me know and I'll address them

PR #8589, opened 2022-05-20
Use created_at to order notifications

asdofindia:
See #8579 for context

This was the only instance of "order(id" I could find in the code base.

I haven't tested this as I don't have the development environment set up.

### DIFF
--- a/app/controllers/api/v1/notifications_controller.rb
+++ b/app/controllers/api/v1/notifications_controller.rb
@@ -12,6 +12,6 @@ class API::V1::NotificationsController < API::V1::RestfulController
   end
 
   def accessible_records
-    current_user.notifications.includes(:actor, :event).order(id: :desc)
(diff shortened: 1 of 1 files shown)


robguthrie:
Thanks for the PR!

Seems like an opportunity to discuss code.

My first thought about this is "Wow, cool!!"

My second and third throughts are:
- What's the cost of this? is there an index for notifications.created_at, is it written in DESC mode, what's the right kind of index for this?
- Is this going to help in practice, because notifications are fairly short lived in usefulness, there are A LOT of them. Could we just rewrite the notification ids according to the created_at date via a migration, and save ourselves carrying and updating an index?

asdofindia:
Yeah, anyhow with ID sequences corrected now, once 30 new notifications come in it'll be sorted on its own.

PR #8590, opened 2022-05-20
Ability to switch to previously uploaded avatar

asdofindia:
Presently, if a user switches to gravatar after uploading their image, they will not be able to switch back from the front-end.

Screenshot:
image

There are a couple of problems with this PR:

1. When a user chooses something else, say gravatar. For a brief second the "Restore previous upload" shows up. This is a bit ugly.
2. I'm not sure if I'm doing translations correctly. Should I add the key to all the translation files?
3.  What happens if the user doesn't already have an uploaded image?

The problem with 3 is that if the user is having "initials" or "gravatar", there is no way for the frontend to know if an upload already exists. Maybe once an API is built for that we can use that to check if this option should be made visible or not.

I'm not sure if the vue is good either.

Feel free to suggest modifications or alternate approaches.

### DIFF
--- a/app/controllers/api/v1/profile_controller.rb
+++ b/app/controllers/api/v1/profile_controller.rb
@@ -62,7 +62,7 @@ class API::V1::ProfileController < API::V1::RestfulController
   end
 
   def avatar_uploaded
-    render json: {avatar_uploaded: current_user.uploaded_avatar_url_if_any}
+    render json: {avatar_uploaded: current_user.uploaded_avatar_url}
   end
 
   def destroy
--- a/app/controllers/api/v1/profile_controller.rb
+++ b/app/controllers/api/v1/profile_controller.rb
@@ -61,6 +61,10 @@ class API::V1::ProfileController < API::V1::RestfulController
     respond_with_resource
   end
 
+  def avatar_uploaded
+    render json: {avatar_uploaded: current_user.uploaded_avatar_url}
+  end
+
   def destroy
     service.deactivate(user: current_user)
     respond_with_resource
--- a/app/controllers/api/v1/profile_controller.rb
+++ b/app/controllers/api/v1/profile_controller.rb
@@ -62,7 +62,7 @@ class API::V1::ProfileController < API::V1::RestfulController
   end
 
   def avatar_uploaded
-    render json: {avatar_uploaded: current_user.uploaded_avatar_url}
+    render json: {avatar_uploaded: current_user.uploaded_avatar_url_if_any}
   end
 
   def destroy
--- a/app/models/concerns/has_avatar.rb
+++ b/app/models/concerns/has_avatar.rb
@@ -54,6 +54,12 @@ module HasAvatar
     )
   end
 
+  def uploaded_avatar_url_if_any(size = 512)
+    uploaded_avatar_url(size)
+  rescue
+    nil
+  end
+
   def has_gravatar?(options = {})
     return false if Rails.env.test?
     hash = Digest::MD5.hexdigest(email.to_s.downcase)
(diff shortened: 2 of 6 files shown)


robguthrie:
I think you're going to need to query the server and ask "do I have an uploaded avatar file associated to me" and the response will tell you if you should show the "restore previous upload" button. It will be very quick.

Here's an idea, if that's helpful:

change_picture_form.vue
```
Records.remote.fetch('/profile/avatar_uploaded'). then (res) => @show_restore_previous_upload = res.avatar_uploaded
```

profile_controller.rb
```
def avatar_uploaded
  render json: {avatar_uploaded: current_user.uploaded_avatar.attached?}
end
```

asdofindia:
Thanks for the code snippet! I've modified a few more files now. See how it comes up:

For user without an upload:
image

For user with an upload:
image

I'm not really happy with what I did in has_avatar.rb. For the users who do not have an existing upload, it just throws this error:

```

NoMethodError in API::V1::ProfileController#avatar_uploaded
undefined method `blob' for nil:NilClass

```ruby
  def uploaded_avatar_url(size = 512)
    Rails.application.routes.url_helpers.rails_representation_path(
        uploaded_avatar.representation(resize_to_limit: [size,size], saver: {quality: 80, strip: true}),
        only_path: true
    )
```

robguthrie:
Nice progress. Let me know when you'd like me to review it :)

asdofindia:
Thanks @robguthrie, I just wrapped that error in another method. It might still be ugly but I'm not able to refactor further because I don't fully understand what the avatar_url method is doing. So, this PR is ready for review.

asdofindia:
@robguthrie - this is ready for review. (Sorry for being confusing in my previous comment)

robguthrie:
No problem. Sorry for the delay! I apprecaite the PR, but I just had a few
weeks break. I'll review this tomorrow. Thanks for reminding me.

robguthrie:
Hi @asdofindia - I've just had a look now. This is a neat PR. I like it.

I've made the following commit, but I don't actually know how to push it into this branch:
https://github.com/loomio/loomio/commit/848e69a373688596dd3cb141c86e18ef833cc9ae

Just a couple of tidy-ups, what do you think?

asdofindia:
I pulled that commit into this branch.

robguthrie:
Thank you. I hope the changes make sense. I probably should have just commented the changes so that you understand them, I'm really out of practice with reviewing code!

The main change was avoiding the rescue call, which was catching all errors, and we want to avoid that incase the error that was raised was not what we were expecting. Plus ActiveStorage has a nice method to check for existence of an attachments, so we'll use that.

You did a great job, especially since so much of Loomio is custom libraries or rather rare languages.

Thank you again.

asdofindia:
Thank you for the detailed response.

> Thank you. I hope the changes make sense. I probably should have just commented the changes so that you understand them, I'm really out of practice with reviewing code!

Yeah, that's one workflow where we can discuss code directly. But it takes longer to do. So maintainer adding commits directly is just as fine in most situations.

> 
> The main change was avoiding the rescue call, which was catching all errors, and we want to avoid that incase the error that was raised was not what we were expecting. Plus ActiveStorage has a nice method to check for existence of an attachments, so we'll use that.

So, now, these lines in [45-47](https://github.com/loomio/loomio/blob/715beeb0178bbae7a347d7e7a95f0ee6e73c5b47/app/models/concerns/has_avatar.rb#L45-L47)
```
  rescue ActiveStorage::UnrepresentableError
    update_columns(avatar_kind: :initials)
    nil
```
when do they get triggered ever?

I was not sure about the behaviour of this rescue call and that's why I decided to not mess with the function uploaded_avatar_url which it depends on, but rather create a new function unused by anyone else.

In other words, have we now changed the behaviour of avatar_url function?

> 
> You did a great job, especially since so much of Loomio is custom libraries or rather rare languages.
> 

Thanks! I had once looked at coffeescript long time ago for literate programming. The build tooling around javascript is so fragile that these days I prefer to remove as many layers of complexity as possible. Happy to be working on something useful in ruby on rails too.

PR #8754, opened 2022-07-26
Update

Justman10000:


Justman10000:
@robguthrie

Justman10000:
Saw that Transifex being used... Moved translations there!

Justman10000:
See #8757

PR #8757, opened 2022-07-27
Update

Justman10000:


Justman10000:
Would like to become reviewer or even better coordinator

Justman10000:
Would like to become reviewer or even better manager

robguthrie:
Hi @Justman10000, thanks for your help. Happy to make you a manager, what is your transifex username?

robguthrie:
However, I'm not sure that these modifications to the en locale file are improvements. Can you tell me any more about why you made these changes?

Justman10000:
> Hi @Justman10000, thanks for your help. Happy to make you a manager, what is your transifex username?

I would like to be a manager for German... On Transifex my name is the same as here on GitHub! I looked at the role hierarchy... If it weren't too much to ask, I would even like to be a project maintainer... Have worked on a few projects, of which I even manage 2! Can therefore bring references and trust

robguthrie:
I really appreciate the help!

I've made you a language maintainer, and happy to add you as maintainer to other languages if you wish.

config/locales/client.en.yml
```diff
@@ -380,17 +379,15 @@ en:
...
       label: Notifications
       loud: Subscribed
       normal: Normal
       quiet: Ignore
       question: When do you want to be notified?
-      loud_explain: whenever there is activity
-      normal_explain: when someone asks for your attention
-      quiet_explain: no notifications
-
+      loud_explain: Whenever there is activity
```
robguthrie, line 388: This is composed into a larger string, so that's why it begins with lower case. I would not do that now, but this was done a long time ago.

config/locales/client.en.yml
```diff
@@ -505,7 +502,7 @@ en:
     recent_threads: 'Recent threads'
     invite_only_threads: 'Invite-only threads'
     direct_threads: 'Direct threads'
-    unread_threads: 'Unread threads (%{count})'
+    unread_threads: 'Unread threads: (%{count})'
```
robguthrie, line 505: This just does not look good in the sidebar, I've tried it before :)

config/locales/client.en.yml
```diff
@@ -600,19 +597,19 @@ en:
     cancel_invitation: Cancel invitation
 
   notification_models:
-    discussion: 'thread: '
-    comment: 'comment in '
-    poll: 'poll: '
-    outcome: 'outcome: '
-    stance: 'vote: '
+    discussion: 'Thread: '
```
robguthrie, line 600: These are composed into a larger string, again, I would not do it this way if I did it now. but this is old stuff. This change will make it look quite odd.

config/locales/client.en.yml
```diff
@@ -600,19 +597,19 @@ en:
...
+    stance: 'Vote: '
 
   notifications:
     header: "Notifications"
     no_notifications: "You don't have any notifications"
 
     with_title:
       comment_replied_to: "%{actor} replied to you in: %{title}"
-      discussion_announced: "%{actor} invited you to a thread: %{title}"
+      discussion_announced: "%{actor} invited you to the thread: %{title}"
```
robguthrie, line 612: Sorry, I think 'a' is better here. Making a change would mean all the translations of this become invalidated.

Justman10000: It's clear, but if you go after that, you shouldn't update anything in the source file! What if there really are typos? you add strings? It will show you when translations are out of date if they have been updated! And the is better precisely because you have the %{title} tag! From this I conclude that there is a single notification for each thread! Since the tag %{title} exists, it doesn't mean just any thread... As an overview:


%{actor} invited you a the thread

%{actor} invited you to the thread: %{title}

Justman10000:
> I really appreciate the help!
> 
> I've made you a language maintainer, and happy to add you as maintainer to other languages if you wish.

Well, was hoping for Project Maintainer 😅

config/locales/client.en.yml
robguthrie: If there are typos, I'm happy to fix them.
Yea, I guess 'the' is a little better. Thanks.

robguthrie:
Is there a reason why you need Maintainer? I'm just being cautious because I'm unfamiliar with what those permissions involve, I can always increase it if a reason presents itself.

Justman10000:
> Is there a reason why you need Maintainer? I'm just being cautious because I'm unfamiliar with what those permissions involve, I can always increase it if a reason presents itself.

I have a lot of experience, can also enclose many references... I do need it if something should happen, like when nobody does a good job and I can give reviewers...

robguthrie:
I don't understand what this means: "like when nobody does a good job and I can give reviewers".

I'll give you more permission if a situation arises that requires it.

Justman10000:
> I don't understand what this means: "like when nobody does a good job and I can give reviewers".
> 
> I'll give you more permission if a situation arises that requires it.

No, I need the rights now so that I can do it if it happens! If we are waiting for such a situation to arise and you can give me the rights, you can do it yourself... And the above was just _an_ example...

Justman10000:
@robguthrie All fixed

Justman10000:
@robguthrie

All fixed

Justman10000:
@robguthrie

And?

robguthrie:
You have not tested these changes, for instance Node 14 -> 16 does not work. Ruby latest will not work.

Justman10000:
> Ruby latest will not work.

Where do you want to get that from?

Justman10000:
Moved to new PR (#8818) due to outdated code!

PR #9007, opened 2022-10-02
A working Procfile.dev for booting up both backend and frontend server

chetanmittaldev:
This pull request contains:-

1. A Procfile.dev file which starts a Rails server on port 3000 (api target at ```$PORT=3000``` is hardcoded in $application_path/vue/vue_config.js line no: 46); and a Vue server on port 8000 (by default ```npm run serve``` starts on ```$PORT=8080``` which might conflict with Apache server if it is installed)
2. A TODO.md which lists tasks (Todos | Doing | Done) in plain Markdown


chetanmittaldev:
This pull requests resolves:-

foreman Procfile doesn't boot the frontend server #9034

PR #9083, opened 2022-10-21
Add vendor folder to gitignore

anarute:
When running the application locally the vendor folder keeps appearing as unstaged changes, so better to ignore it to avoid pushing by mistake and also make easier to track the unstaged changes.


robguthrie:
thank you

PR #9400, opened 2023-01-24
Add deps and config for using GCS as storage

### LINKED ISSUE #9388: Allow using GCS on active_storage
I'd like to use GCS instead of s3 or disk to store the blobs from loomio.

After looking into the code it looks like this already works, but the `google` service should be added to the `config/storage.yaml` and the dependency for GCS is not installed by default in the `loomio/loomio` docker image.

I can create a PR with those changes if they are ok with you.

Thanks for creating loomio!

ludat:
Fixes #9388


robguthrie:
Thanks! Just checking: is it possible to put credentials into the ENV, rather than a json file?

ludat:
Kind of yes.

The credentials file is what you get from gcs, That's kind of the standard for the credentials. ActiveStorage allows to break down that file into its components.

So for example the file may look like this (taken from the official activestorage guide):

```yaml
google:
  service: GCS
  credentials:
    type: "service_account"
    project_id: ""
    private_key_id: <%= Rails.application.credentials.dig(:gcs, :private_key_id) %>
    private_key: <%= Rails.application.credentials.dig(:gcs, :private_key).dump %>
    client_email: ""
    client_id: ""
    auth_uri: "https://accounts.google.com/o/oauth2/auth"
    token_uri: "https://accounts.google.com/o/oauth2/token"
    auth_provider_x509_cert_url: "https://www.googleapis.com/oauth2/v1/certs"
    client_x509_cert_url: ""
  project: ""
  bucket: ""
```

but it's quite a bit lengthier so if you make a mistake the library will probably do something really weird and if google decides to change that format it will be really annoying to support

All that said if you are convinced I can make the change to use env vars.

robguthrie:
Ah! Yes. I actually use a  google cloud credentials ENV already. 

To enable google translation in Loomio, you we define a TRANSLATE_CREDENTIALS env, which is the string of the JSON.

My preference is that you can always configure all the settings for Loomio with ENV's, this PR, if unchanged, would introduce the first config that needs a file, and there isn't an existing way to add files in the loomio-deploy system.

Do you think we can assume that the translate credentials would be the same as the GCS credentials? Seems likely to me.

Actually, reviewing the translation config stuff: https://cloud.google.com/ruby/docs/reference/google-cloud-translate/latest/AUTHENTICATION

It looks like the user can set GOOGLE_CLOUD_CREDENTIALS and use either the file path or the contents of the json.

Can we do this rather than setting yml files?

robguthrie:
Hi @ludat any thoughts from my last comment? does it make sense?

ludat:
Yep and it turns out it's right! I just tried it and it works like a charm. I'll update the PR.

(Sorry for the delay I was on vacation and testing this is not as easy as I expected)

robguthrie:
Thanks @ludat !

PR #9627, opened 2023-03-12
Fixed forever loading in the Invite-only threads page

ludat:
I just deployed a staging version of loomio for our org and I noticed that the invite-only threads page is slightly broken:

If the backend responds with an empty list the page just stays there loading forever.

Please try if this PR actually works I don't know much vue, and I couldn't check if this actually works.


robguthrie:
Thanks for giving it a go, but it was not a correct fix. I've fixed it with this commit: https://github.com/loomio/loomio/commit/63c7e5ec0158c2882089ea31830359be6e17ccf6

PR #9863, opened 2023-08-24
Copy template title to poll title as well

anarute:
When using a template, the expected behavior is that it will prefill
the template values to the poll so it saves time when creating
similar proposals. Having the value only in the placeholder is not
as useful as copying the value to the title since the user would
need to manually type the whole title every time they use the template.

This is a regression from the previous behavior, so this patch
makes it behave as before.

### DIFF
--- a/vue/src/shared/models/poll_template_model.coffee
+++ b/vue/src/shared/models/poll_template_model.coffee
@@ -63,7 +63,7 @@ export default class PollTemplateModel extends BaseModel
         poll['titlePlaceholder'] = @[attr]
       else
         poll[attr] = @[attr]
-      
+
 
     poll.pollTemplateId = @id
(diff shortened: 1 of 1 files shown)


robguthrie:
Hi Ana,

In the thread-templates branch, I've added support for both behaviours -
with two title fields. One prefills the title (your preference), the other
is an example title to place in the background/placeholder.

The placeholder value is certainly useful for how most Loomio customers use
the app, but I wanted to support how you told us you use the app too.
They're really different values for different purposes, so I added both.

So I won't merge this patch, I've got a solution that works for everyone
coming, but thank you anyway.

anarute:
Ok, thank you for the feedback! We had to do this because it's a blocker for us to use the new template as is, when the new solution is released I'll remove this patch from our instance then.

PR #9870, opened 2023-08-30
Wrap txt and remove text-truncate from poll option

anarute:
The layout completely breaks for polls that have options with long text.

Without the patch, the vote reason overflows and can't be read entirely.

Screenshot from 2023-08-30 11-55-24

With the patch the text wraps correctly:

Screenshot from 2023-08-30 12-16-59

This patch also displays the full option before the user can vote, so the user can read the whole text:

Screenshot from 2023-08-30 12-16-47

Without the patch, the option that overflows will be hidden and the user only see one part of the option:

Screenshot from 2023-08-30 11-56-35

### DIFF
--- a/vue/src/components/poll/common/stance_choice.vue
+++ b/vue/src/components/poll/common/stance_choice.vue
@@ -41,7 +41,7 @@ export default
 </script>
 
 <template lang="pug">
-span.poll-common-stance-choice.text-truncate(:class="'poll-common-stance-choice--' + pollType" row)
+span.poll-common-stance-choice(:class="'poll-common-stance-choice--' + pollType" row)
   v-avatar(tile :size="size" v-if='poll.config().has_option_icon')
     img(:src="'/img/' + pollOption.icon + '.svg'", :alt='optionName')
   v-chip(v-if='poll.pollOptionNameFormat == "iso8601"'
--- a/vue/src/components/poll/common/stance_choices.vue
+++ b/vue/src/components/poll/common/stance_choices.vue
@@ -28,7 +28,7 @@ export default
   span.caption(v-if='stance.castAt && stance.totalScore() == 0' v-t="'poll_common_votes_panel.none_of_the_above'" )
   template(v-else)
     template(v-if="!datesAsOptions")
-      .poll-common-stance-choice.text-truncate(
+      .poll-common-stance-choice(
         v-for="choice in choices"
         v-if="choice.score > 0 || pollType == 'score'"
         :key="choice.id"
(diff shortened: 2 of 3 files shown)


robguthrie:
Hi Anna,

I'm not sure about this.. When there are multiple long options, the votes section becomes too long.

For example if there was multiple 'very long..' options, and it was a ranked choice poll, with many votes, then every vote with a reason would repeat all the option names. A single vote could fill most of the screen, and multiple votes would be unusable.

When you're reading the vote reasons, you don't need to read the whole option name, they've already been printed in the results table. you just need to know which option it is.

Additionally, this is why we have a "name" and a "meaning" field on poll options - ideally the name is kept short, and the meaning field is where the detail is added, so that we don't have long names that break the UI.

So, what I would prefer we do for the bug which you've identified, is to truncate the text rather than wrap it, and possibly add a length limit to the poll option name, and make it obvious/easier to fill in the meaning field. 

I'm not sure why the text isn't truncating, and rather is overflowing. Are you sure that you're running the latest version? 

Here is how long options look on loomio.com:

image

anarute:
> When you're reading the vote reasons, you don't need to read the whole option name, they've already been printed in the results table. you just need to know which option it is.

Right that makes sense, so it's fine to truncate them in the answer part, but I think not when the user hasn't voted yet because they need to read the whole option.

> Additionally, this is why we have a "name" and a "meaning" field on poll options - ideally the name is kept short, and the meaning field is where the detail is added, so that we don't have long names that break the UI.

I guess this is another issue then, I cannot see the "meaning" field when creating the poll, only when editing it:

Screenshot from 2023-08-31 12-02-52

That's why we are not using it :(

> So, what I would prefer we do for the bug which you've identified, is to truncate the text rather than wrap it, and possibly add a length limit to the poll option name, and make it obvious/easier to fill in the meaning field.

Adding a limit could solve it too, but still truncating in the time to vote I don't think is a good idea because the user needs to read the whole tittle. I understand long options can make the UI not so good, but reading the whole option before voting is more important in this case, no?

> I'm not sure why the text isn't truncating, and rather is overflowing. Are you sure that you're running the latest version?

Yes, I've tested on master and on v2.19.0 - which is what we have in our instance. Mind that this happens in the Poll view, not in the thread. So for polls without threads or if you go to the poll view. I tested with the Simple Poll type.

anarute:
> Maybe I should remove the current "add option" field, and replace it with a button that opens a dialog (basically the edit option dialog)

I think that makes sense :)

robguthrie:
Ok I'll do that tomorrow, and adjust those truncate rules.. Hopefully they
won't be needed much when the new dialog is in place.

anarute:
Sounds great, thank you!

PR #9875, opened 2023-09-02
Change 'add option' ui from a name field, to a button that opens the poll_option_form

robguthrie:
This is because many people did not realise there was an option 'meaning' which handles long information much better than option name.

This way, we can reference detailed options by name, and only show detail when user is voting.

Very long names are really hard to work with throughout the UI - so we must encourage use of the meaning field

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -1764,11 +1764,12 @@ en:
     edit_option: Edit option
     option_name: Option name
     option_name_hint: A short name for this option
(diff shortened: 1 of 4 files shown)


robguthrie:
@anarute this is what we discussed. Hopefully you won't need to add long option names now :) Would love to hear how it works for you.

robguthrie:
@anarute oh, I need to copy these changes to the poll-template-form too.. However this should work for you

robguthrie:
@anarute this work is completed now.

PR #10220, opened 2024-01-11
emailing in to public only groups will create public discussion

### LINKED ISSUE #10219: Mail to group for public_only groups fails because discussion is private by default
We've been trying out the mail to group feature and it was failing to create discussion. I did some puts based debugging and found out that it fails at DiscussionService on

```
```

The only validation that seemed applicable was:

```
  def privacy_is_permitted_by_group
    if self.public? and group.private_discussions_only?
      errors.add(:private, "must be private")
    end

    if self.private? and group.public_discussions_only?
      errors.add(:private, "must be public")
    end
  end
```

In the group I was trying this for, Discussion Privacy Options was `public_only` and the discussion created from email looked like:

```
#<Discussion id: nil, group_id: 2014, author_id: nil, created_at: nil, updated_at: nil, title: "Just testing something - ignore", last_comment_at
: nil, description: "<div dir=\"ltr\"><div>Hi <br></div><div>testing emai...", items_count: 0, closed_at: nil, private: true, key: nil, iframe_src: nil, last_activity_at: nil, last_sequence_id: 0, first_sequence_id: 0, versions_count: 0, closed_polls_count: 0, pinned: false, importance: 0, seen_by_count: 0, ranges_string: nil, descr
iption_format: "html", attachments: [], info: {}, max_depth: 2, newest_first: false, discarded_at: nil, secret_token: nil, members_count: nil, anonymous_polls_count: 0, content_locale: nil, link_previews: [], pinned_at: nil, discarded_by: nil, template: false, tags: [], discussion_template_id: nil, discussion_template_key: nil, clos
er_id: nil>                                                                                          
```

with private: true

So, I tried sending an email to a group which has it set to `public_or_private` and sure it got posted as a new discussion there. And it was created as a private discussion.

---

I think what we need is for a public_only group the emails should create public discussion by default (and hence will pass validation too).

For private_only group it can remain as it is.

For public_or_private groups I suppose private by default might be okay too.

asdofindia:
Fixes #10219

### DIFF
--- a/app/services/received_email_service.rb
+++ b/app/services/received_email_service.rb
@@ -139,8 +139,10 @@ class ReceivedEmailService
 
   def self.discussion_params(email)
     params = parse_route_params(email.route_path)
+    group = Group.find_by!(handle: (params['handle'] || email.route_path))
     {
-      group_id: Group.find_by!(handle: (params['handle'] || email.route_path)).id,
+      group_id: group.id,
+      private: group.discussion_private_default,
       title: email.subject,
       body: email.full_body,
       body_format: email.body_format,


app/services/received_email_service.rb
```diff
@@ -139,8 +139,10 @@ def self.actor_from_email_and_group(email, group)
 
   def self.discussion_params(email)
     params = parse_route_params(email.route_path)
+    group = Group.find_by!(handle: (params['handle'] || email.route_path))
     {
-      group_id: Group.find_by!(handle: (params['handle'] || email.route_path)).id,
+      group_id: group.id,
+      private: !group.public_discussions_only?,
```
robguthrie, line 145: Nice! Here you can use group.discussion_private_default

asdofindia: ah.

robguthrie:
So wonderful to have a code PR. Thank you, @asdofindia.

PR #10537, opened 2024-03-06
Allow admins to manage memberships of any group

anarute:
When using the API to manage memberships is very inconvenient to have to add bots to every group and make them admin of the group in order to be able to manage members programmatically. This generates noise because for every group people will be able to see the bot users as members.

Considering that global admins are already able to manage the application from the admin interface, it would make sense that they can also have more permissions when using the API.

This patch allows users that are global admins of the application to be able to manage users of any group when using the API.

This is a proposal and I understand if you don't want to accept it, but we ended up adding it downstream to our instance so I decided to send it here in case you find it useful too :)

### DIFF
--- a/app/controllers/api/b2/memberships_controller.rb
+++ b/app/controllers/api/b2/memberships_controller.rb
@@ -43,8 +43,12 @@ class API::B2::MembershipsController < API::B2::BaseController
   end
 
   def group
-    group = current_user.adminable_groups.find_by(id: params[:group_id])
-    raise CanCan::AccessDenied unless group
+    group = Group.find(params[:group_id])
+    raise ActiveRecord::RecordNotFound, "Group not found" unless group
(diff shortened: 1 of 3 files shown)


robguthrie:
Oh interesting! Thanks for sharing this. 

It would be really good to have a is_admin and is_not_admin testcase associated with this, testing allow and deny, would you be up for that? Otherwise I can add something.

anarute:
> Oh interesting! Thanks for sharing this.
> 
> It would be really good to have a is_admin and is_not_admin testcase associated with this, testing allow and deny, would you be up for that? Otherwise I can add something.

Great, thanks for the quick review! I've added a pair of tests, please let me know if they are enough.

robguthrie:
Yeah, that looks great. Thank you!

PR #10581, opened 2024-03-26
Enable admins to revoke memberships

anarute:
Currently global admins can add new members to groups they don't belong too but they cannot remove them. When using the API, they should be able to add and revoke memberships, so this patch fixes this.

This is related to https://github.com/loomio/loomio/pull/10537


robguthrie:
So cool that you're contributing changes like this. Thank you @anarute. Happy to merge after you've considered my suggestion.

anarute:
> So cool that you're contributing changes like this. Thank you @anarute. Happy to merge after you've considered my suggestion.

My pleasure! I'm afraid I don't see any suggestion, would you mind double checking if it was sent? :pray:

app/controllers/api/b2/memberships_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
     PollService.group_members_added(group.id)
 
     removed_user_ids = []
-    if params[:remove_absent]
+    if params[:remove_absent] && params[:remove_absent].to_s == "1"
```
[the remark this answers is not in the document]
anarute: Got it! Just made the changes, thanks for the suggestion.

app/models/ability/membership.rb
```diff
@@ -21,9 +21,11 @@ def initialize(user)
     end
 
     can [:remove_admin, :revoke, :destroy], ::Membership do |membership|
-      (membership.user == user ||
-       user_is_admin_of?(membership.group_id) ||
-       (membership.inviter == user && !membership.accepted_at?))
+      user.is_admin || (
```
robguthrie, line 24: I wonder if this should also check that the user is a bot. I'll have a think.. 

The reason i say this is that, generally we don't want site admins getting different behaviors than other users, except from within the admin panel.

anarute: Right, constraining it only for bots would work for us too so if it makes more sense to you I can change in both places. Although I would say site admins can even impersonate other users so not sure if it would make a lot of difference.
Another thing to consider is that people can mark themselves as "bot" in the profile since it's been used to temporarily disable them from being added to polls too, which would give them this permission if they are admin as a side effect. Maybe we need another check to differentiate bot and who is on leave.

robguthrie: Thanks for waiting for me to think this though. You're right,iIt's pointless if you an just change the value.

PR #10934, opened 2024-07-03
[Comment] Mention @discussion or @group to represent all members of the thread or group

### LINKED ISSUE #8148: Provide a @all alias shorthand for notifying everyone about a reply
In rare occasions, it makes sense to notify everyone in a reply, and instead of listing all names, it would be good to be able to use e.f. `@all` to achieve the same.

jerinkc:
Mentioning @discussion in the comment will notify all members of the discussion. 
Mentioning @group in the comment will notify all members of the group.

Closes #8148


robguthrie:
Hi @jerinkc - it's really wonderful to see you've made a PR to start on this feature. Thank you. I've not seen a loomio feature PR in years.

With the greatest respect, I don't think it's going in the right direction currently - I can understand why, this is a super challenging feature to add, not only is the existing mentioning code complex because we support both markdown and html comment formats, it has some legacy behavior support included in it too. 

Some tips I can offer:
- have a look at `app/services/announcement_service.rb` we have the key words that reference group and discussion_group (thread) in there for when you invite people to a discussion thread.
- have a look at `app/models/concerns/has_mentions.rb` and make a new file `app/models/concerns/has_audience_mentions.rb` copying the style in has_mentions but looking for 'group' and 'discussion_group' rather than usernames.
- Then create a clone of `app/models/concerns/events/notify/mentions.rb` called `.../notify/audiences.rb` which looks for mentioned_audiences and sends `app/models/events/comment_announced.rb` (also a new file)

Does that make any sense? You're free to explore the approach however you like, but this is how I recommend you look at it

jerinkc:
Hey @robguthrie ,

Thank you for your quick review and insightful comment on the current approach. I genuinely appreciate your feedback!

I’ll take the suggested changes into account and try to implement them.

jerinkc:
Hey @robguthrie,

Thank you once again for the tips! The approach is perfectly clear to me. I want to get your opinion on a couple of additional points:

- In your previous comment, you mentioned ‘discussion_group’ instead of ‘thread.’ Should I use the term ‘discussion_group’ instead of ‘thread’ throughout the implementation?

- Regarding the current approach, I’ve used two users to represent the audience in the database. I find this helpful when listing users for mentions in the UI. If this approach is acceptable, I plan to use these users to store notifications for audiences.

robguthrie:
Hi @jerinkc 
- On the server side we tend to use the term discussion rather than thread. discussion_group or discussion would be best.
- I don't think you should continue with the approach of pushing this through a fake user account. It would be better to list "audiences" with users when mentioning, then with the separate has_audience_mentions concerns you detect the use of the discussion or group token. The has_mentions concern will just ignore this.

I suggest you ignore the frontend stuff for now. Write some tests that create a comment with a <span data-mention-id="group"> in them, then use that to build the has_audience_mention concern, which creates it's own notification type. It will share a similar pattern to has_mentions but be parallel to it, and the has_mentions probably won't need to change.

jerinkc:
@robguthrie please review whenever you can.

robguthrie:
Thanks @jerinkc - I'll try to have a look this weekend :)

db/schema.rb
```diff
@@ -545,6 +545,18 @@
     t.index ["volume"], name: "index_memberships_on_volume"
   end
 
+  create_table "motions", id: :serial, force: :cascade do |t|
```
robguthrie, line 548: can you take these (the montions and votes tables) out please? It's interesting that your db has them. They were removed years ago

config/locales/server.en.yml
```diff
@@ -322,6 +322,7 @@ en:
         You must be signed in to proceed.
         Please <a href="/dashboard" target="_blank">sign in</a> then reload this page to continue.
       recaptcha: "We weren't able to verify that you're not a robot. Please try again!"
+      forbidden_name: "Forbidden name, please choose another name"
```
robguthrie, line 325: How about "You name contains a reserved word, please remove it"

app/models/events/comment_announced.rb
```diff
@@ -0,0 +1,29 @@
+class Events::CommentAnnounced < Event
+  include Events::Notify::InApp
+  include Events::Notify::ByEmail
+
+  def self.publish!(model, actor, audiences)
+    super model,
+          user: actor,
+          custom_fields: { audiences: }
```
robguthrie, line 8: is this valid ruby? Does it mean { audiences: audiences }

app/models/events/comment_announced.rb
```diff
@@ -0,0 +1,29 @@
...
+  def email_recipients
+    notification_recipients.where(email_when_mentioned: true)
+  end
+
+  def notification_recipients
+    recipients_ids = custom_fields['audiences'].inject([]) do |ids, audience|
+                      ids |= AnnouncementService.audience_users(eventable,
+                                                                Audience.send(audience).alias,
+                                                                eventable.author,
+                                                                false,
```
robguthrie, line 22: you can remove false, false from this. They're defaults, sorry that they're not named arguments, they really should have been for readability, but seeing as they're not used, just remove false, false.

app/models/events/comment_announced.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  private
+
+  def email_recipients
+    notification_recipients.where(email_when_mentioned: true)
+  end
+
+  def notification_recipients
+    recipients_ids = custom_fields['audiences'].inject([]) do |ids, audience|
```
robguthrie, line 18: I know reduce/inject is considered really clever, but I've never seen a use of it that did not take me a very long time to understand. 

To me writing 
```
ids = []
custom_fields['audiences'].each do |audience| 
ids.concat AnnouncementService.audience_users(eventable, audience, eventable.author)
end
```
is a lot easier to read. 

Ether that or how about?

```
custom_fields['audiences'].map do |audience| 
AnnouncementService.audience_users(eventable, audience, eventable.author)
end.flatten.uniq
```

I think I would write the last version if it were me.

app/models/audience.rb
```diff
@@ -0,0 +1,27 @@
+class Audience
```
robguthrie, line 1: I wonder if this whole class just wants to disappear. 

It's a lot of code around an constant array of two strings. Can we put that somewhere else?

app/models/events/comment_announced.rb
robguthrie: dont' feel obliged to change this.. I think I might just need to get comfortable with reduce

spec/models/comment_spec.rb
```diff
@@ -82,4 +87,17 @@
       end
     end
   end
+
+  describe "#mentioned audiences" do
+    it "mentioning an audience should return corresponding audience" do
+      audience = Audience.all.sample
+      comment1 = create :comment, body: "Hey, @#{audience}"
+      comment2 = create :comment,
+                        body: "<p>Hey, <span class=\"mention\" data-mention-id=\"#{audience}\" label=\"#{audience}\">@#{audience}</span></p>",
```
robguthrie, line 96: can you take the interpolation of audience out of this, and test it explicitly in each case please? IE:
```
body: "<p>Hey, <span class=\"mention\" data-mention-id=\"group\" label=\"group\">@group</span></p>"
```

spec/models/event_spec.rb
```diff
@@ -117,6 +117,103 @@ def emails_sent
     end
   end
 
+  describe 'audience_mentioned' do
+    let(:discussion_audience){ Audience.discussion.name }
+    let(:group_audience){ Audience.group.name }
+
+    it 'notifies group members' do
+      comment = create :comment, parent: parent_comment, discussion: discussion, body: "hey @#{group_audience} good day"
```
robguthrie, line 125: also, I recommend testing this explicitly here. `hey @group good day` rather than interpolating it.

robguthrie:
Hi @jerinkc this looks like it's coming together fairly well. Amazing work. 

The biggest question I have is around supporting multiple languages. Won't we need to support this keyword in French, German and all the other languages Loomio supports?

Maybe the Audiences class will be important for this afterall?

config/locales/server.en.yml
robguthrie: Now I understand this a little better. "The name you have provided is a reserved keyword in Loomio".

jerinkc:
Hi @robguthrie,

The internationalization feature was the primary reason for having the Audience class initially. However, I inadvertently missed including it in the latest changes. Based on your comments, I’ll promptly address this oversight and incorporate the necessary updates.

Thanks again for catching that, and I’ll ensure everything is in place after a careful review.

app/models/events/comment_announced.rb
jerinkc: Yes it is a ruby 3.1 feature - https://github.com/ruby/ruby/commit/c60dbcd1c55cd77a24c41d5e1a9555622be8b2b8

robguthrie:
Yes, I realized after thinking a bit more that the "group" and "discussion" tokens won't need to change so you can just add a $t('mentioning.group') or something in the UI and it will all just work.

It's looking good. Thanks again for your work!

jerinkc:
Hey @robguthrie, 

I have added more changes along with the internationalization feature. 

Also, should I work with the UI part here or would you prefer it in another PR?

spec/models/comment_spec.rb
```diff
@@ -82,4 +87,27 @@
...
+    end
+
+    it "mentioning an audience should return corresponding audience" do
+      audience = Audience.all.sample
+      audience_translated = Audience.send(audience).translate
+
+      comment1 = create :comment, body: "Hey, @#{audience_translated}"
+      comment2 = create :comment,
+                        body: "<p>Hey, <span class=\"mention\"
+                          data-mention-id=\"#{audience_translated}\"
```
robguthrie, line 105: oh, this part should be "group" or "discussion" no matter what language. So you only need to translate the user generated content.

app/models/audience.rb
```diff
@@ -0,0 +1,41 @@
...
+    {
+      discussion: 'discussion_group'
+    }[name.to_sym] || name
+  end
+
+  def translate
+    I18n.t("mentioning.#{name}", default: name)
+  end
+
+  def self.back_translate(word)
```
robguthrie, line 28: you won't need back translate if we don't translate the audience name...

robguthrie:
You can put the UI stuff in here too, please.

Sorry there was a misunderstanding about how to handle translation. The code/server side should just be english in all cases - it will be surprisingly simple to translate just the visible text with the existing translation library in the UI only.

jerinkc:
@robguthrie 

I understand you want to apply translation only to the visible part of the UI. Which is for the UI code it should be something like, 

`
<p>Hey, 
    <span class=\"mention\"
        data-mention-id=\"discussion\"
         label=\"discussion\">
         @diskussion
    </span></p>
`
So the server will recognize when a discussion is mentioned.

If internationalization applies to the markdown format, how will it detect that the incoming message has mentioned audiences without the current configuration? I think we should retain the current translating functionality(including back_translate method) and only apply it to the markdown format. What do you think?

robguthrie:
Oh Yeah. That's right, it will be necessary for markdown formatted content. Thanks for thinking of that.

jerinkc:
@robguthrie Take a look whenever you get a chance. Thanks!

robguthrie:
Sorry I've taken so long to review this. Will run and test it tomorrow. 
Can you tell me - do you think this is finished? Do you think there is any more work to do, and have you had the opportunity to deploy it and test it with users?

jerinkc:
Hey,

It is done if it applies only to comments. I’ve tested it locally with multiple users, and everything seems to be working fine. Just a heads-up, I haven’t included translations for other languages.

Here’s a quick demo video: [Video Link](https://drive.google.com/file/d/1DxRlItSBEAa_2pczVHq9ojmPiqodLd9I/view?usp=sharing)

config/locales/client.en.yml
```diff
@@ -714,6 +718,7 @@ en:
       user_added_to_group: "%{actor} added you to the group: %{title}"
       user_mentioned: "%{actor} mentioned you in: %{title}"
       user_reminded: "%{actor} is reminding you to vote in: %{title}"
+      comment_announced: "%{actor} mentioned you in: %{title}"
```
robguthrie, line 721: I wonder if this should be more specific. What do you think about having two messages:
"%{actor} notified everyone in the group" and 
"%{actor} notified everyone in the thread"

robguthrie:
Ok, I've given this a test run. It looks really promising! Well done.

image

I notice that you need to type @discussion to mention everyone in the thread.. The Loomio term is "thread" so I think it would make sense to use that as the keyword.

Also when I check the "Notified" button I see this:
image
Which is from notifying the group.. but there are no entries for notifying the thread. I'm not sure why that is. 

I'll look more closely at it and get back to you

jerinkc:
I missed the notification part during implementation, it has been added in the latest commits. Also, the UI now uses @thread instead of @discussion.

Screenshot 2024-09-11 at 1 28 49 PM
Screenshot 2024-09-10 at 11 28 22 PM
Screenshot 2024-09-10 at 11 28 41 PM

robguthrie:
Wow. It's really coming together now. There are two last things that need some attention..

Permissions.

Frontend:
Can you ensure that @thread and @group mention suggestions only show when the user is permitted to notify?
We need to call AbilityService.canNotifyGroup and AbilityService.canAnnounceDiscussion before adding the mention options to the list and presenting them to the user.

https://github.com/loomio/loomio/blob/master/vue/src/shared/services/ability_service.js#L135C3-L135C24

Backend:

To be secure we also need to authorize this on the back end. You can see an example of this here:
 
https://github.com/loomio/loomio/blob/master/app/extras/user_inviter.rb#L72

So you can call `user.ability.authorize! :announce, discussion` and `user.ability.authorize! :announce, group` on the back end to ensure the client is allowed to notify.

Education:
It would be good to explain what these options do, as people will see them and not know.. I wonder if we can put a subtitle in the suggestions list "Notify everyone in the group" under the @group, for example.

jerinkc:
Hey, I have made the necessary changes. Also, I used `user.ability.authorize! :notify, group` instead of `user.ability.authorize! :announce, group`. Let me know if it works.

Screenshot 2024-10-13 at 12 58 25 PM
Screenshot 2024-10-13 at 12 42 20 PM

robguthrie:
Will do! Thanks!

robguthrie:
I'm sorry to do this, but I've created my own approach to group mentions, which has just been merged.

I want to acknowledge how awesome it is that you took on such a big feature, and worked really hard to adopt my feedback. This branch was instrumental in me understanding how I wanted to implement the feature.

I hope there are no hard feelings about this. I'm grateful for your contribution

Issue #11107, opened 2024-09-09
Not logged in users can still see a edit profile, notification settings and log out buttons


poVoq:
I just noticed that non-logged in visitors can still see the edit profile etc. buttons:

![image](https://github.com/user-attachments/assets/edbbffe4-f16f-4fb0-8559-42bc32369c24)

I guess that is a small oversight?

robguthrie:
This would be a great little PR for someone

naaa760:
Hello @poVoq 
I would like to work on this issue

robguthrie:
fixed in v3

PR #11377, opened 2025-02-10
hid sidebar and user dropdown entries that depend on being logged in

Hamsterarsch:
**What**
Resolves https://github.com/loomio/loomio/issues/11107 - I hope this is as intended, the desired state wasn't really described in the issue.

I had to insert multiple conditionals in order to preserve the order of the menu items. (I did not remove darkmode from the the sidemenu for guests because it looks kind of buggy otherwise).

Guest menu view:
image

### DIFF
--- a/vue/src/components/common/sidebar.vue
+++ b/vue/src/components/common/sidebar.vue
@@ -195,24 +195,25 @@ v-navigation-drawer.sidenav-left.lmo-no-print(app v-model="open")
         v-list-item-subtitle(v-t="'profile_page.set_your_profile_picture'")
   v-divider
 
-  v-list-item.sidebar__list-item-button--recent(dense to="/dashboard")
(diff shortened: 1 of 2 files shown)


robguthrie:
I'm not sure I want to merge this right now. Mainly because I want to prevent merge conflicts with my vue3 branch, and this does not really add a lot of value right now. 

It's worth thinking about what the logged out sidebar should be, but I think I want to get my vue3 branch merged then fix the problem

Hamsterarsch:
That totally makes sense! Looking forward to saying goodbye to vue2 already. Maybe it would be a good idea to flag the originating issue as blocked or smth.

Hamsterarsch:
@robguthrie if merging is the issue, should I rewrite this (and future PRs) to branch of the vuetify3 branch then?

PR #11383, opened 2025-02-11
Env toggle for thread from mail feature

### LINKED ISSUE #10112: Option to disable incoming email support?
I noticed that in the latest version it now very prominently displays a "group-name@null" at the top of the group landing page.

I assume this is meant to show the mail-in address?

I don't have this configured at all and would rather like to disable this feature. Previously it was sufficiently hidden that most people would have probably never noticed the lack of support for it.

Thanks for considering this!

Hamsterarsch:
Resolves #10112 

Adds `FEATURES_DISABLE_THREAD_FROM_MAIL` env which is used to hide/show the group email action. GUI part is tested, but I could not test the changes to `received_email_service.rb` because I don't have mail set up in my development environment. Would appreciate if someone else could try if threads are still created when mail is incoming or not when the flag is disabled or enabled (I don't think me setting up local mail is worth this).
Alternatively, discarding `received_email_service.rb`  changes might also be an option?

### DIFF
--- a/app/extras/app_config.rb
+++ b/app/extras/app_config.rb
@@ -102,6 +102,7 @@ class AppConfig
       public_groups:              !ENV['FEATURES_DISABLE_PUBLIC_GROUPS'],
       help_link:                  !ENV['FEATURES_DISABLE_HELP_LINK'],
       example_content:            !ENV['FEATURES_DISABLE_EXAMPLE_CONTENT'],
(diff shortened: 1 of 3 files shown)


robguthrie:
Thanks for this. Nice a simple. I've run the received_emails_controller specs myself before merging. I am not sure why Github does not run these tests for me.

received_emails_controller is well tested and this does not break it, and the code is easy to read, so I'm happy to merge

PR #11394, opened 2025-02-16
Mobile friendly group navigation

Hamsterarsch:
V-tabs are not properly styled on small screens, resulting in overflowing navigation tabs that are hard to traverse (going right by clicking each tab one by one works, but especially for new users, the tabs to the right might as well not exist).

image

From what I read during trying to fix this, this might already be fixed in vue3, so feel free to reject this PR (just couldn't help myself ✨). Basically, this replaces v-tabs with v-bottom-navigation, making use of icons to save enough space to prevent overflow on average devices. 

## Look on mobile:

image

 ## Look on desktop:
Although tabs look fine on wider formats and changing layouts for smaller screens would be an option, I didn't want put more to work into this at this stage. (Doesn't look too bad either imo).

image

### DIFF
--- a/vue/src/components/group/page.vue
+++ b/vue/src/components/group/page.vue
@@ -50,11 +50,11 @@ export default
       if (this.$route.query.subgroups) { query = '?subgroups='+this.$route.query.subgroups; }
 
       return [
-        {id: 0, name: 'threads',   route: this.urlFor(this.group, null)+query},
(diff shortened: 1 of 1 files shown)


robguthrie:
The vue3 branch removes the subgroups tab on this page, moving all group stuff to the sidebar. So the problem is less. I think the tab scrolling is fixed in that too.

I don't think I can move from words to icons only. If someone does not already know what the icon means, then it does not do the job of explaining the tab.

Nice experiment though :)

PR #11400, opened 2025-02-17
Fix reaction display

anarute:
If a thread is closed the user shouldn't be able to send reactions but they should be able to see the existing ones.

image


robguthrie:
Nice one, thank you!

PR #11402, opened 2025-02-18
@group mentions 

robguthrie:
We already had a PR for this - I'm sorry I left that unmerged for so long. After finally making space to review it  I did not feel confident I could maintain it. So I've attempted to implement this feature myself. 

Here's the beginning of a detailed description about this.

Done:
- Supports mentioning a group from anywhere (thread context, comment, poll etc)
- In most cases, there will be no loading wait after hitting @
- Users mention the group by it's handle (eg: @enspiral)  rather than @group, which I think reads better, and means we can support mentioning subgroups - also does not need translations/back translations
- it does not suggest a group when you can't notify it
- it does not mention groups you don't belong to
- it does not double notify when combinations of subscribed, notified, mentioned or group mentioned are used

TODO:

- update "x users will be notified" notifications count to include mentioned users
- add notifications count to comment form?
- email footer:
Reply to this email directly, __view it on Loomio__, or __unsubscribe__
You are receiving this because @enspiral was mentioned.


Hamsterarsch:
Amazing :tada:

Issue #11378, opened 2025-02-10
Task Overview Improvements


Hamsterarsch:
### Feature scope 
In the task overview under /tasks
1. Split tasks into assigned to user and created by user
2. Add filter for tasks (open/closed)
3. Allow to 'archive' tasks in the overview (banishing them to a separate section)
4. Allow restore of 'archived' tasks 

## Background
Hi! I happen to use loomio for research groups now and then. These usually involve a lot of tasks - and it's so amazing loomio supports basic task management - but my task overview under /tasks has been turned into an unmanageable mess by now. There also seems to be little hope on /tasks ever being clean and tidy again because their number is growing and the only way to fight seems to be to delete the oginating thread. 

Because of this, **it's hard to keep track of what still has to be done, which tasks are assigned to me, and which tasks I simply happened to write when doing meeting protocol on behalf of others.** So basically, everything a task overview is for? 
I think the changes I listed at the top should amend all of this, or at least greatly improve the current situation. 

Feel free to give more ideas or confirm/reject my perceptions. If none of the other issues I have an eye on get activity, I'm planning to start a PR for this soon.

robguthrie:
Hi, thanks for writing this up. I agree with you, this is a good thing to do. Happy to support you with it.

robguthrie:
If you would like to have a zoom call to discuss how you might approach this, please email me and I'll send some details along.

Hamsterarsch:
Sure! I'm still getting familiar with everything though (including ruby and vue 😅). I'll play around a bit more and get back to you next week?

PR #11413, opened 2025-02-25
Task view improvements 

Hamsterarsch:
Contributes to #11378

**Sneak peak**
 
image

**Change Overview** 
- changes task view in the side bar
- adds three tabs that can be used to filter displayed tasks (closed,open,hidden)
- adds the ability to hide tasks (moves them to hide tab)
- adds a button to hide/reveal all currently displayed tasks in bulk (intended for users with many present tasks)
- removes grouping of tasks by source discussion (link to discussion is now available using the arrow button instead of using discussion title). I did not feel like the discussion title added much to the recognizability of a task entry and keeping them would have made frontend code much more complex.
- changes a bunch of locales (only en so far) to match changed functionality
- Backend: Adds a new table tasks_users_extension that represents a many many relationship. It stores per user and per task data such as the now required 'hidden' value for a task. (The existing tasks_users relation was unsuitable for this type of storage because it's entries represent what users have been mentioned by a task. Adding more data to it would have required a more extensive change of the task parsing code). Although right now, the tasks_users_extension table only holds the 'hidden' column, I saw old comments in code about a potential mute/silence feature for individual tasks - this would be the place to add such future persistent state.
- Pre-existing _ruby_ tests are passing (see below)
- layout shrinks properly on mobile (at least on my machine)

**Problem Areas**
1. The JS end to end test for the new task page is incomplete. I couldn't figure out which test suite is used by only going off of the existing tests so I couldn't really work on it. Also, I wasn't sure if they are used at all because they are so few of them.
2. I am unsure about the convention around dependent: destroy; I added dependent destroy between task and the new table tasks_users_extension but then I saw that there is no such dependency between task and user. I was wondering if this is intentional. If it isn't, I would add the dependency between both user/task and user/tasks_users_extension. **Therefore,** the last test in task_service_spec which tests user deletion = extension deletion currently fails.
3. The frontend code to hide/reveal all is not using a dedicated update all function on the backend. I don't think this functionality will be used often or with so many tasks that it becomes a performance issue so it didn't seem valuable enough to justify adding more code. 

I'm sure there will be more problems - looking forward to them!


robguthrie:
Wow. This is incredble. Love how you've understood the Loomio architecture and are making great progress. Also really like the tab UI.

Hamsterarsch:
(I just noticed the changes to sidebar/userdropdown are still in here - just ignore those.)

Hamsterarsch:
- [ ] todo: Add filter or view for assigned/created by you

Hamsterarsch:
**Updates**
- Added a dropdown to filter by assigned to user, created by user, all (assigned to user is default)
- Changed layout to vcard to support expiry counter (things didn't fit in a single line anymore). A nice side effect is that the tasks usually display in full now, although 1/3rd of the viewport is the limit
- Removed e2e tests because (1) don't know how to run them (2) not sure how to select elements in the DOM without copying overly verbose selectors from debugger (3) it's not that critical if the task view breaks

**Remarks**
- When filtering tasks for assigned/authored I noticed that the doer_id column in Tasks does not seem to actually represent who is supposed to do a task? I had to remove it from the query in the task controller to get the required result.

Present ruby tests are passing

Screens




@robguthrie if this looks good to you, this should be in a mergeable state now as far as I am concerned

robguthrie:
thanks, very exciting. 

I've just tested positive for covid, so I'll need a few days before my brain is back online. Sorry for the delay.

Hamsterarsch:
> thanks, very exciting.
> 
> I've just tested positive for covid, so I'll need a few days before my brain is back online. Sorry for the delay.

Get well soon; it's not urgent after all (and I suppose the next release is still a while off anyways)

robguthrie:
Just had a chance to review this. Looks good but I think the following changes will improve it. Lemme know what you think.

Thanks gain for this work. It's looking good :)

Title of page should be "Tasks"
"Expires in 8 Days" should say "Due in 8 days"
Heading should not be the dropdown. I think it should be separate.
Default to "all tasks" view
Show author name on task
The icons are not super obvious, there is room for a word: Hide, Unhide
"Toggle Hide All" should be replaced by "Hide all" and "Unhide all"
Open should be "To do"
Closed should be "Done"
Please put 'title' elements on icon-only buttons
'Tasks to do' should be 'Assigned to you'
'Tasks you Made" should be "Made by you"
"All your tasks" should be  "All yourtasks"
-> icon could say "View thread"

Hamsterarsch:
I generally agree (I didn't consider adding text to the icons when switching from the previous, tight, one-row layout).

Could you explain why you think that displaying all tasks by default is better? From a UX perspective, it seems more likely to me that a user wants to quickly check their own todo/assigned tasks as opposed to those they assigned to others (because checking if a tasks has been done by others seems more rare to me, e.g. right when a known or unknown deadline is approaching).

also, is this a typo? `"All your tasks" should be "All yourtasks"` ?

robguthrie:
Yeah, it's a typo :)

For the default filter, I expect that you see "all" by default, it's easy to scan down and find the things you're interested in, but if you don't see something you're expecting it's confusing unless you know a filter is applied.

I wouldn't expect done task in the default view, but it's likely that I'm thinking of tasks I created or tasks I'm assigned when I open the page.

Filtering is a little more advanced. at first, a user does not know about the filter - but you think of it when you have too many in your list  and need to remove some because of overload.

robguthrie:
When you look at the page without interacting, you have no idea what options the dropdown has. The Done tab is obvious, so looking at the page I can tell that done tasks are not onscreen - but unless I decide to click the dropdown I don't know what else is available.

Hamsterarsch:
I agree that filter options should be more obviously displayed; I stuffed it into a dropdown for better mobile layout but I suppose I will have to come up with a different GUI.

Todo:
- [ ] Title of page should be "Tasks"
- [ ] "Expires in 8 Days" should say "Due in 8 days"
- [ ] Heading should not be the dropdown. I think it should be separate.
- [ ] Default to "all tasks" view
- [ ] Show author name on task
- [ ] The icons are not super obvious, there is room for a word: Hide, Unhide
- [ ] "Toggle Hide All" should be replaced by "Hide all" and "Unhide all"
- [ ] Open should be "To do"
- [ ] Closed should be "Done"
- [ ] Please put 'title' elements on icon-only buttons
- [ ] 'Tasks to do' should be 'Assigned to you'
- [ ] 'Tasks you Made" should be "Made by you"
- [ ] -> icon could say "View thread"

Hamsterarsch:
Task Card Progress; I'd appreciate opinion on the layout @robguthrie 

image

robguthrie:
- How does this work in a mobile phone width page?
- I suggest that the filter look like the group page thread filter. A dropdown/select.
- I suggest that the CAPS headings are unnecessary, and layout could imply what they mean. ("Task" content can just appear in a list of tasks and it will make sense)
- The "written in" should be a grouping label sitting above a group of tasks from the same place. A thread is likely to have 2 or 3 tasks that are assigned to me.
- Mark as done is the primary action and I would give it a label
- I would put all the actions for a task under the task content, so that the the task acts as a title for the task listitem

By the way, I wonder if we should arrange a call to meet? I feel unsure about how to deliver feedback and what you're looking for here. I invite you to email me if you would like to arrange a call. rob at loomio dot org

Hamsterarsch:
@robguthrie Got around to updating to vue3 and working on this again. Applied your input except for the "source discussion" display since I wanted to keep the task sorting simple. Right now, tasks are ordered top to bottom by expired > expiring soon > modification date new > modification date old. I chose this ordering because it puts the likely most relevant tasks at the top (where relevant means at due date or with recent activity). Adding grouping by discussion seems quite incompatible with that idea of ordering by relevancy and would introduce a whole new level of complexity. Therefore, I don't really want to implement it (also because I personally don't have a use case for a task view by discussion).
If a view of tasks by discussion is necessary, perhaps it could be introduced as a second page in the task menu in the future, or as an overview in the discussion pages?
Alternatively, I could also remove the source discussion display from the task cards - I only added it because I thought it might give additional context in cases where tasks between discussions are too similar to be distinguished by their content alone.

Hope we can nail down the layout soon so this still has a chance to be merged.

image

Issue #9231, opened 2022-11-29
Support web push notifications


airblag:
It would be really nice, if the notifications on the bell could be pushed to the browser using web push notifications.

I think this would help also people who are used to using apps and added the loomio dashboard to the homescreen, having the possibility to get notifications directly on the phone when something happens, and get a red dot on the homescreen shortcut.

I'm not into web development, and I don't know much about how this feature works, and if it creates dependencies to some kind of push provider, or if it's possible to make it work without them. 

But I think it would help my community of looking more into our loomio instance.

robguthrie:
Thanks. Yes we would like to add this, as it's critical to developing a PWA style mobile app.

In the meantime, if you use Slack or other chat, it's a good idea to use setup a Chatbot integration.

Hamsterarsch:
hey @robguthrie, I think I read on the 'old' loomio instance that someone wanted to work on this but has there been any progress so far?
Otherwise, I'd say this is my highest priority feature and I'd like to work on it.

Hamsterarsch:
## PWA 1: Push Support

**Goal**
Mirror existing email notifications to PWA push notifications (Email subject = push content)

**Research**
- [Related Community Thread](https://www.loomio.com/d/5WIpOUeu/push-notifications)
- [Community Thread 2](https://www.loomio.com/d/8zItQmau/redirecting-loomio-smtp-notifications-to-telegram) (same user as first thread)
- Currently preferred, [maintained Ruby Gem](https://github.com/pushpad/web-push) for web-push support
- [External issue](https://github.com/scoutplan/scoutplan/issues/1939) with useful links around PWA setup

**Preliminary Implementation Approach**
- [ ] Add webpush based notify mixin to `app/models/concerns/events/notify/`
- [ ] Add new notify mixin to all classes in `app/models/events/` that `include Events::Notify::ByEmail`
- [ ] Add new service called by mixin to deliver notifications (see `app/mailers/event_mailer.rb` for how to do that). The service might use the ruby gem linked above (obviously, init for that - e.g. VAPID certificate - has to be done somewhere as well)
- [ ] Add client side service worker (see [here](https://alicia-paz.medium.com/make-your-rails-app-work-offline-part-1-pwa-setup-3abff8666194)) and request push permissions from client hardware
- [ ] Check current manifest and backend service worker controller (or add if missing)

Hamsterarsch:
[...]

PR #11428, opened 2025-03-09
PWA Push Notifications

Hamsterarsch:
Relates to #9231 

Image

## Update
First PWA session; Very rough but the basic idea is standing (using desktop/chrome, notifications are delivered). Manifest creation is done using vite pwa plugin.
Push notifies share so many requirements with `event_mailer.rb` but are also different (especially the content) so I'm not quite sure how/if to reduce the redundancy with `webpush_service.rb` yet.
Also not sure if putting the service init into the initializer folder was the right call?


Hamsterarsch:
Next steps

- [ ] figure out how to supply icons to vite during manifest
- [ ] figure out how to supply icons to client side push display
- [ ] figure out if displaying the same title/subject in push that is used by email is realistically going to read nicely. Alternatively, a nearly complete fork of the current email localization tokens would be necessary.
- [ ] figure out how to do push content localization (clientside, serverside)
- [ ] fix/test onclick > openWindow (simply putting in the url was blackscreen in chrome, might have to be routed through vue?)
- [ ] refactor mailer/pusher redundancy
- [ ] test manually with firefox
- [ ] test inside pwa (desktop), especially caching - how little client side awareness about offline mode will work?
- [ ] install popup / install buttons in settings
- [ ] what about push notifications sent during pwa offline?

Hamsterarsch:
Things to know: Without icons supplied in the manifest, the pwa won't be installable

Hamsterarsch:
## Current Implementation
### General
- manifest for pwa and website was updated to comply with requirements for native-feel pwa installation  (e.g. scope is now correctly set to loomio root instead of dashboard, added icons, 'screenshots' (just the default loomio banner) to get advanced install popup on chrome)
- users are now requested to enable notifications on login by their browser
- user are now able to visit loomio sites and install them as a pwa
- users will get a push notification (from the pwa or when they agreed to notifications) to all devices setup that way

### The Notification
image
- notification title is email title without [loomio] prefix
- notification content is now based on email body content. That means notification content is that of the raw html for the email body, except with all tags removed. This should largely remove need for separate push notification content. If there is no template body, only the notification title is displayed.
If this strategy doesn't work for a specific notification type for some reason (I didn't test most of them because I can't imagine why they wouldn't work) a separate content override in `WebpushService` should be enough.
- notifications use loomio icon/items from theme specified in environment. Imo, requiring a vector graphic in default and custom theming would be more elegant (less likely to have scaling issues)
- on mobile, app **should** display using icon specified in environment theme (untested, see below)
- notifications can be viewed which redirects to the `Notification.url` or - if none exists - /dashboard

Mobile functionality is untested but can only be tested in production because of browser security policies (phones will not be able to install the pwa from a development instance without tls/https because of content security policies).

@robguthrie awaiting review

robguthrie:
Amazing to see this. I've had a skim of the code and it looks like it's on the right track. I have a contract thing to finish before I can support you on this. 

Have you thought about what the UX implications are.. in terms of choosing between email or push notifications?

Hamsterarsch:
> Have you thought about what the UX implications are.. in terms of choosing between email or push notifications?

Regarding UX

Disabling/enabling push notifications can be done inside your browser/by uninstalling the pwa/by not granting push permissions on the mobile OS.
For control over what to get push notifications for, I was planning to just mirror the email settings for now but I don't recall if/how I implemented that (I sure didn't put it in tests so I'll look into it).

The only other UX concern I can see right now is how push notifications interact with the "Seen by" feature in threads; however, just seeing the title in the push notification and a few of its words didn't strike me to be the same as opening the notification email (which I believe Seen By is informed of when opening a notification email)

Do you see any other UX concerns beyond this?

Hamsterarsch:
PS: A user setting to disable email while keeping push also seems like a good idea then

PR #11675, opened 2025-07-24
wrap auth inputs under a single form tag

### LINKED ISSUE #11673: Login form autofill broken in some browsers due to missing form elements
## Environment
- **Loomio version:** 3.0.4 (self-hosted)
- **Browsers tested:** Firefox 140.0.4, Fennec, LineageOS Jelly
- **Working browser:** Chromium 138.0.7204.168

## Issue
Multi-step login form doesn't trigger any browser autofill behavior in some browsers, creating onboarding friction for new users.

## Symptoms
- Email field never pre-populates despite correct `autocomplete="username email"` attribute
- **Firefox doesn't even offer to save/remember the password**
- Users must manually type credentials every login
- Chromium works correctly, suggesting browser-specific compatibility issue

## Root Cause (Technical Investigation)
Login fields are not wrapped in proper `<form>` elements:
- **Email field structure:** DIV containers only, no form parent
- **Password field structure:** Same issue - DIV containers only  
- Both fields have correct HTML attributes but violate HTML standards for login forms

### Current HTML Structure
Email field is nested in:
```
DIV v-field__field
└── DIV v-field v-field--active...
    └── DIV v-input__control
        └── DIV v-input v-input--horizontal...
            └── DIV auth-email-form mx-auto...
```

**No `<form>` element found in parent hierarchy.**

### Email Field HTML
```html
<input name="email" placeholder="you@example.com" size="1" type="email" 
       id="email" aria-describedby="email-messages" 
       autocomplete="username email" class="v-field__input">
```

### Password Field HTML  
```html
<input autofocus="" name="password" size="1" type="password" 
       id="password" aria-describedby="password-messages" required="" 
       autocomplete="current-password" class="v-field__input">
```

## Test Results
JavaScript injection creating proper `<form>` wrappers:
- **Email field:** Firefox recognizes and saves as form data ✓
- **Password field:** Firefox now offers to save password ✓ 
- **Issue:** Separate forms prevent credential association

## Proposed Fix
Wrap both email and password fields in a single `<form>` element to enable proper browser autofill and password saving behavior.

robguthrie:
waiting for e2e's to pass. not manually tested but might fix #11673

### DIFF
--- a/vue/src/components/auth/modal.vue
+++ b/vue/src/components/auth/modal.vue
@@ -45,7 +45,7 @@ export default {
 }
 </script>
 <template lang="pug">
-.auth-modal
+form.auth-modal
   auth-form(v-if="!user.authForm" :user='user' :prevent-close="preventClose")
   auth-signin-form(v-if='user.authForm == "signIn"' :user='user')
   auth-signup-form(v-if='user.authForm == "signUp"' :user='user')


bobemoe:
Thanks for this PR, I tried the fix and it didnt seem to work... it did create a form element so its a good start and I can see where to edit things now, so I will compare the form to the form I injected in my testing in original issue and see what else is needed, I think I added autocomplete="on" or some other attributes to the form, but when trying that in the inspect element editor its not working either. I will test further and try and get back to you with some working vue code.

PR #11801, opened 2025-09-20
Pr 11796 fix:Add zh_CN locale

hpeXt:
Manually checked and edited Chinese version; corrected mistakes from Google Translate.


robguthrie:
@hpeXt Thanks for your efforts so far. 

However, It looks like there are problems introduced in your PR. I see lots of places where %{placeholder} is now %s{placeholder}, which will break the system. Also I've notices some places where your version is missing placeholders entirely. For example: 'unauthorized.manage.all' on line 258 in my PR and 263 in yours.

I've not checked the client files yet. 

Can you tell me what processing steps you have used?

PR #11860, opened 2025-10-09
fix: Add missing import for I18n in form

dmtrek14:
During some testing in our deployment, we came across this error:
Screenshot from 2025-10-09 12-20-00

It seems that the `vue/src/components/poll/common/form.vue`  was missing the import statement for the I18n package. This tiny fix PR adds the missing import.

### DIFF
--- a/vue/src/components/poll/common/form.vue
+++ b/vue/src/components/poll/common/form.vue
@@ -2,6 +2,7 @@
 import AppConfig from '@/shared/services/app_config';
 import Session from '@/shared/services/session';
 import { mapKeys, without, some, pick, snakeCase } from 'lodash-es';
+import { I18n } from '@/i18n';
 import Flash from '@/shared/services/flash';
 import Records from '@/shared/services/records';
 import EventBus from '@/shared/services/event_bus';


robguthrie:
thank you!

robguthrie:
Actually, your screenshot shows the existence of another bug I've seen recently. User without a secret_token value, it can mean that editing comments does not work. I'll deliver a fix for this shortly

anarute:
@robguthrie we also noticed that sometimes when editing something (a thread or a poll for example) the content comes up blank, could this be related with the token issue?

robguthrie:
yes, exactly. 

I’m releasing a fix for it at this moment.

anarute:
Btw we just tested 3.0.7 and we still see `[HocuspocusProvider] An authentication token is required, but you didn’t send one. Try adding a `token` to your HocuspocusProvider configuration. Won’t try again.`

And it's a bit random, sometimes the content comes up sometimes not which lead me to think it's a race condition of some sort

robguthrie:
Oh thanks for testing, that is disappointing. 

Did you run the migration that comes with it? My understanding of the problem is this:
- Newly created user records did not have a secret_token defined. This was due to a change in behavior of a library that was upgrade. You can run `User.where(secret_token: nil).count` to see if there are any users without secret_tokens.
- I've created a migration to add missing secret tokens, then force that the secret_token cannot be null.

So it was only happening for some users. I did not see inconsistency for the same user across refreshes.

anarute:
Yep we run the migration.

Running the command you suggested:

```
loomio(prod)> User.where(secret_token: nil).count
=> 0
```

I recorded a video, see how everything seems fine but then suddenly (at second 38) one of the edits come blank. After pressing Reset the content always comes correct, so we are using this to mitigate

https://github.com/user-attachments/assets/f58fb528-e821-4542-99fe-b660e95ba723

robguthrie:
Thanks for the video, however, you don't need to check for the missing text - If you see the warning text then there is a problem.

I'm not able to see the warning text anymore. Would you mind checking that you're definitely up to date?

PR #11874, opened 2025-10-16
Show how voted after closed

robguthrie:
Thanks for the inspiration @anarute. I like your attempt but I needed to have a go myself.

There are a few things happening. 
- We no longer need to build a new stance if myStance() returns null, as the server will always give us a vote if we can vote now. So we can delete that code.
- I wanted to document the purpose of the logic in watchRecords
- We can reduce duplication in the template between single and multiple choice cases.


anarute:
That's great, thanks!

PR #11925, opened 2025-11-02
Fix CSRF cookie secure flag for local development

etiennechabert:
## Problem

When `FORCE_SSL` is not set (for local development), the CSRF cookie is still hardcoded with `secure: true`, causing browsers to refuse sending it over HTTP. This results in silent 422 errors during login/signup.

## Solution

Set CSRF cookie `secure` flag based on `Rails.application.config.force_ssl` instead of hardcoding to `true`.

**File changed:** `app/helpers/protected_from_forgery.rb` (line 20)

```ruby
-secure: true
+secure: Rails.application.config.force_ssl
```

## Testing

With this change, when FORCE_SSL is not set:
- CSRF cookie is sent over HTTP
- Login/signup forms work correctly
- No more silent 422 errors

When FORCE_SSL is set (production):
- CSRF cookie remains secure
- No behavior change

### DIFF
--- a/app/helpers/protected_from_forgery.rb
+++ b/app/helpers/protected_from_forgery.rb
@@ -17,7 +17,7 @@ module ProtectedFromForgery
       cookies[:csrftoken] = {
         value: form_authenticity_token,
         expires: 1.day.from_now,
-        secure: true
+        secure: Rails.application.config.force_ssl
       }
     end
   end


robguthrie:
Thanks for your efforts, however, I'm unsure about this approach.

There are two things going on in this PR. Changes to how we read ENV's and the CSRF stuff.

All Loomio's boolean ENV's use the .present? approach, and there are a lot of them. I don't want to have to cast "false", 0, "0" and "" to false, everywhere. It's going to be hard to apply the practice everywhere - if it's not consistently applied then more people will trip up on it in various places.

It may be helpful to add a warning comment to DEV_SETUP.md, .env.development, .env.test and loomio-deploy/env_template explaining the only way to disable a flag is comment it. It is well known to ruby developers that the only falsely values are nil and false.

As for the CSRF issue, we already have line 11 in app/helpers/protect_from_forgery.rb#verified_request?, which returns true if the Rails.env is development.

It makes sense to keep your line 20: `secure: Rails.application.config.force_ssl` for consistency but it shouldn't make a difference, right? What browser did you have issues with? I would like to verify the problem.

At this stage I'm not happy to merge this. How about a PR with line 20, and adding some comments to the relevant help and .env files.

etiennechabert:
Hi @robguthrie, thanks for the thorough review and for taking the time to explain the approach.

## Context on the issue

I encountered this after setting `FORCE_SSL=false` in my `.env` for local development. The `.present?` check treated this as true, which then caused the CSRF cookie to be set with `secure: true` (line 19 in `protect_from_forgery.rb`). This prevented the cookie from being sent over HTTP, resulting in 422 errors on login/signup. I was testing on Chrome 131 on macOS.

While line 11's `Rails.env.development?` check should have bypassed CSRF validation, the issue was that the cookie wasn't being sent by the browser at all due to the `secure: true` flag - so there was nothing to validate.

## Regarding your points

I completely agree with your concerns about consistency. Changing the boolean parsing approach in one place while leaving `.present?` everywhere else would create confusion. The root cause was my misunderstanding of the ENV convention - I assumed `FORCE_SSL=false` would work like a boolean flag rather than needing to be commented out entirely.

## On the error handling

What made this particularly frustrating was the lack of visible error feedback. The backend returned 422, but neither the browser console nor the UI showed any actionable error message - just a silent failure on form submission. Even the Rails logs didn't clearly indicate the CSRF token was missing/invalid. This made debugging unnecessarily difficult and could affect other developers and users in similar situations.

Would you be open to improvements in this area? Specifically:
- **Frontend**: Display clear user-facing error messages when CSRF validation fails (e.g., "Security token invalid or expired. Please refresh the page and try again.") and log detailed diagnostic info to the browser console
- **Backend**: Enhanced application logging when CSRF validation fails, clearly indicating the failure reason

## Proposed path forward

I'll update this PR to:
1. Keep only line 20: `secure: Rails.application.config.force_ssl` (fixes the CSRF cookie issue)
2. If you're interested, add comprehensive CSRF error visibility improvements on both frontend and backend

Does this approach work for you?

etiennechabert:
I've simplified this PR to focus only on the essential fix (line 20: `secure: Rails.application.config.force_ssl`). 

Removed the ENV parsing changes and the initializer as discussed. The PR description has been updated for clarity about what's being merged.

If you're interested in improving CSRF error visibility (better logging and user-facing error messages), I'm happy to prepare that as a separate follow-up PR.

robguthrie:
Thank you! It's really wonderful to have your contribution here.

The need for better CSRF error visibility is smaller now the error is fixed. However I'm trying to improve the client errors throughout the app. If you feel strongly and want to do the work I'll help but I personally don't think it's necessary.

PR #11949, opened 2025-11-06
Fix NoMethodError in avatar_kind for Rails 7.2

etiennechabert:
## Problem

When running Loomio on Rails 7.2, the `avatar_kind` method in `HasAvatar` concern causes a `NoMethodError`:

```
NoMethodError: super: no superclass method 'avatar_kind' for an instance of User
at /loomio/app/models/concerns/has_avatar.rb:25
```

This error occurs when Sidekiq workers (like `PublishEventWorker`) try to serialize user data for email notifications, causing email delivery to fail.

## Root Cause

In Rails 7.2, when you override an attribute method that corresponds to a database column, calling `super` no longer works correctly. The method tries to call a parent implementation that doesn't exist in the expected way.

## Solution

Replace `super` with `read_attribute(:avatar_kind)` - the correct Rails pattern for accessing database column values in overridden attribute methods.

### Before
```ruby
def avatar_kind
  return 'mdi-duck' if deactivated_at?
  return 'mdi-email-outline' if !name
  super  # ❌ Doesn't work in Rails 7.2
end
```

### After
```ruby
def avatar_kind
  return 'mdi-duck' if deactivated_at?
  return 'mdi-email-outline' if !name
  read_attribute(:avatar_kind)  # ✅ Correct Rails 7.2 pattern
end
```

## Testing

Tested on:
- Rails 7.2.2.1
- Ruby 3.4.7
- Linux ARM64 (Raspberry Pi) and AMD64

After applying this fix:
- Invite emails send successfully
- No more `NoMethodError` in Sidekiq worker logs
- User serialization works correctly

## Impact

- **Scope**: Minimal - single line change
- **Risk**: Low - `read_attribute` is the canonical way to access column values
- **Backward compatible**: Yes - works on earlier Rails versions too

## References

- ActiveRecord attribute methods: https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods.html
- Related Rails issue pattern: https://github.com/rails/rails/issues/42918

### DIFF
--- a/app/models/concerns/has_avatar.rb
+++ b/app/models/concerns/has_avatar.rb
@@ -22,7 +22,7 @@ module HasAvatar
   def avatar_kind
     return 'mdi-duck' if deactivated_at?
     return 'mdi-email-outline' if !name
-    super
+    read_attribute(:avatar_kind)
   end
 
   def thumb_url


robguthrie:
This looks good, but I'm confused. Can you tell me how to produce the buggy behaviour?

etiennechabert:
I am sorry I cannot tell anything special I was doing, except probably run on ARM64, everything bundle together on a Raspberry Pi ? 

Testing my SMTP integration, I was confused not to see any email arrive, and then noted the exceptions taking place in the worker container.

I have my clone of your repositories now (https://github.com/etiennechabert/loomio, https://github.com/etiennechabert/loomio_channel_server) where I brought back ARM64, and which allowed me to test the fix of this PR. If you are really curious, here is my stack: https://github.com/etiennechabert/loomio-pi-stack

Not sure if this would help you to reproduce anything, I could also go back to your images and see if the issue come back.

robguthrie:
Could you tell me just a couple of discrete steps? Is it logging in? or a comment email? These things are currently working, and they are well tested in rspec and e2e tests. 
I don't think this is an ARM issue. 

Do you have the backlogs of the errors?

robguthrie:
Honestly, I like the change. I'm just confused about why I don't see the error.

etiennechabert:
I disabled regular subscriptions, so I was acting as admin, trying to invite my users, and no emails were actually sent with an error taking place on the worker side.

No worries, I will try to bring back the bug, and if so, I will give you the full stack (again, you will not have the origin, as this is a worker failing).

etiennechabert:
@robguthrie not used to GitHub and fork repo, sorry for all the noise 🤦 
I closed this PR by accident, re-opening it.
Please let me know if more is needed from my side at this stage

PR #11963, opened 2025-11-10
Add B2 API endpoint for creating comments

etiennechabert:
Adds `POST /api/b2/comments` endpoint for comment creation using `api_key` authentication.

**Changes:**
- Add `Api::B2::CommentsController` with create action
- Add route for comments in B2 namespace
- Add test coverage following B2 patterns

**Usage:**
```bash
POST /api/b2/comments
{
  "body": "Comment text",
  "body_format": "md",
  "discussion_id": 123,
  "api_key": "your-api-key"
}
```

### DIFF
--- a/app/controllers/api/b2/comments_controller.rb
+++ b/app/controllers/api/b2/comments_controller.rb
@@ -1,0 +1,10 @@
+class Api::B2::CommentsController < Api::B2::BaseController
+  def create
+    instantiate_resource
+    if CommentService.create(comment: @comment, actor: current_user)
+      respond_with_resource
+    else
+      respond_with_errors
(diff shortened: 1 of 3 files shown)


etiennechabert:
A bit of context for this:  
I'm exploring ways to automate note-taking during meetings. My idea is to post regular comments in a meeting thread with live summaries of the discussion.  

In that context, I think this API endpoint would be a great addition to complement the one that allows thread creation via the API.

robguthrie:
I'm always keen to hear context and about how people are using Loomio. Also delighted to receive contributions that are just genuinely useful. 

Thanks!

PR #11973, opened 2025-11-11
Fix markdown formatting preservation in auto-translate feature

etiennechabert:
## Summary

This PR fixes a bug where the auto-translate feature was breaking markdown formatting. When translating content, Google Translate was treating formatted text as plain text, causing loss of structure (headings, lists, bold, italic, etc.).

## Problem

The `TranslationService` was calling Google Translate API with default parameters, which treats all input as plain text. This caused markdown and HTML formatting to be lost or mangled during translation.

Example of the issue:
- **Before translation (German markdown):** `# Heading\n\n**Bold text**\n\n- List item 1`
- **After translation (broken):** `Heading Bold text List item 1` (all formatting lost)

## Solution

Modified `TranslationService.create` (app/services/translation_service.rb:14-48) to:
1. Detect if a field has a corresponding `_format` attribute (e.g., `description_format`)
2. For markdown content: convert to HTML using `MarkdownService.render_html()`
3. For both markdown and HTML content: pass `format: 'html'` parameter to Google Translate API
4. Store the translated HTML, preserving all formatting structure

## Changes

- Modified `app/services/translation_service.rb` to preserve formatting during translation
- Added comprehensive test coverage in `spec/services/translation_service_spec.rb`

## Testing

The new tests verify:
- Markdown content is converted to HTML before translation
- `format: 'html'` is passed to Google Translate API for formatted content
- Plain text fields are translated without format parameter
- Models without format fields continue to work as before

All tests pass successfully. The fix has been validated in production on our Loomio instance.

### DIFF
--- a/app/services/translation_service.rb
+++ b/app/services/translation_service.rb
@@ -21,7 +21,24 @@ class TranslationService
 
       model.class.translatable_fields.each do |field|
         next if model.send(field).blank?
-        translation.fields[field.to_s] = service.translate(model.send(field), to: locale)
+
+        # Check if the field has a format attribute (e.g., description_format)
+        format_field = "#{field}_format"
+        content = model.send(field)
+        translate_options = { to: locale }
+
+        if model.respond_to?(format_field)
+          format = model.send(format_field)
+
+          # If format is markdown or html, preserve structure by translating as HTML
+          if format == 'md' || format == 'html'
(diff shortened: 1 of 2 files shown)


etiennechabert:
Closing to recreate with only the translation fix commit (without other unrelated commits)

PR #11974, opened 2025-11-11
Fix markdown formatting preservation in auto-translate feature

etiennechabert:
## Summary

This PR fixes a bug where the auto-translate feature was breaking markdown formatting. When translating content, Google Translate was treating formatted text as plain text, causing loss of structure (headings, lists, bold, italic, etc.).

## Problem

The `TranslationService` was calling Google Translate API with default parameters, which treats all input as plain text. This caused markdown and HTML formatting to be lost or mangled during translation.

Example of the issue:
- **Before translation (German markdown):** `# Heading\n\n**Bold text**\n\n- List item 1`
- **After translation (broken):** `Heading Bold text List item 1` (all formatting lost)

## Solution

Modified `TranslationService.create` (app/services/translation_service.rb:14-48) to:
1. Detect if a field has a corresponding `_format` attribute (e.g., `description_format`)
2. For markdown content: convert to HTML using `MarkdownService.render_html()`
3. For both markdown and HTML content: pass `format: 'html'` parameter to Google Translate API
4. Store the translated HTML, preserving all formatting structure

## Changes

- Modified `app/services/translation_service.rb` to preserve formatting during translation
- Added comprehensive test coverage in `spec/services/translation_service_spec.rb`

## Testing

The new tests verify:
- Markdown content is converted to HTML before translation
- `format: 'html'` is passed to Google Translate API for formatted content
- Plain text fields are translated without format parameter
- Models without format fields continue to work as before

All tests pass successfully. The fix has been validated in production on our Loomio instance.

### DIFF
--- a/app/services/translation_service.rb
+++ b/app/services/translation_service.rb
@@ -21,7 +21,24 @@ class TranslationService
 
       model.class.translatable_fields.each do |field|
         next if model.send(field).blank?
-        translation.fields[field.to_s] = service.translate(model.send(field), to: locale)
+
+        # Check if the field has a format attribute (e.g., description_format)
+        format_field = "#{field}_format"
+        content = model.send(field)
+        translate_options = { to: locale }
+
+        if model.respond_to?(format_field)
+          format = model.send(format_field)
+
+          # If format is markdown or html, preserve structure by translating as HTML
+          if format == 'md' || format == 'html'
+            # Convert markdown to HTML if needed
+            content = MarkdownService.render_html(content) if format == 'md'
+            translate_options[:format] = 'html'
+          end
+        end
+
+        translation.fields[field.to_s] = service.translate(content, **translate_options)
       end
 
       translation.save!
--- a/spec/services/translation_service_spec.rb
+++ b/spec/services/translation_service_spec.rb
@@ -1,0 +1,105 @@
+require 'rails_helper'
+
+describe 'TranslationService' do
+  let(:user) { create(:user) }
+  let(:group) { create(:group) }
+  let(:google_service) { double('google_translate_service') }
+
+  before do
+    ENV['TRANSLATE_CREDENTIALS'] = 'fake_credentials'
+    allow(Google::Cloud::Translate).to receive(:translation_v2_service).and_return(google_service)
+  end
+
+  after do
+    ENV.delete('TRANSLATE_CREDENTIALS')
+  end
+
+  describe '.create' do
+    context 'with markdown formatted content' do
+      let(:discussion) do
+        create(:discussion,
+               author: user,
+               group: group,
+               title: 'Test Title',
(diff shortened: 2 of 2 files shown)


robguthrie:
This is somewhat helpful, but I'm starting to find the AI stuff quite time consuming to review. I wonder if you can send me less AI stuff please.

For example: The PR text is always a full page and it's not super helpful. As a staring point, just explain the problem, how to repeat it, and how you've tested the solution provided.

The tests are overkill and I would not like to add them. They spend a lot of lines stubbing the service, then proving that the stub returns the value you just told it to return. In this case, we can't go sending real API requests for testing, so it's necessary, but it's really a lot of lines to test that an if condition works correctly. I'm not a big fan.

Ok, for the rest of the solution:

The default value for the format option is :html. So the condition and some of the code presented is pointless - but quite close to helpful! I'm currently struggling with the translate service returning HTML encoded characters for text strings. This highlights what I need to do to fix that. Thanks!

So when translating, we want to convert md to html. Yep. Good. But .. we want to send format: :text when the model does not have a field_format value.

I'm not used to getting PRs for Loomio, so I'm not sure how to communicate what I expect for code, and I'm not used to the job of reviewing other peoples code. So it's a bit of a blessing and a learning curve for me. Thanks for bearing with me while I level up. I'm a big fan of a little AI, but I don't want to read any more than necessary, if possible, please.

robguthrie:
So, thanks for this. Overall, it's been helpful - particularly because the timing was uncanny. I've updated [my current PR](https://github.com/loomio/loomio/pull/11954) to ensure we handle markdown translation correctly.. and also plaintext translation correctly.

etiennechabert:
@robguthrie thanks for the transparency, overall my process so far has been:
  - Using Loomio as documented, and it's an absolutely great tool
  - When finding a bug, trying to find a fix with help of AI indeed
  - Validating the fix works on my fork/stack
  - Only then, opening a PR on your side
  - I understand your complaint regarding this been too verbose and time-consuming

  I guess going forward, when if I find bugs I'll still try to find a fix (for my own benefit), but only describe them as GitHub Issues instead of PRs, while referencing what I fixed on my side (just in case that would be any help to you), would that be better ?

  For the time being, we'll stay on our fork, as images for Loomio on dockerhub are still lacking ARM64, and we're still ahead regarding this fix: https://github.com/loomio/loomio/pull/11949

Thanks again for the great tool, and time you took with me

robguthrie:
I'm not saying don't make PR's .. overall this is net positive, but if you can take the good stuff the AI is doing, and try to critically look at what should be cut down to make the essence of the fix clear, that would be an improvement.

I'd like to know more about your organization - if you are willing to share. There are not so many Loomio users, and it's valuable to me to understand where it is being found useful. A big part of my job is keeping the business alive by marketing Loomio, and that means understanding what people value and communicating it others.

If you wish, contact me directly rob at loomio dot org and tell me about your group?

PR #12017, opened 2025-11-28
Fix poll option deletion by preserving _destroy parameter

etiennechabert:
Preserve underscore-prefixed keys when converting poll options to snake_case.

## Problem: 
Removing a poll option after creation silently fails. The _destroy key used by Rails to mark nested records for deletion was being mangled by the snake_case conversion.

## Fix: 
Skip keys that already start with an underscore during conversion.

## Result:
Tested and validated on our Loomio instance—poll options can now be removed as expected.

### DIFF
--- a/vue/src/components/poll/common/form.vue
+++ b/vue/src/components/poll/common/form.vue
@@ -161,7 +161,7 @@ export default {
       const actionName = this.poll.isNew() ? 'created' : 'updated';
       this.poll.setErrors({});
       this.setPollOptionPriority();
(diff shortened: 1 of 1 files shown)


robguthrie:
Hi!

Can you tell me how to repeat the issue and what the bug looks like when it happens?

etiennechabert:
Hi Rob! 

Screenshot 2025-11-30 at 22 37 32

Not able to remove a poll option after the poll have been published a first time, no particular error showing on the front, but opening the console you can see:
```
{
    "exception": "ActionController::UnpermittedParameters",
    "error": "found unpermitted parameter: :destroy"
}
```

robguthrie:
Thank you for the further explanation. That does make sense. Nice work :)

The only special key I know of is _destroy so we'll scope it like that.

PR #12149, opened 2026-01-30
Add Azure Translator support with modular provider architecture

etiennechabert:
Refactors translation service to support multiple providers with an abstraction pattern.
Adds Azure Translator as an option alongside Google Translate, and fixes translation service tests to work with the new provider architecture.

**Why:** Azure's free tier offers 2M characters/month vs Google's 500k.

**How it works:** Prefers Azure if credentials are configured, falls back to Google otherwise. Fully backward compatible - existing Google setups continue working unchanged.

**Changes:**
- Introduce modular provider architecture (`TranslationProviders::Base`, `Azure`, `Google`) replacing direct Google Translate coupling
- Add Azure Translator as an alternative provider, selected automatically when `AZURE_TRANSLATE_KEY` is set
- Fix translation service tests to stub at the provider level instead of the removed `Google::Cloud::Translate.translation_v2_service` path

**QA:** Our loomio instance is now using Azure, everything seems to work the same on simple and complex formats

**Possible next step:** This enables #12152 which could add automatic fallback between multiple providers to maximize combined free tiers. I am also planning to add more providers (DeepL next one)

**Configuration:**
```bash
# Azure (recommended for free tier)
AZURE_TRANSLATOR_KEY=your_key
AZURE_TRANSLATOR_REGION=eastus  # optional

# Google (existing, still works)
TRANSLATE_CREDENTIALS=path/to/credentials.json
```

### DIFF
--- a/test/services/translation_service_test.rb
+++ b/test/services/translation_service_test.rb
@@ -1,5 +1,4 @@
 require 'test_helper'
-require 'minitest/mock'
 
 class TranslationServiceTest < ActiveSupport::TestCase
   setup do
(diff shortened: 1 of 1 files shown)


etiennechabert:
@robguthrie if you want to move forward with the translation topic, I feel this is the first PR to be merged, as it contains the refactoring of the translation service ot make it provider agnostic

PR #12162, opened 2026-02-04
Add app feature to mask participants in anon poll receipts

tyler-dsa:
This PR adds an app feature, configured by a `FEATURES_MASK_ANONYMOUS_PARTICIPANTS` environment variable, to remove information on individuals from anonymous poll receipts, when the receipt is generated by a non-admin.

With the addition of verified participants this option is important for us as a mass political organization. We routinely use loomio polls to make decisions and need a way to do so that does not leak identifying information of members to any participant.

This PR also adds a few tests for the new functionality, which pass, but we have not tested it end to end just yet with the frontend.

Feedback is also (of course) welcome on approach. I was learning ruby on rails and the stack on the fly here.

### DIFF
--- a/app/controllers/api/v1/polls_controller.rb
+++ b/app/controllers/api/v1/polls_controller.rb
@@ -12,6 +12,7 @@ class Api::V1::PollsController < Api::V1::RestfulController
     memberships = @poll.group.present? ? @poll.group.memberships.where(user_id: receipts.map(&:voter_id)).index_by(&:user_id) : {}
     voters = User.where(id: receipts.map(&:voter_id)).index_by(&:id)
     inviters = User.where(id: receipts.map(&:inviter_id)).index_by(&:id)
+    mask_receipts = AppConfig.app_features[:mask_anonymous_participants] && @poll.anonymous && !is_admin
 
     render json: {
       voters_count: @poll.voters_count,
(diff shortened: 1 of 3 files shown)


robguthrie:
Thanks for kicking off a contribution. My first reaction is why don't we just have a ENV to disable the receipts feature entirely?

robguthrie:
Oh reading more closely this is just about who we show the receipts to.. and I see it as effectively making receipts admin only. 

So I would suggest we look at changing the abilities for a poll.

So at the to of api/v1/polls_controller#receipts
```
@poll = load_and_authorize(:poll, :show_receipts)
```

Then in ability/polls.rb
```
    can :receipts, ::Poll do |poll|
      # if env is set then require poll admin, else require can vote in poll
    end
```

Are you working for NYC-DSA? I'd like to support you closely. Please email me. rob@loomio.org

robguthrie:
Hi @tyler-dsa I'm also interested in developing STV ranked choice, is this something you're familiar with?

tyler-dsa:
> My first reaction is why don't we just have a ENV to disable the receipts feature entirely?

This might be ideal honestly! Masking what users can see when generating the receipt just seemed like the path of least resistance to me (someone without much ruby experience). Happy to adjust the approach if you have a pointer or two?

> Are you working for NYC-DSA?

Yes :)

> I'm also interested in developing STV ranked choice, is this something you're familiar with?

Absolutely! STV (usually Scottish, sometimes Meek) elections are super common. Most chapters use Opavote but I don't think anyone likes it. I'll ask around for thoughts on STV in loomio and follow up over email.

robguthrie:
Thanks for kicking things off. I've added the feature to master now. See https://github.com/loomio/loomio/pull/12176

Looking forward to more collaboration with you!

PR #12203, opened 2026-02-24
Poll comments

robguthrie:
This restructures the app so that polls can be threads, just like discussions. This means you can start a poll and then reply to a voter. Or start a poll and have someone ask a question, which you can reply to.

How?
Introduce a topic model, this sits between group and discussion/poll. The topic is responsible for thread state, reader state, guest permissions and group.

So discussions have a topic and polls have a topic. When you have a thread with a discussion and polls, they share the same topic.

There are a number of places where this significantly reduces complexity, because we dont have to check poll, discussion, group for permission, we just check the topic.

It's such a massive change, but i think it will be really helpful from a UI/UX perspective, and a simplification of the codebase, once the refactor is done.


app/mailers/group_mailer.rb
```diff
@@ -13,4 +13,18 @@ def destroy_warning(group_id, recipient_id, deletor_id)
       I18n.t("group_mailer.destroy_warning.subject")
     }
   end
+
+  def trial_expired(group_id, recipient_id)
```
robguthrie: leaked in from other branch?

app/models/ability/topic.rb
```diff
@@ -0,0 +1,42 @@
+module Ability::Topic
+  def initialize(user)
+    super(user)
+
+    can [:show, :mark_as_seen, :mark_as_read, :dismiss, :set_volume], ::Topic do |topic|
+      can?(:show, topic.topicable)
```
robguthrie: is this heading circluar? it's fundamentally topic permissions

PR #12348, opened 2026-04-21
Add list endpoints to api/b2 for discussions and polls

robguthrie:
## Summary

External customer (via support) couldn't list discussions or polls via the public API — only `create` and `show` were exposed on `api/b2`. This adds `GET /api/b2/discussions` and `GET /api/b2/polls`, scoped to a `group_id` the caller is a member of (or a global admin).

Also standardises pagination params across Snorlax: `limit` / `offset` are now the primary names; legacy `per` / `from` still work as aliases.

- `status=open|closed|all` for discussions (default `open`)
- `status=active|closed|all` for polls (default `active`)
- `limit` / `offset` for pagination (or legacy `per` / `from`)
- Poll responses include `current_outcome` via `PollSerializer`, so callers can list "consented decisions" with `status=closed` and inspect outcomes client-side.

Docs at `/help/api2` updated with both endpoints and the new param names.

## Test plan

- [x] `bin/rails test test/controllers/api/b2/` — 50 runs, 119 assertions, 0 failures
- [x] `bin/rails test test/controllers/api/` — 409 runs, 1010 assertions, 0 failures (no v1 regressions from the shared `page_collection` change)
- [ ] Spot-check `/help/api2` rendering in the browser
- [ ] `curl '$ROOT_URL/api/b2/discussions?api_key=$KEY&group_id=$ID'` against a dev instance
- [ ] `curl '$ROOT_URL/api/b2/polls?api_key=$KEY&group_id=$ID&status=closed'` to verify outcome inclusion

### DIFF
--- a/app/controllers/api/b2/discussions_controller.rb
+++ b/app/controllers/api/b2/discussions_controller.rb
@@ -9,4 +9,28 @@ class Api::B2::DiscussionsController < Api::B2::BaseController
     DiscussionService.create(actor: current_user, discussion: @discussion, params: params)
     respond_with_resource
   end
+
+  def index
+    instantiate_collection { |collection| collection.order_by_latest_activity }
+    respond_with_collection
+  end
+
+  def accessible_records
+    scope = Discussion.where(group_id: group.id)
+    case params[:status]
+    when 'closed' then scope.is_closed
+    when 'all'    then scope.kept
+    else               scope.is_open
+    end
+  end
+
+  private
+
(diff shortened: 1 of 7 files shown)


Whelkaholism:
Hiya, it is I, Mr External Customer. Thanks so much for this! Only thing is I am seeing duplicate keys for polls/discussions in the applicable responses; I am using a .NET client which chokes on them.

PR #12379, opened 2026-04-29
Add pg_trgm GIN indexes for ILIKE search columns

robguthrie:
## Summary

- Enables `pg_trgm` (on DO managed Postgres allowlist) and adds GIN trigram indexes on the columns currently searched with leading-wildcard `ILIKE '%q%'`.
- Indexed: `users.name/username/email`, `groups.name/handle/description`, `discussions.title`, `polls.title`, `documents.title`, `active_storage_blobs.filename`.
- Indexes built `CONCURRENTLY` (no DDL transaction) — safe to deploy without locking writes. Slowest builds will be `users` and `active_storage_blobs`.
- Zero application code changes — Postgres uses GIN trgm indexes automatically for `ILIKE '%foo%'`. The 8 existing `ilike` scopes across `app/models/{user,group,membership,discussion,poll,document}.rb` and `app/queries/{membership_query,attachment_query}.rb` are unchanged.

## Why

The audit of search code found two layers: (1) `pg_search` multisearch on `pg_search_documents` for the global search box, and (2) leading-wildcard `ILIKE` for every secondary surface (member picker, group lookup, polls list, discussion filter, attachment search). Layer (2) was sequential-scan only. This PR fixes that with the cheapest possible change — extension + indexes, no query rewrites.

## Test plan

- [ ] CI green
- [ ] After deploy, EXPLAIN one of the `ILIKE` queries on prod (e.g. `users` member picker) and confirm `Bitmap Index Scan on index_users_on_name_trgm`
- [ ] Spot-check that the global search box, member picker, and group/poll search still return the expected results

### DIFF
--- a/db/migrate/20260430104335_add_pg_trgm_indexes.rb
+++ b/db/migrate/20260430104335_add_pg_trgm_indexes.rb
@@ -1,0 +1,21 @@
+class AddPgTrgmIndexes < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    enable_extension :pg_trgm
+
+    add_index :users,    :name,     using: :gin, opclass: :gin_trgm_ops, algorithm: :concurrently, name: :index_users_on_name_trgm
+    add_index :users,    :username, using: :gin, opclass: :gin_trgm_ops, algorithm: :concurrently, name: :index_users_on_username_trgm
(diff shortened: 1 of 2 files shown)


robguthrie:
Closing — not pursuing this change.

PR #12430, opened 2026-05-19
[codex] require verified email for generic oauth

robguthrie:
## Summary
- Read the generic OAuth verified-email claim.
- Reject generic OAuth callbacks when the provider has not verified the email.
- Add an explicit override for providers that cannot supply this claim.

## Why
Generic OAuth should not auto-link or verify local accounts from an unverified email claim.

## Tests
- bundle exec rails test test/controllers/identities/oauth_controller_test.rb test/controllers/identities/google_controller_test.rb test/controllers/identities/nextcloud_controller_test.rb


robguthrie:
dont need this.

Issue #12471, opened 2026-06-02
GroupExportService.import fails with RuntimeError


jmakdah2:
I was attempting to re-create a Loomio instance my group is self-hosting on a new server and after exporting the data via the group export function, I followed the steps outlined in https://help.loomio.com/en/user_manual/groups/data_export/index.html but ran into this error:

```
app/services/group_export_service.rb:268:in 'block (3 levels) in GroupExportService.import': failed to import groups record - conflict on unique column. handle that here (RuntimeError)
	from app/services/group_export_service.rb:241:in 'Array#each'
	from app/services/group_export_service.rb:241:in 'block (2 levels) in GroupExportService.import'
	from app/services/group_export_service.rb:238:in 'Array#each'
	from app/services/group_export_service.rb:238:in 'block in GroupExportService.import'
	from app/services/group_export_service.rb:236:in 'GroupExportService.import'
	from (loomio):8:in '<main>'
```

I did a couple of debugging steps including making sure that the new instance's database was empty (including the groups table) and checking the outputted json for duplicate group records (which was not found).

I was able to successfully run this command a couple of months ago although we were on a different version of loomio at that time. Could you please let us know if there is any ideas how to debug this further, and ideally also add some more descriptive output to the error message like which specific records are marked as duplicates? Thank you!

robguthrie:
Are both servers running the same version of Loomio? Please include versions and I'll see what I can do

jmakdah2:
Yes they are both on v3.0.20. Thanks!

robguthrie:
I'm slowly working on this. I'm actually getting errors at a different stage. I'll let you know when I've fixed it for the test cases I have.

pmelanson:
[jmakdah2 and I have both been looking into this, I'm working off the same data and versions as them]

I'm also working towards a minimum reproducible JSON dump. Haven't quite gotten there, but I at least tried to make this more reproducible.

I've gotten into a workflow where I drop the `loomio_production` database, recreate it with `rake db:setup`, and then try importing to see what happens (technically our loomio deployment is a [multi-docker container thing](https://git.coopcloud.tech/coop-cloud/loomio/src/branch/main/compose.yml) but I don't think this should affect anything).

So if I do:

```shell
$ dropdb loomio_production && createdb loomio_production
$ rake db:setup
$ rails console
> GroupExportService.import('/import/loomio-testdata.json')
```

[...]

PR #12495, opened 2026-06-08
Rewrite group importer to pre-map all ids, insert once

robguthrie:
## Problem

`GroupExportService.import` crashed on real-world ("gnarly") imports:

```
PG::UniqueViolation: duplicate key value violates unique constraint
"index_topics_on_topicable_type_and_topicable_id"
```

The old two-phase importer inserted records with Postgres-assigned ids (negating topics' `topicable_id` to dodge the unique index during the window before remapping), then walked `BACK_REFERENCES` in a second pass to `update_all` every foreign key. That rewrite pass had collision windows it couldn't escape on complex data.

## Solution

Pre-map every id, then insert once:

1. **Allocate** — reserve every new id up front from the table sequences (`reserve_ids`), building the complete `old_id → new_id` map before any insert.
2. **Insert** — translate every foreign key in-memory (`translate_foreign_keys!`) and insert each record exactly once in its final form.

No row is mutated after insert → no negation, no second `update_all` pass, no collision window by construction. Also drops a full rewrite sweep over the dataset, so it's **faster**.

### Why it's safe

- The only DB-level FKs (`discussions`/`polls` → `topics`) are `deferrable: :deferred`, so insert order within the transaction is irrelevant.
- `topicable` / `eventable` / `reactable` / `parent` are polymorphic — no FK constraint.
- Reserved ids come from `nextval` (after `setval` to `max(max_id, last_value)`), so no existing row can already point at them.

`FORWARD_REFERENCES` inverts `BACK_REFERENCES` (still the single source of truth) into `source_table → {column → target_table}`; polymorphic columns resolve their target from the record's stored `<column>_type`.

## Verification

- `test/services/group_export_service_test.rb` — green (round-trip + the collision regression test).
- Confirmed working on real-world gnarly export data that previously crashed.

## Known follow-ups (pre-existing, not introduced here)

- Polymorphic refs whose target table isn't in the export keep a stale old id (same as old code).
- Attachment download still the commented-out block.

## Commits

- Rewrite group importer to pre-map all ids, insert once
- Normalize tag names by stripping whitespace (supports tag uniqueness on import)


robguthrie:
#12471

PR #12811, opened 2026-09-03
Drop unused variables from lmo_textarea editor destructuring

marekl11:
`md_editor.vue` and `collab_editor.vue` both pull a bundle of values out of
the attaching/mentioning composables and then never use some of them -
`imageFiles` and `mentionsCache` in both files, plus `updateFiles` in
`md_editor.vue` (`collab_editor.vue` still calls it, so it stays there).

Just removing the unused names from the destructuring; the composables
themselves are untouched.


robguthrie:
Arr already conflicted and no real value, thanks anyway.

PR #12836, opened 2026-09-08
Warn before deleting expired subscription groups

robguthrie:
Expired trial cleanup needs two paths based on whether the group tree contains user content. Topic-free trial trees are deleted automatically 60 days after expiry; trials with topics and subscriptions cancelled for 60 days are discarded, their administrators are warned, and permanent deletion is scheduled two weeks later.

Warning emails include tree-wide counts for subgroups, current members, discussions, polls, and comments, plus data-export guidance and a reply path for restoration. The daily workers recheck eligibility before acting, exclude billed/shared subscription trees from empty cleanup, process warning candidates oldest-first in bounded batches, and write structured audit records. Instance administrators retain an immediate deletion option.

This PR is stacked on #12835 because it uses the new group discard lifecycle and availability contract.

Validation:

- 206 focused Rails tests, 1,955 assertions
- Fixture matrix covers recent/no-expiry/discarded/content-bearing/billed/shared/child-subscription groups
- Direct discussion and `NullGroup` behavior remains covered by #12835
- RuboCop syntax checks on all warning and cleanup paths
- User manual build: 94 pages and 10,623 internal links/assets
- Brakeman: 0 warnings
- Bundler Audit: no vulnerabilities
- New server strings translated across all supported locales; interpolation variables verified

### DIFF
--- a/app/controllers/admin/groups_controller.rb
+++ b/app/controllers/admin/groups_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Admin::GroupsController < Admin::BaseController
-  before_action :load_group, only: %i[show edit update move handle discard undiscard delete_group export_group]
+  before_action :load_group, only: %i[show edit update move handle undiscard warn_then_destroy destroy_immediately export_group]
 
   def index
     groups, pagination = paginate(filtered_groups)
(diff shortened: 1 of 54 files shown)


robguthrie:
Superseded by #12837, which removes subscription-cancellation cleanup and covers expired trials only.

